PR #1, opened 2015-03-06
Make package file paths more specific

dashea:
Avoid claiming ownership of system directories, and split the devel
packages so that each includes only the headers relevant to the
particular package.

Also the .gir file is a devel file.


vpodzime:
Looks good to me, thanks!

PR #4, opened 2015-03-20
Better exception/error reporting and handling

vpodzime:
These patches add the *ErrorProxy* class and its instances that make exception reporting and handling in Python more granular and thus better usable. See the commit cdc09ea for some basic rationale. Additional commits add some very basic tests and documentation for what is being added.


vpodzime:
Unfortunately it seems that the commit 3047561 will have to be dropped because eea32e9 doesn't really work the way I thought it works. See the following piece of IRC chat with the introspection guys:

    <vpodzime> thiblahute: if no overrides are installed in the system it uses the ones in src/python/gi/overrides/
    <thiblahute> vpodzime, Right, for a particular override yes
    <vpodzime> thiblahute: but if there are overrides installed on the system, it uses the ones from /usr/lib64/python-2.7/...
    <thiblahute> vpodzime, Yeah, there is not much we can do about that with the current implementation
    <vpodzime> thiblahute: well, that is kind of a show-stopper for testing
    <thiblahute> vpodzime, You need to make sure you do not have it installed (really stupid, I know :()
    <vpodzime> thiblahute: that's impossible for 'make test'
    <thiblahute> Erg, indeed

vpodzime:
Turns out we only need one more nasty hack to make it work well enough to keep the tests in place.

vpodzime:
Added one more commit that implements the hack mentioned in the previous comment.

mulkieran:
This looks good to me, but I'm puzzled by the context in which it is going to be used. Is the idea or possibillty that clients e.g., blivet can override BlockDev.py somehow to make their own exception mapping. Or will all clients need to use whatever libblockdev provides? IOW, is this override just an introspection thing or can clients use it for customization?

Related is the fact that the ErrorProxy.__init__() method takes two lists, basically a source Error and a proxy Error list, and maps from one to the other based on the index. I can't imagine any situation where the lists would be constructed separately, so I don't see why not a list of tuples or some other thing with the mapping already built in.

When blivet gets the new version of libblockdev it will have to change all invocations of plugins, like swap.swapon instead of swap_swapon. But other than that it will just need to import the exceptions declared in BlockDev.py and use them where appropriate in exception handlers?

Since it might be necessary to match on strings using xrule to make dispatch work properly, it might make sense if that happens to decide on a grammar for strings so that they started with some distinguishing tag that the regular expression can find easily. Otherwise, the dispatch is going to be brittle, like the various tests in blivet that use self.assertRaisesRegex which I tend to break when improving an error message.

It would be nice to have a unit test of the XRule part of the dispatch in ErrorProxy __getattr__ method, if possible.

vpodzime:
> Is the idea or possibillty that clients e.g., blivet can override BlockDev.py somehow to make their own exception mapping. Or will all clients need to use whatever libblockdev provides? IOW, is this override just an introspection thing or can clients use it for customization?

libblockdev is a library and as such it provides some API. A part of the API are exceptions/error raised and this patch only makes this easier/better/nicer to process such errors/exceptions in python. libblockdev cannot raise blivet's exceptions (instances of blivet's exception classes) as part of its API. It could provide a way to let blivet inject its own (or any other use code's) exceptions into ``ErrorProxy`` instances, but I'd like to leave that for a further discussion and a follow-up patch because I'm not sure that's a good thing to do (basic question: any other library that allows/does that?).

> Related is the fact that the ErrorProxy.init() method takes two lists, basically a source Error and a proxy Error list, and maps from one to the other based on the index. I can't imagine any situation where the lists would be constructed separately, so I don't see why not a list of tuples or some other thing with the mapping already built in.

That's a good point. I remember I had a reason for the two-lists implementation, but now I can no longer see the reason. Maybe it was related to a different implementation of ``XRules`` I originally had. Looking at the code now I think there's no reason to not use a single list of tuples. Fixing locally again. Thanks!

> When blivet gets the new version of libblockdev it will have to change all invocations of plugins, like swap.swapon instead of swap_swapon. But other than that it will just need to import the exceptions declared in BlockDev.py and use them where appropriate in exception handlers?

Blivet will have a chance to use the new API, the old one will still be available. But yes, calls should be changed to use the ErrorProxy instances (i.e. ``swap.swapon`` instead of ``swap_swapon`` just like you mentioned) and blivet should start catching BlockDevError instances (and SwapError,... instances) instead of GLib.GError instances.

> Since it might be necessary to match on strings using xrule to make dispatch work properly, it might make sense if that happens to decide on a grammar for strings so that they started with some distinguishing tag that the regular expression can find easily. Otherwise, the dispatch is going to be brittle, like the various tests in blivet that use self.assertRaisesRegex which I tend to break when improving an error message.

That would be nice, but libblockdev in many cases just passes errors (GError instances) reported by various GLib functions further or constructs the error messages from utils' stderr/stdout -- it's easier to distinguish between them in the high-level code if needed. I'm not saying that's never going to change, but right now, it would make the libblockdev's code more complicated for no real improvement -- the GLib API (including error messages) is stable and so should be the outputs of utilities/tools that are considered API (like ``lvm``, ``btrfs``, ``mdadm``,...).

> It would be nice to have a unit test of the XRule part of the dispatch in ErrorProxy getattr method, if possible.

Absolutely agreed. There should be much more tests for the ErrorProxy and what you mention here should be one of the first such tests.

dwlehman:
@vpodzime please open an issue against blivet for using the new API and catching BlockDevError, &c throughout.

PR #6, opened 2015-03-24
Sort BTRFS subvolumes in a way that child never appears before parent

vpodzime:
Resolves: rhbz#1201120

When processing subvolumes, it is often desirable to process the parent
(sub)volume first because child subvolumes often have references to their
parents. That's why we need to sort subvolumes that way when returning a list of
them.


vpodzime:
I checked the functionality on data extracted from the bug report and it seems to work as expected.

PR #7, opened 2015-04-02
FW RAID fixes

vpodzime:
Follow-up fixes for [rhbz#1207317] (https://bugzilla.redhat.com/show_bug.cgi?id=1207317)


AdamWill:
Isn't this missing one of the UUID changes? It only fixes the UUID stuff in one place, didn't we find two places it needs fixing?

When I build libblockdev 0.8 plus this patch, I still get a 'malformed or invalid UUID' error.

vpodzime:
> Isn't this missing one of the UUID changes? It only fixes the UUID stuff in one place, didn't we find two places it needs fixing?

No, that was pushed before to master as commit 4d917ebb5.

PR #8, opened 2015-04-08
check functions for plugins (#1201933)

vpodzime:
When loading plugins, we should check if the environment provides everything they need. This is the first step towards that goal.

This PR is expected to get more commits (with other plugins' check functions), but an early review of the basic design and implementation decisions would be nice.

### DIFF
--- a/src/lib/blockdev.c
+++ b/src/lib/blockdev.c
@@ -171,16 +171,16 @@ gboolean bd_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func, GErro
         return FALSE;
     }
 
+    if (log_func && !bd_utils_init_logging (log_func, error))
+        /* the error is already populated */
+        success = FALSE;
+
     if (!load_plugins (require_plugins, FALSE)) {
         g_set_error (error, BD_INIT_ERROR, BD_INIT_ERROR_PLUGINS_FAILED,
                      "Failed to load plugins");
         /* the library is unusable without the plugins so we can just return here */
         success = FALSE;
     }
-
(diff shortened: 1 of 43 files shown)


vpodzime:
This is related to [blivet's pull #60] (https://github.com/rhinstaller/blivet/pull/60).

mulkieran:
I didn't delve into individual procedures, but overall it looks fine.

I think the name "check" should be made more specific...maybe check_availability or check_external or something.

And the error message in that method should probably be more specific to its function as well, something like
"Unable to verify availability of required external dependency 'lvm': %s". Someday it may seem like the correct idea to load the plugin even if the external dependency is not there at the time.

vpodzime:
> I think the name "check" should be made more specific...maybe check_availability or check_external or something.

The name ``check`` is intentionally that generic. At some point, some plugin could check itself or whatever. I don't see any benefit in giving it a more specific name. I could move the code into a new function called ``check_external`` and call that from ``check``, but I don't see any benefit in that neither.

> And the error message in that method should probably be more specific to its function as well, something like
"Unable to verify availability of required external dependency 'lvm': %s".

The error message going from where? The ``bd_utils_check_util_version`` has no idea it's an external dependency, the ``check`` function just prints a warning and returns ``FALSE`` with no error passed to the caller. And that warning is like *"Cannot load the LVM plugin: The 'lvm' utility is not available"*. I think that's enough specific.

> Someday it may seem like the correct idea to load the plugin even if the external dependency is not there at the time.

Yes and I'm gonna bother with that **someday**. :)

vpodzime:
What I'd love to be the final version of this PR is now pushed and ready for a review.

PR #9, opened 2015-04-08
Disable pylint checking on the new exception proxy.

clumens:
pylint is not clever enough to figure this out.


vpodzime:
What bugs me about this a bit is that pylint on my Fedora 21 (pylint-1.3.1-3.fc21.noarch) doesn't complain about this. Anyway, thanks for the patch!

PR #12, opened 2015-05-06
The KBD plugin

vpodzime:
These commits implement the KernelBlockDevices plugin as it was planned before.

I'm sorry for the commit 09a61a2, but when I got to ``time.sleep(10)`` basically between every two lines of the ``bcache`` tests, I have up and just excluded the tests from normal runs. I run the myself as I know how to get the system back to the original state if some *bcache magic* happens.


vpodzime:
No objections, let's move with this.

PR #14, opened 2015-05-13
Reimplement the is_mpath_member() function using device mapper

vpodzime:
This way we don't have to call the 'multipath' binary which is suboptimal.


vpodzime:
I just manually checked the new implementation works as expected.

PR #16, opened 2015-05-18
Set encoding to NULL before writing the escrow packet.

dashea:
By default GIOChannel expects data to be UTF-8, and this data definitely
isn't.


dashea:
Hold up, it's still crashing on me.

dashea:
Something was getting freed in gobject-introspection that shouldn't have been. consting everything up worked in the one attempt I've run.

dashea:
And one more so the file is named right

vpodzime:
Looks good to me, thanks! We should probably make all the ``gchar *`` parameters const as none of them is changed in any of the functions. But I'd like to merge this as it is and do the bigger change as a separate patch/PR.

Issue #15, opened 2015-05-14
Enhance granularity and reporting when initializing/loading plugins


vpodzime:
Right now, an error/exception is reported/raised if some plugins fail to load in ``init()``, ``try_init()`` and ``reinit()``. That's a bit unfortunate for the "try to load as much as possible and report back" use case which seems to be useful in some cases.

vpodzime:
Should be resolved by PR #19.

PR #19, opened 2015-05-20
Make the init functions better usable

vpodzime:
This tries to address the issue #15 by renaming the ``try_init()`` function to ``ensure_init()`` which is a better name for what it does and by adding new ``try_init()`` and ``try_reinit()`` functions that do "best-effort" plugin loading and report back the list of loaded plugins.

### DIFF
--- a/src/lib/blockdev.c
+++ b/src/lib/blockdev.c
@@ -316,6 +316,60 @@ gboolean bd_ensure_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func
 }
 
 /**
+ * bd_try_init:
+ * @request_plugins: (allow-none) (array zero-terminated=1): %NULL-terminated list
+ *                   of plugins that should be loaded (if no so_name is specified
+ *                   for the plugin, the default is used) or %NULL to load all
+ *                   plugins
+ * @log_func: (allow-none) (scope notified): logging function to use
+ * @loaded_plugin_names: (allow-none) (out) (transfer container): names of the successfully
+ *                                                                loaded plugins
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the library was successfully initialized with all the
+ *          required or default (see @require_plugins) plugins or not
+ *
+ * *UNLIKE IN CASE OF bd_init() AND bd_ensure_init(), FAILURE TO LOAD A PLUGIN
+ *  IS NOT CONSIDERED ERROR*
+ */
+gboolean bd_try_init(BDPluginSpec **request_plugins, BDUtilsLogFunc log_func,
+                     gchar ***loaded_plugin_names, GError **error) {
+    gboolean success = TRUE;
+    guint64 num_loaded = 0;
+
+    g_mutex_lock (&init_lock);
+    if (initialized) {
+        g_warning ("bd_try_init() called more than once! Use bd_reinit() to reinitialize "
+                   "or bd_is_initialized() to get the current state.");
+        g_mutex_unlock (&init_lock);
+        return FALSE;
+    }
+
+    if (log_func && !bd_utils_init_logging (log_func, error))
+        /* the error is already populated */
+        success = FALSE;
+
+    success = load_plugins (request_plugins, FALSE, &num_loaded);
+
+    if (num_loaded == 0) {
+        if (request_plugins && (*request_plugins == NULL))
+            /* requested to load no plugins (NULL is the first item in the
+               array), none loaded -> OK */
+            initialized = TRUE;
+        else
+            initialized = FALSE;
+    } else
+        initialized = TRUE;
+
+    if (loaded_plugin_names)
+        *loaded_plugin_names = bd_get_available_plugin_names ();
+
+    g_mutex_unlock (&init_lock);
+
+    return success;
+}
+
+/**
  * bd_reinit:
  * @require_plugins: (allow-none) (array zero-terminated=1): %NULL-terminated list
  *                 of plugins that should be loaded (if no so_name is specified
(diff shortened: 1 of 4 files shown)


src/lib/blockdev.c
```diff
@@ -267,17 +361,85 @@ gboolean bd_try_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func, G
...
+               array), none loaded -> OK */
+            initialized = TRUE;
+        else
+            initialized = FALSE;
+    } else
+        initialized = TRUE;
+
+    if (loaded_plugin_names) {
+        *loaded_plugin_names = bd_get_available_plugin_names ();
+    } else
```
mulkieran: missing else block here?

src/lib/blockdev.c
```diff
@@ -267,17 +361,85 @@ gboolean bd_try_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func, G
...
+ */
+gboolean bd_try_reinit (BDPluginSpec **require_plugins, gboolean reload, BDUtilsLogFunc log_func,
+                        gchar ***loaded_plugin_names, GError **error) {
+    gboolean success = TRUE;
+    guint64 num_loaded = 0;
+
+    g_mutex_lock (&init_lock);
+    if (log_func && !bd_utils_init_logging (log_func, error))
+        /* the error is already populated */
+        success = FALSE;
```
mulkieran: This construction occurs in lots of places, but it always seems weird to me. Why continue if failure has already occurred? Why say that you have failed when you're going to keep on trying? Isn't the success of failure of initializing logging orthogonal to the success or failure of the actual task?

src/lib/blockdev.c
vpodzime: Yes, I accidentally added this right before pushing this branch to my GH. Fixing.

src/lib/blockdev.c
vpodzime: That's a very good point, I just copied the chunk of code here, but it should be fixed in all places. Adding one more commit fixing this.

src/lib/blockdev.c
```diff
@@ -221,29 +237,107 @@ gboolean bd_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func, GErro
...
+                   "or bd_is_initialized() to get the current state.");
+        g_mutex_unlock (&init_lock);
+        return FALSE;
+    }
+
+    if (log_func && !bd_utils_init_logging (log_func, error))
+        /* the error is already populated */
+        success = FALSE;
+
+    success = load_plugins (request_plugins, FALSE, &num_loaded);
```
mulkieran: It would make more sense to me if ```load_plugins``` had an array of plugins as its third parameter, not a number, which is an information poor summary of the result of the loading effort. I think that ultimately it would be possible to be more direct in calling code with that interface.

mulkieran:
My highest order worry is the case of the duelling libblockdev users. It seems like one process, ignorant of the other, can unload a bunch of plugins, while the other process, confident that it has all the plugins it needs, since it loaded them prior to the unloading, will forge ahead until it gets an error due to the missing plugin.

vpodzime:
> My highest order worry is the case of the duelling libblockdev users. It seems like one process, ignorant of the other, can unload a bunch of plugins, while the other process, confident that it has all the plugins it needs, since it loaded them prior to the unloading, will forge ahead until it gets an error due to the missing plugin.

Processes do not share the library state, each process has a separate state and a separate set of plugins loaded. What's described above can only happen if multiple libraries/modules/... in a single process use libblockdev and call ``reinit(reload=True)`` in ugly ways. But I don't think something like that needs to be addressed by the library it self.

mulkieran:
I'm still puzzled why new third parameter of load_plugins is a number, which is a lot less information rich than a list or an array of the actually loaded plugins. It certainly has access to that information and I think user code could be a bit more direct with that API.

src/lib/blockdev.c
```diff
@@ -221,29 +239,111 @@ gboolean bd_init (BDPluginSpec **require_plugins, BDUtilsLogFunc log_func, GErro
...
+ *  IS NOT CONSIDERED ERROR*
+ */
+gboolean bd_try_init(BDPluginSpec **request_plugins, BDUtilsLogFunc log_func,
+                     gchar ***loaded_plugin_names, GError **error) {
+    gboolean success = TRUE;
+    guint64 num_loaded = 0;
+
+    g_mutex_lock (&init_lock);
+    if (initialized) {
+        g_warning ("bd_init() called more than once! Use bd_reinit() to reinitialize "
```
mulkieran: warning should use "bd_try_init()" here.

vpodzime:
> I'm still puzzled why new third parameter of load_plugins is a number, which is a lot less information rich than a list or an array of the actually loaded plugins. It certainly has access to that information and I think user code could be a bit more direct with that API.

Since that info is not always needed and constructing the list of strings in C is not that easy, so is not comparing them as unordered lists, etc. The number of loaded plugins is enough internally as we only check it for being 0 or something else.

src/lib/blockdev.c
vpodzime: Good catch, fixing locally.

mulkieran:
Ok, so this is what I think I know.

* The sole purpose of ```initialized``` is to help multiple threads to coordinate
their use of libblockdev.

* ```is_initialized``` evaluates to TRUE if:
  1. all plugins have been unloaded explicitly
  2. at lease one plugin has been loaded

  This is pretty much equivalent to "some thread has had an impact on
  libblockdev global state since it was loaded", which feels like a reasonable
  meaning for ```initialized```.

* I could build a non-standard plugin by making a ```PluginSpec``` and then setting its ```name``` attribute to a recognized name and its ```so_name``` to my special so name. Then I could load it
using any of the *init() methods as appropriate.

Things I would like:

* Extended documentation on ```is_initialized```, since it can be affected by
calls to ```reinit()``` and ```try_reinit()``` as well as the others.

What I don't really understand:

* Why do we care in blivet if blockdev is initialized?
Doesn't it make just as much sense to just call a ```reinit()``` method without checking ```is_initialized```, like below.

```
try:
    succ, loaded = blockdev.try_reinit(require_plugins=_requested_plugins, reload=False, log_func=log_bd_message)
except GLib.GError:
    raise some exception # there was an actual failure here

missing_plugs = _REQUESTED_PLUGIN_NAMES - set(loaded)
for p in missing_plugs:
    log.info(...)
```

My assumption is that all blivet cares about is loading as many of the plugins that it thinks it needs as it can.

Let's assume there was another thread running. How could code like the above change observable state for that thread? The only thing I could think of is that it will:
* see more loaded plugins than formerly OR
* find that the non-standard plugin that it loaded previously has been replaced by the standard
version, requested by blivet

vpodzime:
> Things I would like:
> 
>     Extended documentation on is_initialized, since it can be affected by calls to reinit() and try_reinit() as well as the others.

Fixed documentation locally.

vpodzime:
And as for the changes in blivet:

Yes, you're right, that's what we should change the libblockdev initialization in blivet to with libblockdev-1.0. AFAICT, a non-standard plugin being replaced by the standard version might be an issue, but only in really specific cases we are not encountering (yet). Let's deal with such situation if we have to at some point.

vpodzime:
Thanks for the input, @mulkieran. If I see it correctly, there's nothing more to be addressed with this change in relation to the version 1.0. So let's merge this and continue the discussion in a new issue/PR if desired.

PR #20, opened 2015-05-20
Remove private macros from the gtkdoc sections file.

dashea:
This was the easy gtkdoc warning to fix.


vpodzime:
> This was the easy gtkdoc warning to fix.

True. Thanks!

PR #21, opened 2015-05-22
Add overrides for the lvm_is_valid_thpool_chunk_size() function

vpodzime:
Its 'discard' parameter should default to 'False' as that covers majority of use
cases.

This is an alternative solution for an issue targetted by (blivet's PR 121) [https://github.com/rhinstaller/blivet/pull/121]

### DIFF
--- a/src/python/gi/overrides/BlockDev.py
+++ b/src/python/gi/overrides/BlockDev.py
@@ -226,6 +226,12 @@ def lvm_thsnapshotcreate(vg_name, origin_name, snapshot_name, pool_name=None):
     return _lvm_thsnapshotcreate(vg_name, origin_name, snapshot_name, pool_name)
 __all__.append("lvm_thsnapshotcreate")
 
+_lvm_is_valid_thpool_chunk_size = BlockDev.lvm_is_valid_thpool_chunk_size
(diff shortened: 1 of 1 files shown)


cgwalters:
Looks OK to me.  :+1: 

This will fix existing Blivet instances which is good.

Though I will say as one of the upstream g-i people, I am of the group that discourages overrides.  In this case libblockdev is itself wrapped by Blivet, so the number of Python callers here is quite small.  I'd agreed overrides were necessary when you had tens/hundreds+ of apps, but this case they seem like overkill.

Regardless though it will fix the bug, so :+1: from me on the content.

vpodzime:
> Though I will say as one of the upstream g-i people, I am of the group that discourages overrides. In this case libblockdev is itself wrapped by Blivet, so the number of Python callers here is quite small. I'd agreed overrides were necessary when you had tens/hundreds+ of apps, but this case they seem like overkill.

Well, as one of the people using g-i, I'd be really grateful if the proposed ``default=X`` annotation option was made reality. ;) Other than that I don't see why overrides could be a problem or overkill. It's just few lines of code per function that make the use from Python much nicer and natural.

Thanks for the ACK.

PR #22, opened 2015-06-10
Escrow tests and some bugs I found running the escrow tests

dashea:


vpodzime:
First of all big THANKS for these tests! Could you please add the extra packages required for running these tests as comments under ``BuildRequires`` to the ``dist/libblockdev.spec.in`` file?

vpodzime:
Thanks!

Issue #23, opened 2015-06-10
the libblockdev package requires python2


dashea:
And I guess that's a problem in the bright new tomorrow. I guess the python2 and python3 gi/overrides files need to go in separate packages. I don't know what all that will do to packages that depend on libblockdev, or if one of the overrides should stay in the main libblockdev package, or what. Unsurprisingly there's no guidance from the Fedora guidelines for things like this.

dashea:
Another idea: maybe the leave the files as-is and override the requires to not require python? Anyone using the python bindings is going to require python from their end anyway, and it seems nicer to just have all the gobject typelib and overrides and everything all in one package instead of creating one or two more packages for a single file

vpodzime:
*Gtk*, *GLib*, *Pango* and some other overrides are included in the ``pygobject3-base``/``python3-gobject`` packages which seems weird to me because either these two packages require *Gtk*, *GLib*, etc. or they provide overrides for something that may not be installed.

Providing subpackages with just the overrides looks like an overkill to me. So I hope just dropping the dependencies will work.

vpodzime:
Looking at the spec file now, I only see ``BuildRequires`` for ``python-devel`` and ``python3-devel``. So the problem is that *RPM* adds the requirements on its own.

```
$ rpm -qp --requires ~/rpmbuild/RPMS/x86_64/libblockdev-1.0-1.fc21.x86_64.rpm 
/sbin/ldconfig
/sbin/ldconfig
libbd_utils.so.0()(64bit)
libc.so.6()(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libglib-2.0.so.0()(64bit)
python(abi) = 2.7
python(abi) = 3.4
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
rtld(GNU_HASH)
```

dashea:
Right, I was thinking the spec file could add something like this:

```
diff --git a/dist/libblockdev.spec.in b/dist/libblockdev.spec.in
index f1f4e37..bb8bfcb 100644
--- a/dist/libblockdev.spec.in
+++ b/dist/libblockdev.spec.in
[...]

PR #25, opened 2015-06-11
Filter the python files from automatic rpm requires

dashea:
The libblockdev python bindings don't really require python. Programs
using libblockdev via the python bindings require python. Remove the
gi.overrides files from autorequires so that the main libblockdev
package can contain all it needs for gobject-introspection bindings
without adding python-abi dependencies.


vpodzime:
Looks good to me and seems to work great. Thanks!

vpodzime:
This fixes issue #23

PR #30, opened 2015-06-17
Makefile updates

atodorov:
Makefile.am:
  - add separate check target - previously this was repeated in all the test targets
  - add coverage targets - makes it possible to use the Python coverage tool to generate reports
  - make it possible to test with Python3 - just use variables instead of binary names so it is possible to override them on the command line.

Add description of these new Makefile targets to the docs as well.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,26 +12,39 @@ MAINTAINERCLEANFILES = Makefile.in aclocal.m4 config.guess config.sub \
 LIBDIRS = src/utils/.libs:src/plugins/.libs:src/lib/.libs
 GIDIR = src/lib
 
+PYTHON ?= python2
+COVERAGE ?= coverage
+
 run-ipython: all
 	GI_TYPELIB_PATH=${GIDIR} LD_LIBRARY_PATH=${LIBDIRS} PYTHONPATH=src/python G_MESSAGES_DEBUG=all ipython
 
 run-root-ipython: all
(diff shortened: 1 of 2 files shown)


mulkieran:
Is it possible to make use of csmock as part of check target?

atodorov:
Do you mean csmock-plugin-pylint ? I need to look into this b/c I'm not familiar with it.

mulkieran:
I'm not familiar w/ it, either. I remember hearing about it a while back, but I hadn't heard that it was a reality. I was thinking about csmock for the C source since that is most of the project. It would be kind of neat if you could run csmock with all plugins enabled and get pylint, C, and shell checking all at once, but I feel like that is a lot to expect and probably doesn't happen.

Makefile.am
```diff
@@ -12,26 +12,39 @@ MAINTAINERCLEANFILES = Makefile.in aclocal.m4 config.guess config.sub \
 LIBDIRS = src/utils/.libs:src/plugins/.libs:src/lib/.libs
 GIDIR = src/lib
 
+PYTHON=python2
```
vpodzime: I think this should be ``PYTHON ?= python2``

vpodzime:
Other than the comment above this looks good to me. Please create a seprate issue for the discussion about ``csmock``

Makefile.am
atodorov: Updated. Same goes for COVERAGE ?= coverage

vpodzime:
Thanks!

PR #33, opened 2015-07-22
Add a new makefile target that does everything needed for jenkins.

clumens:
We should have as much of the logic of how the CI tests are run in source
control as possible, so that's what this target is for.  Besides this, jenkins
just runs a "git clean" first.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,3 +83,5 @@ rpm: local
 
 release: tag
 	$(MAKE) archive
+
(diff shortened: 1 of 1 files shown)


clumens:
I think those failures from jenkins are unrelated to this patch.

vpodzime:
> I think those failures from jenkins are unrelated to this patch.

Agreed. This looks good to me, thanks!

PR #36, opened 2015-08-11
BTRFS: allow an arbitrary label to be set for a btrfs volume

phatina:
Allow libblockdev to read info about filesystems with more complex labels.

This suggestion can't read multi-line labels, though; but is sufficient for any arbitrary one-line label.


vpodzime:
Looks good to me, thanks!

PR #38, opened 2015-08-31
Better document how MD RAID superblock size should be calculated

vpodzime:
The superblock size should be calculated from an MD RAID member size not from
the size of the whole array. Make this clear in the bd_md_get_superblock_size()
function's documentation and size parameter name.


snbueno:
This change looks fine.

PR #47, opened 2015-10-13
Produce coverage report in CI

atodorov:
Minor change but when I come to think about it we can do better. 

At the moment the report only includes `src/python/gi/overrides/BlockDev.py`. Maybe we can combine coverage.py and gcov to report on the C files as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .*.swp
+.coverage
+coverage-report.log
 **/*.pyc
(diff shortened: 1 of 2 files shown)


vpodzime:
> At the moment the report only includes src/python/gi/overrides/BlockDev.py. Maybe we can combine coverage.py and gcov to report on the C files as well.

Yeah, that would be great. Even this is a nice/good step forward. Thanks!

PR #48, opened 2015-10-16
Add pv_free to BDLVMPVdata

vojtechtrefny:


dwlehman:
I don't object, but I'm curious what this is for. It seems like this is going to be a required piece of information if we ever add support for user-specified pv subsets for new lvs.

vpodzime:
Exactly. And for LVM RAID, also for LVM cache support in blivet-gui and also to prevent issues with calling ``pvmove()`` on PVs with no data. Generally a useful piece of information, I think.

vpodzime:
A test for this new info would be nice, but considering the fact how ``pvinfo()`` and ``pvs()`` tests look like right now I think we can deal with all that together at some point in the future.

vpodzime:
Looks good to me otherwise.

PR #51, opened 2015-11-12
Produce test execution logs

atodorov:
This PR produces pylint.log and a test-suite-python*.log. These are only the test results without all the build stuff from the console. Easier to read.


vpodzime:
Looks good to me, thanks!

PR #57, opened 2016-02-12
Fix error reporting when running "make test".

clumens:
That target pipes stuff to tee, which means the error code that gets
reported is the error code of tee.  Thus if a test fails, we will still
see success and keep going.  The fix is to simply use bash as the shell
and enable pipefail on it.


vpodzime:
Nice! Thanks a lot! Since the tests on this PR are correctly reported as failed (due to an unrelated problem) it seems like this works.

PR #69, opened 2016-03-15
Trigger LVM DBus autostart

vpodzime:
The LVM DBus service may be available, but not running. However, we can easily trigger its autostart.


vojtechtrefny:
Looks good to me.

PR #70, opened 2016-03-15
Add bd_md_create_with_chunk_size() function

vojtechtrefny:


tests/md_test.py
```diff
@@ -157,6 +157,27 @@ def test_create_deactivate_destroy(self):
         succ = BlockDev.md_destroy(self.loop_dev3)
         self.assertTrue(succ)
 
+class MDTestCreateWithChunkSize(MDTestCase):
+    @unittest.skipIf("SKIP_SLOW" in os.environ, "skipping slow tests")
+    def test_create_deactivate_destroy(self):
```
vpodzime: This should mention the chunk size too.

src/plugins/mdraid.c
```diff
@@ -402,6 +402,88 @@ gboolean bd_md_create (gchar *device_name, gchar *level, gchar **disks, guint64
...
+ * @version: (allow-none): metadata version
+ * @bitmap: whether to create an internal bitmap on the device or not
+ * @chunk_size: chunk size of the device to create
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the new MD RAID device @device_name was successfully created or not
+ */
+gboolean bd_md_create_with_chunk_size (gchar *device_name, gchar *level, gchar **disks, guint64 spares, gchar *version, gboolean bitmap, guint64 chunk_size, GError **error) {
+    gchar **argv = NULL;
+    /* ["mdadm", "create", device, "--run", "level", "raid-devices",...] */
```
vpodzime: Curly brackets would be better here as that's how static arrays are defined in C.

vpodzime:
Other than the two nitpicks above this looks good to me.

vojtechtrefny:
Fixed failing tests and changes based on @vpodzime comments.

vpodzime:
Talked to @vojtechtrefny that we should probably add the ``chunk_size`` to the result of ``md_examine()``.

vpodzime:
> Talked to @vojtechtrefny that we should probably add the chunk_size to the result of md_examine().

Talked to @vojtechtrefny a bit more and we decided to leave this for a separate PR.

PR #72, opened 2016-03-18
Make the tests run faster and a bit better

vpodzime:
We don't really need to run the tests with both Python 2 and Python 3. Also, we don't need to run gtk-doc just to recompile a plugin. And we can try harder to clean after ``mdadm``.


dwlehman:
Looks good to me.

PR #73, opened 2016-03-18
First pieces of the 'part' plugin

vpodzime:
This PR covers the current requirements for the *part* plugin for its use by *storaged*. This is the squashed version of the patches I have on my [devel branch](https://github.com/vpodzime/libblockdev/tree/master-part). Feel free to check that branch for some more detailed information about stuff.


vpodzime:
Interesting, new gcc in Rawhide is more strict. Nice.

src/python/gi/overrides/BlockDev.py
```diff
@@ -294,6 +295,13 @@ def kbd_zram_create_devices(num_devices, sizes, nstreams=None):
 __all__.append("kbd_zram_create_devices")
 
 
+_part_create_table = BlockDev.part_create_table
+@override(BlockDev.part_create_table)
+def part_create_table(disk, type, ignore_existing=True):
+    return _part_create_table(disk, type, ignore_existing)
+__all__.append("part_create_table")
+
```
dwlehman: What determines which functions get overrides?

vpodzime: Those that can make use of python-specific features like default values for parameters or dynamic typing and conversion of the arguments. Thinking about it here, this function could probably be able to take ``type`` as a string and convert it to the BlockDev.PartTableType enum.

dwlehman:
Looks okay to me.

PR #74, opened 2016-03-22
First pieces of the 'fs' plugin

vpodzime:
Support for the *ext4* file system.


src/plugins/fs.c
```diff
@@ -0,0 +1,484 @@
...
+BDFSExt4Info* bd_fs_ext4_info_copy (BDFSExt4Info *data) {
+    BDFSExt4Info *ret = g_new0 (BDFSExt4Info, 1);
+
+    ret->label = g_strdup (data->label);
+    ret->uuid = g_strdup (data->uuid);
+    ret->state = g_strdup (data->state);
+    ret->block_size = data->block_size;
+    ret->block_count = data->block_count;
+    ret->free_blocks = data->free_blocks;
+    ret->mount_count = data->mount_count;
```
dwlehman: Why include mount_count if not including maximal mount count or whatever it's called?

src/plugins/fs.c
```diff
@@ -0,0 +1,484 @@
...
+ *          successfully set or not
+ */
+gboolean bd_fs_ext4_set_label (gchar *device, gchar *label, GError **error) {
+    gchar *args[5] = {"tune2fs", "-L", label, device, NULL};
+
+    return bd_utils_exec_and_report_error (args, error);
+}
+
+/**
+ * parse_mdadm_vars: (skip)
```
dwlehman: Name needs updating here.

dwlehman:
Looks good except for the question about including `mount_count` without also including `maximal_mount_count`.

Also, I noticed that you're not making things like `device` function argument `const`. It seems like it would be worthwhile to get in the habit of doing that when adding new code like this.

src/plugins/fs.c
vpodzime: I think it is a neat way to find out if the file system is mounted. And in tools like blivet-gui or gnome-disks, this is often shown to the user together with the other info about the file system.

src/plugins/fs.c
vpodzime: Good catch, thanks! Fixing.

vpodzime:
> Also, I noticed that you're not making things like device function argument const. It seems like it would be worthwhile to get in the habit of doing that when adding new code like this.

@vojtechtrefny had the same suggestion for the *part* plugins. The answer is that I want to keep the API of *libblockdev-1.x* consistent across the plugins. All the other plugins don't have ``const`` arguments. I think this should be changed all at once with *libblockdev-2.x*.

src/plugins/fs.c
dwlehman: Is this the field that's identified as `Mount count:` by dumpe2fs? If so, I don't think it means what you think it means. The mount count for my /boot is currently 12. It's used for enforcing an fsck every N mounts, where N is shown as `Maximum mount count`. Or is this `mount_count` something different?

vpodzime: Oh, you're exactly right. I'm dropping this field then.

vojtechtrefny:
Looks good to me.

PR #76, opened 2016-04-04
XFS support

vpodzime:
We already have support for the *ext4* file system. *xfs* is the other one that is most commonly used, so it should be supported too.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -65,6 +65,55 @@ GType bd_fs_ext4_info_get_type () {
     return type;
 }
 
+#define BD_FS_TYPE_XFS_INFO (bd_fs_xfs_info_get_type ())
+GType bd_fs_xfs_info_get_type();
+
+typedef struct BDFSXfsInfo {
+    gchar *label;
+    gchar *uuid;
+    guint64 block_size;
+    guint64 block_count;
+} BDFSXfsInfo;
+
+/**
+ * bd_fs_xfs_info_copy: (skip)
+ *
+ * Creates a new copy of @data.
+ */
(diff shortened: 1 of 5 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -147,4 +196,77 @@ BDFSExt4Info* bd_fs_ext4_get_info (gchar *device, GError **error);
...
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether an xfs signature was successfully wiped from the @device or
+ *          not
+ */
+gboolean bd_fs_xfs_wipe (gchar *device, GError **error);
+
+/**
+ * bd_fs_xfs_check:
+ * @device: the device the file system on which to check
```
dwlehman: Something like "the device containing the file system to check" would be clearer.

src/lib/plugin_apis/fs.api
```diff
@@ -147,4 +196,77 @@ BDFSExt4Info* bd_fs_ext4_get_info (gchar *device, GError **error);
...
+ * Returns: whether an xfs file system on the @device is clean or not
+ *
+ * Note: if the file system is mounted it may be reported as unclean even if
+ *       everything is okay and there are just some pending/in-progress writes
+ */
+gboolean bd_fs_xfs_check (gchar *device, GError **error);
+
+/**
+ * bd_fs_xfs_repair:
+ * @device: the device the file system on which to repair
```
dwlehman: same here as above ("the device containing the file system to repair")

dwlehman:
Aside from the wording of several comments (all along the same lines in the same file) this looks good to me.

vpodzime:
Fixed the wording. I'm going to wait for the check to run and finish and then merge this if everything looks okay.

PR #77, opened 2016-04-05
Don't pass sizes in bytes to LVM (#1317373)

vpodzime:
LVM does preliminary checks on the given size and in case it's not a multiple of
512, it exits with an error never reaching the code that would actually round
the given size to usually a multiple of extent size. Let's use KiB ("K" in LVM's
twisted notation) instead.

### DIFF
--- a/dist/libblockdev.spec.in
+++ b/dist/libblockdev.spec.in
@@ -21,6 +21,7 @@ BuildRequires: glib2-doc
 BuildRequires: kmod-devel
 BuildRequires: parted-devel
 BuildRequires: libblkid-devel
+BuildRequires: libbytesize-devel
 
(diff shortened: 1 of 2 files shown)


dwlehman:
Does lvm generally round down like you're doing? Does it matter?

vpodzime:
> Does lvm generally round down like you're doing? Does it matter?

Not generally, I'd say. Usually it's to the number of extents (which happens afterwards). Other than that I don't know, because LVM don't accept things that are not multiple of 512 bytes so there's no rounding. However, I think we are safe to round down to KiB. All the inputs from users are going to be in KiBs or some bigger units anyway. And results of computations hardly ever expect things to be rounded up (there might be space missing later).

dwlehman:
Works for me.

PR #79, opened 2016-04-14
Support for extra arguments

vpodzime:
For now only for functions that can actually do something with them. Before the time for ``libblockdev-2.0`` comes we will decide whether all functions should take extra arguments or not.


dwlehman:
Looks good to me.

vpodzime:
Added the support for extra arguments and while doing so I realized that the ``bcache_create`` function had the arguments in a wrong order so I fixed that too.

dwlehman:
Still looks good to me.

vpodzime:
Added one little improvement to the code transforming the ``**kwargs`` so that ``None`` is passed instead of ``[]`` if there are none extra arguments nor ``**kwargs`` specified.

Issue #56, opened 2016-02-12
Please declare the function arguments with const where appropriate


tsmetana:
Hello.
There are many functions taking arguments of the gchar* type. I suspect they don't actually want to modify the strings. It would be nicer/better to change those arguments to const gchar*. It would feel "safer" to use them.

Thanks.

PR #83, opened 2016-04-27
Mark pointer parameters we don't change as const

vpodzime:
Otherwise the caller cannot tell if we are going to change the referenced
value/data or not.

Note: Requested in the issue #56

### DIFF
--- a/scripts/boilerplate_generator.py
+++ b/scripts/boilerplate_generator.py
@@ -207,7 +207,7 @@ def get_funcs_info(fn_infos, module_name):
 def get_loading_func(fn_infos, module_name):
     # TODO: only error on functions provided by the plugin that fail to load
     # TODO: implement the 'gchar **errors' argument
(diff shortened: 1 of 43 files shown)


vpodzime:
A tip for doing the review of these changes - it might be best to just go through the [plugins' header files](https://github.com/vpodzime/libblockdev/tree/master-const_params_pr/src/plugins) and check that all pointer parameters that you feel like they should be ``const`` are ``const``. The correctness of the rest of the changes should be assured by the fact that it all compiles and the tests pass.

vojtechtrefny:
Looks good to me.

snbueno:
Other than the two params in bd_kbd_zram_create_devices() from src/plugins/kbd.h that we discussed in IRC (which should also be marked const), this looks fine to me as well.

vpodzime:
> Other than the two params in bd_kbd_zram_create_devices() from src/plugins/kbd.h that we discussed in IRC (which should also be marked const), this looks fine to me as well.

Fixed, thanks!

Issue #82, opened 2016-04-26
optimal alignment creates partitions of weird sizes


vpodzime:
e.g. a 260 MiB partition instead of 2 MiB partition, etc.

vpodzime:
Should be fixed now.

PR #84, opened 2016-05-06
Optimal align fix

vpodzime:
All the real work is in the second commit which actually fixes the problems reported in the issue #82 and describes what the problem was.


dwlehman:
Looks good to me.

Keep in mind that `ped_alignment_align_up` can actually align down in some cases (generally when weird things are going on, eg: the free region is too small for one alignment grain).

vpodzime:
> Keep in mind that ped_alignment_align_up can actually align down in some cases (generally when weird things are going on, eg: the sector isn't initially aligned and the free region is too small for one alignment grain).

Making note, thanks!

Issue #81, opened 2016-04-21
RFE: Allow to set multiple flags for a partition at the same time


phatina:
Please, provide an API which would allow Storaged to set multiple partition flags at one commit. For MSDOS and GPT partition tables, of course.

vpodzime:
Should be fixed now in master.

vpodzime:
The full(er) GPT support will be added as part of #80

PR #86, opened 2016-05-09
Add a function for setting multiple partition flags at once

vpodzime:
Not sure how often multiple partition flags are set, but if it happens, it's
better to do it in a single function call with a single commit to disk. This
also brings some atomicity to the operation as if (un)setting one flag fails,
nothing is commited to the disk.

Requested in the issue #81.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -227,7 +227,6 @@ BDPartSpec* bd_part_get_part_spec (const gchar *disk, const gchar *part, GError 
     if (!ped_disk) {
         set_parted_error (error, BD_PART_ERROR_FAIL);
         g_prefix_error (error, "Failed to read partition table on device '%s'", disk);
-        ped_disk_destroy (ped_disk);
         ped_device_destroy (dev);
         return FALSE;
     }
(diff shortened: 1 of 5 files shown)


src/plugins/part.c
```diff
@@ -666,6 +666,100 @@ gboolean bd_part_set_part_flag (const gchar *disk, const gchar *part, BDPartFlag
...
+        set_parted_error (error, BD_PART_ERROR_INVAL);
+        g_prefix_error (error, "Device '%s' invalid or not existing", disk);
+        return FALSE;
+    }
+
+    ped_disk = ped_disk_new (dev);
+    if (!ped_disk) {
+        set_parted_error (error, BD_PART_ERROR_FAIL);
+        g_prefix_error (error, "Failed to read partition table on device '%s'", disk);
+        ped_disk_destroy (ped_disk);
```
snbueno: Is it necessary to call `ped_disk_destroy (ped_disk);` here since ped_disk is NULL?

snbueno:
Other than my one question, this looks good to me.

vpodzime:
> Other than my one question, this looks good to me.

It isn't, good catch! I found a couple more places with the same issue so I'm adding one more commit fixing all of those.

PR #88, opened 2016-05-16
Tips from blivet

vpodzime:
These commits implement extra features identified as useful by [analysis of blivet and its use of parted](https://gist.github.com/vpodzime/5ad01d65f76551016d4f0bd98c3060d5) .

The first three patches can be ignored as they are part of the PR #87

### DIFF
--- a/src/lib/plugin_apis/part.api
+++ b/src/lib/plugin_apis/part.api
@@ -248,6 +248,17 @@ gboolean bd_part_delete_part (const gchar *disk, const gchar *part, GError **err
 gboolean bd_part_set_part_flag (const gchar *disk, const gchar *part, BDPartFlag flag, gboolean state, GError **error);
 
 /**
+ * bd_part_set_disk_flag:
+ * @disk: disk the partition belongs to
+ * @flag: flag to set
+ * @state: state to set for the @flag (%TRUE = enabled)
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the flag @flag was successfully set on the @disk or not
+ */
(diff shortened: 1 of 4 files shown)


src/plugins/part.c
```diff
@@ -558,6 +558,78 @@ BDPartSpec** bd_part_get_disk_free_regions (const gchar *disk, GError **error) {
...
+ * @size: size of the partition to be added
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: (transfer full): spec of the best free region on @disk for a new partition of type @type
+ *                           with the size of @size or %NULL if there is none such region or if
+ *                           there was an error (@error gets populated)
+ *
+ * Note: For the @type %BD_PART_TYPE_NORMAL, the smallest possible space that *is not* in an extended partition
+ *       is found. For the @type %BD_PART_TYPE_LOGICAL, the smallest possible space that *is* in an extended
+ *       partition is found. For %BD_PART_TYPE_EXTENDED, the biggest possible space is found is long as there
```
vojtechtrefny: "found is long as" should be "as long as"

src/lib/plugin_apis/part.api
```diff
@@ -72,7 +72,7 @@ typedef struct BDPartSpec {
     gchar *path;
     gchar *name;
     gchar *type_guid;
-    BDPartType type;
+    guint64 type;
```
vojtechtrefny: You should probably mention this change in commit message.

vojtechtrefny:
Looks good to me otherwise.

PR #90, opened 2016-05-28
Zfcp functionality

snbueno:
Some pretty isolated and self-explanatory (I hope) changes.

One commit fixes a tiny amount of fall-out from commit b733870 by making sure to pass const parameters to blockdev utility functions.

The other commit adds zFCP functionality. Unfortunately, there is no real way to test these other than manually.

### DIFF
--- a/src/lib/plugin_apis/s390.api
+++ b/src/lib/plugin_apis/s390.api
@@ -78,3 +78,44 @@ gchar* bd_s390_zfcp_sanitize_wwpn_input (const gchar *wwpn, GError **error);
  * Returns: (transfer full): a synthesized zFCP LUN
  */
 gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error);
+
+/**
+ * bd_s390_zfcp_online:
+ * @devno a zFCP device number
+ * @wwpn a zFCP WWPN identifier
+ * @lun a zFCP LUN identifier
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a zFCP device was successfully switched online
+ */
+gboolean bd_s390_zfcp_online (const gchar *devno, const gchar *wwpn, const gchar *lun, GError **error);
+
+/*
+ * bd_s390_zfcp_scsi_offline:
+ * @devno a zFCP device number
+ * @wwpn a zFCP WWPN identifier
+ * @lun a zFCP LUN identifier
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a LUN was successfully removed from associated WWPN
+ *
+ * This function looks through /proc/scsi/scsi and manually removes LUNs from
+ * associated WWPNs. zFCP devices are SCSI devices accessible over FCP protocol.
+ * In z/OS the IODF (I/O definition file) contains basic information about the
+ * I/O config, but WWPN and LUN configuration is done at the OS level, hence
+ * this function becomes necessary when switching the device offline. This
+ * particular sequence of actions is for some reason unnecessary when switching
+ * the device online. Chalk it up to s390x being s390x.
+ */
+gboolean bd_s390_zfcp_scsi_offline(const gchar *devno, const gchar *wwpn, const gchar *lun, GError **error);
+
+/*
+ * bd_s390_zfcp_offline:
+ * @devno: zfcp device number
+ * @wwpn: zfcp WWPN (World Wide Port Number)
+ * @lun: zfcp LUN (Logical Unit Number)
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a zfcp device was successfully switched offline
+ */
+gboolean bd_s390_zfcp_offline (const gchar *devno, const gchar *wwpn, const gchar *lun, GError **error);
--- a/src/plugins/s390.c
+++ b/src/plugins/s390.c
@@ -71,7 +71,7 @@ gboolean check() {
 gboolean bd_s390_dasd_format (const gchar *dasd, const BDExtraArg **extra, GError **error) {
     gboolean rc = FALSE;
     gchar *dev = g_strdup_printf ("/dev/%s", dasd);
-    gchar *argv[] = {"/sbin/dasdfmt", "-y", "-d", "cdl", "-b", "4096", dev, NULL};
+    const gchar *argv[8] = {"/sbin/dasdfmt", "-y", "-d", "cdl", "-b", "4096", dev, NULL};
 
     rc = bd_utils_exec_and_report_error (argv, extra, error);
     g_free (dev);
@@ -136,7 +136,7 @@ gboolean bd_s390_dasd_online (const gchar *dasd, GError **error) {
     gint online = 0;
     gchar *path = NULL;
     FILE *fd = NULL;
-    gchar *argv[] = {"/usr/sbin/dasd_cio_free", "-d", dasd, NULL};
+    const gchar *argv[4] = {"/usr/sbin/dasd_cio_free", "-d", dasd, NULL};
 
     path = g_strdup_printf ("/sys/bus/ccw/drivers/dasd-eckd/%s/online", dasd);
     fd = fopen(path, "r+");
--- a/src/plugins/s390.c
+++ b/src/plugins/s390.c
@@ -18,6 +18,7 @@
  */
 
 #include <glib.h>
+#include <glob.h>
 #include <linux/fs.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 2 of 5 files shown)


src/lib/plugin_apis/s390.api
```diff
@@ -78,3 +78,36 @@ gchar* bd_s390_zfcp_sanitize_wwpn_input (const gchar *wwpn, GError **error);
...
+ * @wwpn a zFCP WWPN identifier
+ * @lun a zFCP LUN identifier
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a zFCP device was successfully switched online
+ */
+gboolean bd_s390_zfcp_online (const gchar *devno, const gchar *wwpn, const gchar *lun, GError **error);
+
+/*
+ * bd_s390_zfcp_scsi_online:
```
vpodzime: offline?

src/lib/plugin_apis/s390.api
```diff
@@ -78,3 +78,36 @@ gchar* bd_s390_zfcp_sanitize_wwpn_input (const gchar *wwpn, GError **error);
...
+ * @devno a zFCP device number
+ * @wwpn a zFCP WWPN identifier
+ * @lun a zFCP LUN identifier
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a LUNs were successfully removed from associated WWPN
```
vpodzime: Extra "a"? Or maybe a wrong sentence? I don't see how such parameters could result into multiple LUNs being taken offline. It would be nice to make this description a bit more verbose so that it makes more sense in relation to the name of the function.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    gchar *failed = g_strdup_printf ("%s/failed", unitdir);
+
+    /* part 01: make sure device is available/not on device ignore list */
+    fd = fopen (online, "r");
+    if (!fd) {
+        boolrc = bd_utils_exec_and_report_error (zfcp_cio_free, NULL, error);
+
+        if (!boolrc) {
+            fclose (fd);
+            g_free (online);
```
vpodzime: This should free a lot more strings. Maybe allocate them right before they are actually needed?

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        }
+        /* fd is NULL at this point, so try to open it again */
+        fd = fopen (online, "r");
+        g_free (online);
+
+        /* still no luck, fail */
+        if (!fd) {
+            g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                         "Could not open device %s even after removing it from" " the device ignore list.", devno);
+            return FALSE;
```
vpodzime: Same here.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    if (rc == 1) {
+        /*
+         * otherwise device's status indicates that it's arleady online, so
+         * just close the fd and proceed; we don't return because although 'online'
+         * status may be correct, the device may not be completely online and ready
+         * to use just yet
+         */
+        fclose (fd);
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                     "Device %s is already online", devno);
```
vpodzime: So this function may return ``TRUE`` and an error at the same time? That's not really nice and it would result in an exception being raised in Python for a successful run of the function.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    /* part 04: other error checking to verify device turned on properly */
+    fd = fopen (failed, "r");
+    if (!fd) {
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_IO,
+                     "Could not open %s", failed);
+        return FALSE;
+    }
+
+    rc = fgetc (fd);
+    rc -= '0';
```
vpodzime: I think this needs to be done after the check for ``EOF``.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        fclose (fd);
+        return FALSE;
+    }
+    if (rc != 0) {
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                     "Failed LUN %s at WWPN %s on zFCP device %s removed again", lun, wwpn, devno);
+        fclose (fd);
+        return FALSE;
+    }
+    /* if you haven't failed yet, you deserve this */
```
vpodzime: A lot of strings not being freed here.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+            continue;
+        }
+
+        /* tokenize line and assign certain values we'll need later */
+        tokens = g_strsplit (line, delim, 8);
+
+        host = tokens[1];
+        fcplun = tokens[7];
+
+        scsidev = g_strdup_printf ("%s:%s:%s:%s", &host[4], channel, devid, fcplun);
```
vpodzime: ``host + 4`` may be better readable here.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        /* get HBA path value (same as device number) */
+        hba_path = g_strdup_printf ("%s/hba_id", fcpsysfs);
+        len = 0; /* should be zero, but re-set it just in case */
+        fd = fopen (hba_path, "r");
+        rc = getline (&fcphbasysfs, &len, fd);
+        if (rc == -1) {
+            g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                         "Failed to read value from %s", hba_path);
+            fclose (fd);
+            g_free (hba_path);
```
vpodzime: A lot of strings not being freed here.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        wwpn_path = g_strdup_printf ("%s/wwpn", fcpsysfs);
+        len = 0;
+        fd = fopen (wwpn_path, "r");
+        rc = getline (&fcpwwpnsysfs, &len, fd);
+        if (rc == -1) {
+            g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                         "Failed to read value from %s\n", wwpn_path);
+            g_free (wwpn_path);
+            fclose (fd);
+            return FALSE;
```
vpodzime: Again many strings need to be freed here incl. ``fcpwwpnsysfs`` (see *man getline*)

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        g_free (hba_path);
+
+        /* get WWPN value */
+        wwpn_path = g_strdup_printf ("%s/wwpn", fcpsysfs);
+        len = 0;
+        fd = fopen (wwpn_path, "r");
+        rc = getline (&fcpwwpnsysfs, &len, fd);
+        if (rc == -1) {
+            g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                         "Failed to read value from %s\n", wwpn_path);
```
vpodzime: No need for a newline at the end of the error message (applies to some more places below too).

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    gchar *offline = g_strdup_printf ("%s/%s/online", zfcpsysfs, devno);
+    gchar *unitrm = g_strdup_printf ("%s/%s/%s/unit_remove", zfcpsysfs, devno, wwpn);
+    gchar *pattern = g_strdup_printf ("%s/0x??????????????\?\?/0x????????????????", zfcpsysfs);
+    const gchar *chccwdev[4] = {"/usr/sbin/chccwdev", "-d", devno, NULL};
+
+    failed = bd_s390_zfcp_scsi_offline(devno, wwpn, lun, error);
+    if (failed == 0) {
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                     "Could not correctly delete SCSI device of zFCP %s with WWPN %s, LUN %s\n", devno, wwpn, lun);
+        return FALSE;
```
vpodzime: No strings being freed here and in multiple places below.

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    fclose (fd);
+    g_free (unitrm);
+    rc = 0;
+
+    /* gather the luns  */
+    rc = glob (pattern, GLOB_ONLYDIR, NULL, &luns);
+    if (rc == GLOB_ABORTED || rc == GLOB_NOSPACE) {
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                     "An error occurred trying to determine if other LUNs are still associated with WWPN %s\n", wwpn);
+        globfree (&luns);
```
vpodzime: ``return FALSE;`` missing here? Plus a few ``g_free()`` calls? ;-)

src/plugins/s390.c
```diff
@@ -398,3 +399,353 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        return FALSE;
+    }
+    globfree (&luns);
+    rc = 0;
+
+    /* offline */
+    rc = bd_utils_exec_and_report_error (chccwdev, NULL, error);
+    g_free (offline);
+    if (rc == EOF) {
```
vpodzime: I'm quite sure the exec function never returns ``EOF`` except for the stupidity of *C* having thousands names for 0.

src/plugins/s390.h
```diff
@@ -1,4 +1,5 @@
 #include <glib.h>
+#include <glob.h>
```
vpodzime: This is not needed here as nothing from ``glob.h`` is part of the API.

vpodzime:
Other than the above comments (which usually apply to a few more places) this in general looks good to me. Thanks for dealing with this magic. :)

src/plugins/s390.c
snbueno: Yeah, I like that idea.

src/lib/plugin_apis/s390.api
snbueno: I'm not sure what I was thinking, but I obviously couldn't English very well. :P I will expand the description as well.

src/plugins/s390.c
snbueno: Hm, actually it doesn't return TRUE; it proceeds. The reason for that is while "online" may be set to the correct value, the other values may still need to be set properly (in steps 03 and 04).

snbueno: I guess it might be better if I could set a warning, to be honest.

snbueno: Crap, I see what you mean now. I will find you in IRC tomorrow.

snbueno: FWIW, I realized g_warning() exists, so I used that instead since it's more appropriate.

vpodzime: Nice. :)

vpodzime:
Jenkins, please test this.

src/plugins/s390.c
```diff
@@ -398,3 +399,399 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    }
+
+    /* part 02: check to make sure/turn device online */
+    rc = fgetc (fd);
+    if (rc == EOF) {
+        /* there was some error checking the 'online' status */
+        g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_IO,
+                     "Error checking if device %s is online", devno);
+        fclose (fd);
+        g_free (online);
```
vpodzime: I think ``online`` can be freed here second time in a weird case when it's also freed above after the ``fopen()`` retry.

src/plugins/s390.c
```diff
@@ -398,3 +399,399 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+         * to use just yet, so just throw a warning.
+         */
+        fclose (fd);
+        g_warning ("Device %s is already online", devno);
+    }
+    else {
+        /* offline */
+        fclose (fd);
+        boolrc = bd_utils_exec_and_report_error (chccwdev, NULL, error);
+        if (boolrc == 0) {
```
vpodzime: ``if (!boolrc)``

src/plugins/s390.c
```diff
@@ -398,3 +399,399 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        g_free (online);
+
+        /* still no luck, fail */
+        if (!fd) {
+            g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
+                         "Could not open device %s even after removing it from" " the device ignore list.", devno);
+            return FALSE;
+        }
+    }
+
```
vpodzime: AFAICT ``online`` is not used below this line so it should be freed here.

src/plugins/s390.c
```diff
@@ -398,3 +399,399 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+        return FALSE;
+    }
+    /* if you haven't failed yet, you deserve this */
+    g_free (failed);
+    g_free (portdir);
+    fclose (fd);
+    return TRUE;
+}
+
+/*
```
vpodzime: This should be ``/**``.

src/plugins/s390.c
```diff
@@ -398,3 +399,399 @@ gchar* bd_s390_zfcp_sanitize_lun_input (const gchar *lun, GError **error) {
...
+    fclose (scsifd);
+    g_free (scsidel);
+    g_free (fcplunsysfs);
+    g_free (fcpwwpnsysfs);
+    g_free (fcphbasysfs);
+    g_free (scsidev);
+    return TRUE;
+}
+
+/*
```
vpodzime: Same here.

vpodzime:
Looks good to me other than the above minor things.

src/plugins/s390.c
snbueno: Hmm, it is freed though (line 440). Or am I missing something?

src/plugins/s390.c
snbueno: Took me a moment to understand what you meant, but I see now. So I'll certainly fix this, just leaving the `g_free()` call here where you've pointed it out, but I'll only call `g_free()` after the `fopen()` retry if there's an error, which should be perfectly fine.

snbueno:
Vratislav, just the one question I had about the instance of `online` you suggested needs to be freed, and then I'll be ready to push this. Thanks for the thorough review as well!

src/plugins/s390.c
vpodzime: > Hmm, it is freed though (line 440). Or am I missing something?

In the current code, ``online`` is allocated on line 421 then it's used on lines 427 and 439. After that it's not used anywhere so it can be freed right after the big ``if`` on line 449. That way it wouldn't have to be freed on lines 457, 477 and 481.

snbueno: Fixed.

PR #92, opened 2016-06-03
Extra LVM info

vpodzime:
The second commit adds some extra information to the ``LVMLVdata`` structure returned by the ``lvinfo()`` and ``lvs()`` functions. The existing tests should pass, but some more should probably be added which I'll do later before I push this.


vojtechtrefny:
Looks good to me.

snbueno:
Looks good to me as well.

PR #94, opened 2016-06-13
Include the LV roles in the LVdata

vpodzime:
This is a very useful piece of information when trying to determine what the
relations between LVs are together with the roles of such LVs in those
relations.


vpodzime:
The tests are failing due to the support missing in the LVM DBus API available on the jenkins slaves. A new version should be available soon (already built for rawhide).

vojtechtrefny:
Looks good to me.

PR #95, opened 2016-06-14
Documentation improvements

vpodzime:
Documentation for structs and a few missing items added to the right places.

### DIFF
--- a/src/lib/plugin_apis/mdraid.api
+++ b/src/lib/plugin_apis/mdraid.api
@@ -20,6 +20,20 @@ typedef enum {
 #define BD_MD_TYPE_EXAMINEDATA (bd_md_examine_data_get_type ())
 GType bd_md_examine_data_get_type();
 
+/**
+ * BDMDExamineData:
+ * @device: path of the MD device
+ * @level: RAID level of the device
+ * @num_devices: number of devices used by the MD device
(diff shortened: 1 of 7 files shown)


vpodzime:
The tests will probably fail due to some changes in the new kernel (more specifically in the *zram* interface) installed to the jenkins slaves.

src/lib/plugin_apis/mdraid.api
```diff
@@ -20,6 +20,20 @@ typedef enum {
...
+/**
+ * BDMDExamineData:
+ * @device: path of the MD device
+ * @level: RAID level of the device
+ * @num_devices: number of devices used by the MD device
+ * @name: name of the MD device
+ * @size: size of the MD device
+ * @uuid: array UUID
+ * @update_time: update time of the MD device
+ * @dev_uuid: UUID of the MD device
```
dwlehman: You could say something like `UUID of the member device` here to make it a bit clearer.

dwlehman:
Looks good to me other than the one possible language refinement I proposed.

src/lib/plugin_apis/mdraid.api
vpodzime: Good point, thanks!

PR #97, opened 2016-07-04
Add functions for creating thin/cache pools from existing LVs

vpodzime:
These allow for a multi-step creation of thin/cache pools with much better
control of their data/metadata parts' features.


vojtechtrefny:
Looks good to me.

PR #98, opened 2016-07-22
Progress reporting generics

vpodzime:
These three patches add the generic framework for progress reporting based on extracting progress information from an exec`ed process or from the LVM DBus API's *Job* objects (doesn't work properly right now, though). Plus a proof-of-concept implementation for the ``pvmove()`` function.


vpodzime:
The test failure is caused by a bug in the new jenkins slave's kernel.

vojtechtrefny:
Looks good to me.

PR #100, opened 2016-08-02
Progress reporting - the 'swap' plugin

vpodzime:
This is the first plugin to get the progress reporting where applicable so it of course revealed many small issues that needed fixing first. That's what the first 3 patches do, the last one adds some rudimentary progress reporting to the *swapon()* function which is the only one containing multiple steps allowing us to report at least some progress.


vpodzime:
The tests are likely to fail due to changes in the LVM's CLI behavior.

vojtechtrefny:
Looks good to me.

PR #102, opened 2016-08-05
Add progress reporting to the LVM plugins

vpodzime:
The only operation that LVM CLI provides some progress for is 'pvmove' which we
already have covered (the recent proof-of-concept). For other operations we have
no idea so we can just report start and finish which is done by our exec*
functions being used. The only exception are two cache-related functions which
perform multiple steps/operations allowing us to report progress based on that.

The DBus API may or may not provide progress for any operation (method call) and
we are prepared to handle that. So again nothing to do here except for the two
cache-related functions which can use the same trick as mentioned above.

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -2368,14 +2368,24 @@ gboolean bd_lvm_cache_create_pool (const gchar *vg_name, const gchar *pool_name,
     gchar *lv_id = NULL;
     gchar *lv_obj_path = NULL;
     const gchar *mode_str = NULL;
+    gchar *msg = NULL;
+    guint64 progress_id = 0;
+
+    msg = g_strdup_printf ("Started 'create cache pool %s/%s'", vg_name, pool_name);
+    progress_id = bd_utils_report_started (msg);
+    g_free (msg);
 
     /* create an LV for the pool */
     type = get_lv_type_from_flags (flags, FALSE, error);
     success = bd_lvm_lvcreate (vg_name, pool_name, pool_size, type, fast_pvs, NULL, error);
     if (!success) {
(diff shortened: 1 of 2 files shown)


vpodzime:
This is how the example run looks like:

```
In [4]: BlockDev.lvm.cache_create_cached_lv("test", "cachedLV", 512*1024**2, 500*1024**2, BlockDev.LVMCacheMode.WRITETHROUGH, 0, 0, ["/dev/sda"], ["/dev/sdb"])
Started 'create cached LV test/cachedLV' [1]
Started 'create cache pool test/cachedLV_cache' [2]
[2] 33 % completed
[2] 66 % completed
[2] Completed
[1] 60 % completed
[1] 80 % completed
[1] Completed
Out[4]: True
```

src/plugins/lvm-dbus.c
```diff
@@ -2517,28 +2539,44 @@ gboolean bd_lvm_cache_create_cached_lv (const gchar *vg_name, const gchar *lv_na
...
     success = bd_lvm_cache_create_pool (vg_name, name, cache_size, md_size, mode, flags, fast_pvs, error);
     if (!success) {
         g_prefix_error (error, "Failed to create the cache pool '%s': ", name);
         g_free (name);
+        bd_utils_report_finished (progress_id, (*error)->message);
         return FALSE;
     }
 
+    /* 4/5 steps (cache pool creation has 3 steps) done */
+    bd_utils_report_progress (progress_id, 20, "Cache pool created");
```
dwlehman: Probably supposed to be 80 instead of 20.

vpodzime: Good catch, thanks!

PR #103, opened 2016-08-12
Support for Vfat

Blasterdude:
This definitely needs a little work, but use this as a basis I guess.


vpodzime:
#117 is a replacement for this PR

PR #105, opened 2016-08-17
Correct underlinking of shared libraries

tasleson:
When playing around with the library I discovered that I needed to link in libraries which were needed by libblockdev libraries themselves and not my application.  This __underlinking__ is more than just an inconvenience, it creates direct dependencies to libraries that the end application doesn't directly use.  Thus when a new dependent library is released, the end binaries need to be recompiled too, when only the underlying library needed to be re-compiled.

```
gcc -Wall main.c `pkg-config blockdev --cflags --libs`
/usr/lib/gcc/x86_64-redhat-linux/6.1.1/../../../../lib64/libblockdev.so: undefined reference to `dlopen'
/usr/lib/gcc/x86_64-redhat-linux/6.1.1/../../../../lib64/libblockdev.so: undefined reference to `dlclose'
/usr/lib/gcc/x86_64-redhat-linux/6.1.1/../../../../lib64/libblockdev.so: undefined reference to `dlerror'
/usr/lib/gcc/x86_64-redhat-linux/6.1.1/../../../../lib64/libblockdev.so: undefined reference to `dlsym'
/usr/lib64/libbd_utils.so.0: undefined reference to `pow'
/usr/lib/gcc/x86_64-redhat-linux/6.1.1/../../../../lib64/libblockdev.so: undefined reference to `g_boxed_type_register_static'
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -5,7 +5,7 @@ SUBDIRS = plugin_apis
 lib_LTLIBRARIES = libblockdev.la
 libblockdev_la_CFLAGS = $(GLIB_CFLAGS)
 libblockdev_la_LIBADD = $(GLIB_LIBS) ${builddir}/../utils/libbd_utils.la
-libblockdev_la_LDFLAGS = -L${srcdir}/../utils/ -version-info 4:1:4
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me. Resolve the conflicts, please, so that I can merge this. Thanks!

tasleson:
@vpodzime Conflicts resolved.

PR #106, opened 2016-08-26
Add progress reporting to DM RAID (de)activation

vpodzime:
We don't exec any utilities for these actions so they are not covered by the
generic progress reporting code.


vojtechtrefny:
Looks good to me.

PR #107, opened 2016-08-29
Add progress reporting for the part plugin functions

vpodzime:
There's nothing we can do regarding any real progress reporting because all the
changes happen at once when the model is committed to disk, but we need to
report start and end of each action.


vojtechtrefny:
Looks good to me.

PR #109, opened 2016-08-31
Add progress reporting to the crypto plugin

vpodzime:
The bd_crypto_luks_format() function does multiple steps allowing us to do some
granular progress reporting. Others are just single calls to libcryptsetup that
don't provide any progress for us to propagate.


vojtechtrefny:
Looks good to me.

PR #110, opened 2016-09-01
Add progress reporting to the KBD plugin

vpodzime:
Not much we can do here apart from reporting actions starting a finishing.


vojtechtrefny:
Looks good to me.

vpodzime:
Tests were canceled because the changes were just cosmetic.

PR #111, opened 2016-09-01
Add progress reporting to the get_mpath_members() function

vpodzime:
The only function from the mpath plugin for which progress reporting makes sense
(even if just start and end) and which is not covered by the generic code in the
exec* functions.


vojtechtrefny:
Looks good to me.

vpodzime:
Tests were canceled because the changes were just cosmetic.

PR #114, opened 2016-09-13
Add functions for adding/removing single zRAM devices

vpodzime:
The kernel API now supports such operations so we should support them too.


vpodzime:
Jenkins, test this, please.

PR #116, opened 2016-09-16
Make sure all object paths are passed and extracted as such

vpodzime:
The type signature for a DBus object path is "o". Now that the LVM DBus API
properly sets types of its accepted and returned values, we can finally safely
use the signature in all expected places.


vpodzime:
The tests failed due to some issues in the ``lvm-dbusd`` code.

PR #118, opened 2016-09-19
Add close_plugin() functions to plugins

vpodzime:
So that they can clean after themselves when being unloaded.


vpodzime:
> Are you sure the collision isn't with close(2) from unistd.h?

Could be, but I'm not sure if ``dlsym()`` would "grab" this function from plugins. Anyway, we should avoid using such a name.

vpodzime:
See also https://github.com/rhinstaller/libblockdev/issues/119

PR #125, opened 2016-10-04
Fixes for FW RAID 'mdadm --examine' processing

vpodzime:
mdadm provides information about FW RAID in weird formats we have to deal with.

Related: rhbz#1379865


src/plugins/mdraid.c
```diff
@@ -212,6 +214,9 @@ static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean
         data->size = 0;
 
     data->uuid = g_strdup ((gchar*) g_hash_table_lookup (table, "Array UUID"));
+    if (!data->uuid)
+        /* also try just "UUID" which may be reported e.g for IMSM FW RAID */
+        data->uuid = g_strdup ((gchar*) g_hash_table_lookup (table, "UUID"));
```
AdamWill: which UUID is this going to get us? always the second one? wouldn't it be safer to get this from one of the other output formats, for the fwraid case?

vpodzime: This should be the first UUID (i.e. the container UUID). Looks like all the MD RAID information gathering needs to get much better, probably using attributes in /sys and+or udev data.

PR #129, opened 2016-10-10
Add extra parameters to loop_setup required by storaged

vpodzime:
storaged supports almost everything that's possible to do as part of setting up
a loop device. In order to be a sufficient backend for storaged we need to
support the same set of features.

We need to reorganize the tests a little bit to make sure that everything is
cleaned up properly in case of failure.

### DIFF
--- a/src/lib/plugin_apis/loop.api
+++ b/src/lib/plugin_apis/loop.api
@@ -29,12 +29,16 @@ gchar* bd_loop_get_loop_name (const gchar *file, GError **error);
 /**
  * bd_loop_setup:
  * @file: file to setup as a loop device
+ * @offset: offset of the start of the device (in @file)
+ * @size: maximum size of the device (or 0 to leave unspecified)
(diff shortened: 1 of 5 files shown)


src/lib/plugin_apis/loop.api
```diff
@@ -29,12 +29,16 @@ gchar* bd_loop_get_loop_name (const gchar *file, GError **error);
 /**
  * bd_loop_setup:
  * @file: file to setup as a loop device
+ * @offset: offset of the start of the device (in @file)
+ * @size: maximum size of the device (or 0 to leave unspecified)
+ * @read_only: whether to setup se read-only (%TRUE) or read-write (%FALSE)
```
dwlehman: "whether to setup _as_ read-only"?

vpodzime: Sure thing, thanks! Fixing.

vpodzime:
Jenkins, test this, please.

PR #131, opened 2016-10-17
Pre-release housekeeping

vpodzime:
A few little tweaks we need to do before we reach the 2.0 release milestone.

### DIFF
--- a/README
+++ b/README
@@ -1,4 +1,15 @@
-A library for manipulating block devices. For more information about the
-library's architecture see the specs.rst file. For more infomation about the
-expected functionality and features see the features.rst file. For more
-information about the schedule see the roadmap.rst file.
+libblockdev is a C library supporting GObject introspection for manipulation of
+block devices. It has a plugin-based architecture where each technology (like
+LVM, Btrfs, MD RAID, Swap,...) is implemented in a separate plugin, possibly
+with multiple implementations (e.g. using LVM CLI or the new LVM DBus API).
+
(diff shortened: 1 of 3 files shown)


vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Unfortunately, this requires a different approach. Looks like we must mark just the structs that we export as packed. Otherwise at least parted fails with weird errors.

vpodzime:
> Unfortunately, this requires a different approach. Looks like we must mark just the structs that we export as packed. Otherwise at least parted fails with weird errors.

If only this worked... The reality is that GObject introspection doesn't work with packed structures. So we cannot do this (at least for now).

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Merged changes from #132 here as those are needed for tests to pass.

vpodzime:
> I don't know what would happen if you merged that PR and this one with the other already in it.

I don't either, but I'd like to find it out for future generations (PRs).

vpodzime:
It went pretty well, I'd say. I can see two merge commits for the branch with parted fixes, but e.g. ``tig`` shows it very nicely. Plain ``git log`` just shows the merge commits, ``git log --no-merges`` is as clean as snow. :)

PR #132, opened 2016-10-20
Prevent issues between libparted and udev

vpodzime:
libparted does committing of a new partition table to disk in two steps:

1. committing (writing and flushing) to disk and
2. telling OS about the partitions.

However, that means that if things go wrong, udev may step in between the above
steps and trigger events when the block device file is closed (from being opened
RW) that could lead to a failure of the second step because some other process
may open the block device file in the reaction to the event.


vpodzime:
Jenkins, test this, please.

PR #133, opened 2016-10-24
Handle mdadm --examine output during migration

AdamWill:
If a RAID set is undergoing any operation considered to be a
'migration' - which includes initialization of a new set at any
RAID level that includes redundancy - mdadm --examine will have
lines like this:

     RAID Level : 5 <-- 5
        Members : 3 <-- 3
      Map State : normal <-- uninitialized

At present we don't understand this at all. This is a fairly
minimal fix which just has parse_mdadm_vars notice such lines,
split the 'value' on "<--", and stuff the first value (the one
being migrated *to*, which is usually what we care about, I
think) into the table. We could do something more elaborate
like store both values and make `get_examine_data_from_table`
understand that, but it's more complicated and this should be
all we truly need for now.

### DIFF
--- a/src/plugins/mdraid.c
+++ b/src/plugins/mdraid.c
@@ -183,6 +183,7 @@ static GHashTable* parse_mdadm_vars (const gchar *str, const gchar *item_sep, co
     gchar **items = NULL;
     gchar **item_p = NULL;
     gchar **key_val = NULL;
+    gchar **vals = NULL;
 
     table = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
     *num_items = 0;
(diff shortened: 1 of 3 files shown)


AdamWill:
This is intended to fix https://bugzilla.redhat.com/show_bug.cgi?id=1381996 ; my firmware RAID test box is on the fritz ATM so I'll ask pschindl to test it.

AdamWill:
I took the fake mdadm output for the test from the logs in the bug, BTW.

src/plugins/mdraid.c
```diff
@@ -193,8 +194,16 @@ static GHashTable* parse_mdadm_vars (const gchar *str, const gchar *item_sep, co
         if (g_strv_length ((gchar **) key_val) == 2) {
             /* we only want to process valid lines (with the separator) */
             /* only use the first value for the given key */
-            if (!g_hash_table_contains (table, g_strstrip (key_val[0])))
-                g_hash_table_insert (table, g_strstrip (key_val[0]), g_strstrip (key_val[1]));
+            if (!g_hash_table_contains (table, g_strstrip (key_val[0]))) {
+                vals = g_strsplit (key_val[1], "<--", 2);
+                if (g_strv_length ((gchar **) vals) == 2)
```
vpodzime: I think the ``if`` should first check if the ``"<--"`` string is in ``key_val[1]``, decide on that and then do the ``g_strsplit()`` (as well as ``g_free()`` the second value) only in the particular branch.

AdamWill: ok! I literally have no idea what I'm doing in C so I took the safe path by mimicking the existing code as far as I could, but I'll figure that out tomorrow =)

AdamWill:
how's that?

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

AdamWill:
thanks! Could you please backport for F25 and do a package build, too?

vpodzime:
> thanks! Could you please backport for F25 and do a package build, too?

Done - https://bodhi.fedoraproject.org/updates/FEDORA-2016-4720fedecc

PR #146, opened 2016-11-28
Extra functions for the loop plugin required by storaged

vpodzime:
There are basically two things required by storaged which are not currently provided by the loop plugin:

- [x] functions to get/set the *autoclear* attribute
- [x] function to setup a loop device from/on a file descriptor


vpodzime:
All done now.

PR #154, opened 2017-01-16
Try harder to register a new bcache device

vpodzime:
Sometimes the /sys/.../register file doesn't exist when we try to write to
it. Let's give it a chance to appear in some sensible time frame. With some luck
the file appears and the bcache device is successfully registered (and then
properly created and set up).

Also enable some of the risky tests that seem to have been failing due to the
issue fixed here.


vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

Issue #149, opened 2017-01-13
Out of memory during the input sanitization in s390.c plugin


poncovka:
Functions `bd_s390_sanitize_dev_input` and  `bd_s390_zfcp_sanitize_lun_input` can cause an out of memory error for invalid input.

For example in the command `prepend = g_strnfill((4 - strlen(tmptok)), '0');`, the length of the string `tmpok` is not checked, therefore `4 - strlen(tmptok)` could be a negative number. However, `g_strnfill` expects an unsigned long integer, so the number will be converted to a very large number and `g_strnfill` will try to create a string of this length.

Problematic inputs:
`bd_s390_sanitize_dev_input("123456")`
`bd_s390_zfcp_sanitize_wwpn_input("12345678901234567890")`

vpodzime:
Thanks for the report!

Issue #145, opened 2016-11-23
The wipe() function needs to try harder


vpodzime:
Right now it errors out if it fails to get a *libblkid* ``probe`` for the block device. But the ``wipefs`` utility works in those cases. So we probably need to try multiple times, with some delays.?

Issue #104, opened 2016-08-15
Build dependencies not specified in configure.ac


tasleson:
``` bash
$ ./autogen.sh && ./configure
```

 "GObject introspection must be enabled" which is misleading as what you may need to do is install package for `gobject-introspection-devel`

Once I got `./configure` to run without complaints, I get the following compile errors:

```
crypto.c:24:27: fatal error: libvolume_key.h: No such file or directory
 #include <libvolume_key.h>
fatal error: dmraid/dmraid.h: No such file or directory
 #include <dmraid/dmraid.h>

gtkdoc-scan: Command not found
```

Which can be corrected by installing development libraries for `volume_key-devel`, `dmraid-devel` and `gtk-doc` packages.

PR #158, opened 2017-02-01
A few minor fixes

vpodzime:
These fixes address the following minor issues reported here at GH:

- [x] Build dependencies not specified in configure.ac (#104)
- [x] The wipe() function needs to try harder (#145)
- [x] Out of memory during the input sanitization in s390.c plugin (#149)


vpodzime:
Jenkins, test this, please.

PR #159, opened 2017-02-02
Allow using both path and raid name in bd_md_set_bitmap_location

vojtechtrefny:
'mdadm --grow' accepts either '/dev/mdXXX' or '/dev/md/name' so
allow using both name (or just node name) and full path from
libblockdev.

### DIFF
--- a/src/plugins/mdraid.c
+++ b/src/plugins/mdraid.c
@@ -1162,28 +1162,51 @@ gchar* bd_md_get_status (const gchar *raid_name, GError **error) {
 
 /**
  * bd_md_set_bitmap_location:
- * @raid_name: name of the RAID device to set the bitmap location
+ * @raid_name: name or path of the RAID device to set the bitmap location
  * @location: bitmap location (none, internal or path)
  * @error: (out): place to store error (if any)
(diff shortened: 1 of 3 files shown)


src/plugins/mdraid.c
```diff
@@ -1162,28 +1162,53 @@ gchar* bd_md_get_status (const gchar *raid_name, GError **error) {
...
+    else
+      {
+        name_path_str = g_strdup_printf ("/dev/md/%s", raid_name);
+        md_path_str = g_strdup_printf ("/dev/%s", raid_name);
+        if (access (name_path_str, F_OK) == 0)
+            argv[2] = name_path_str;
+        else if (access (md_path_str, F_OK) == 0)
+            argv[2] = md_path_str;
+        else
+          {
```
vpodzime: Please put the braces after ``else`` to the same line as the ``else`` itself. (you are not writing storaged code here ;-))

vojtechtrefny: Fixed.

vpodzime:
Jenkins, test this, please.

PR #163, opened 2017-02-14
Add support for wipefs

jaymzh:
mkswap leaves the FS signatures at the end of the partition around which
confuses a ton of tools. We should wipe that out first. Adding it here so blivet
can call it.

NOTE: This is not yet tested... mostly sending it for general feedback on if
this is reasonable.

### DIFF
--- a/src/plugins/swap.c
+++ b/src/plugins/swap.c
@@ -85,6 +85,27 @@ void bd_swap_close () {
 }
 
 /**
+ * bd_swap_wipefs:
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
We already have wipefs function (using libblkid) in the FS plugin -- `bd_fs_wipe` -- you should use that one from blivet.

jaymzh:
Oh sweet, will do. Thanks! Sorry i missed that.

PR #165, opened 2017-02-17
Add functions using BLOBs as LUKS passphrases

vpodzime:
A LUKS passphrase can be an arbitrary sequence of bytes including '\0' (0) bytes
which are terminating strings in C. If we take the passphrase argument as
'gchar *', we need it to be '\0'-terminated otherwise we don't know its
length. We need a set of new functions taking the passphrase argument as
'guint8 *' together with its length (gsize) if we want to support arbitrary
sequences of bytes as LUKS passphrases.

Note: this is required by https://github.com/storaged-project/udisks/pull/195


vpodzime:
Jenkins, test this, please.

PR #170, opened 2017-03-03
Add the bd_s390_dasd_is_fba function to check if DASD is FBA

poncovka:
This will enable blivet to choose a correct disklabel type for an FBA device.

### DIFF
--- a/src/plugins/s390.c
+++ b/src/plugins/s390.c
@@ -252,8 +252,16 @@ gboolean bd_s390_dasd_is_ldl (const gchar *dasd, GError **error) {
     dasd_information2_t dasd_info;
 
     memset(&dasd_info, 0, sizeof(dasd_info));
-    devname = g_strdup_printf ("/dev/%s", dasd);
 
+    /* complete the device name */
+    if (g_str_has_prefix (dasd, "/dev/")) {
+        devname = g_strdup (dasd);
+    }
+    else {
+        devname = g_strdup_printf ("/dev/%s", dasd);
+    }
+
+    /* open the device */
     if ((f = open(devname, O_RDONLY)) == -1) {
         g_set_error (error, BD_S390_ERROR, BD_S390_ERROR_DEVICE,
                      "Unable to open device %s", devname);
@@ -263,11 +271,13 @@ gboolean bd_s390_dasd_is_ldl (const gchar *dasd, GError **error) {
 
     g_free (devname);
 
+    /* check if this is a block device */
     if (ioctl(f, BLKSSZGET, &blksize) != 0) {
         close(f);
         return FALSE;
     }
 
+    /* get some info about DASD */
     if (ioctl(f, BIODASDINFO2, &dasd_info) != 0) {
         close(f);
         return FALSE;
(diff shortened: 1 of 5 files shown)


src/plugins/s390.c
```diff
@@ -290,6 +290,52 @@ gboolean bd_s390_dasd_is_ldl (const gchar *dasd, GError **error) {
...
+ * Returns: whether a dasd is FBA
+ */
+gboolean bd_s390_dasd_is_fba (const gchar *dasd, GError **error) {
+    gchar *devname = NULL;
+    gint f = 0;
+    gint blksize = 0;
+    dasd_information2_t dasd_info;
+
+    memset(&dasd_info, 0, sizeof(dasd_info));
+    devname = g_strdup_printf ("/dev/%s", dasd);
```
vpodzime: Please check that ``dasd`` doesn't already start with ``"/dev/"`` here.

src/plugins/s390.c
```diff
@@ -290,6 +290,52 @@ gboolean bd_s390_dasd_is_ldl (const gchar *dasd, GError **error) {
...
+        g_free (devname);
+        return FALSE;
+    }
+
+    g_free (devname);
+
+    if (ioctl(f, BLKSSZGET, &blksize) != 0) {
+        close(f);
+        return FALSE;
+    }
```
vpodzime: What is the ``blksize`` good for? I cannot see it being used anywhere below.

poncovka: The function `bd_s390_dasd_is_fba` is copy-pasted from  `bd_s390_dasd_is_ldl`. The corresponding code for RHEL 7 is here: https://github.com/rhinstaller/blivet/blob/rhel7-branch/blivet/devicelibs/dasd.py#L126. There is a comment "poorly check if this is even a block device...".

src/plugins/s390.c
poncovka: Should I also change `bd_s390_dasd_is_ldl`? (see the comment below)

vpodzime: Yes please.

src/plugins/s390.c
vpodzime: Oh, then this should have the same comment justifying the reason for doing the ``ioctl()``. (and the same applies to ``is_ldl()``). Thanks!

src/plugins/s390.c
poncovka: Fixed.

src/plugins/s390.c
poncovka: Fixed.

src/plugins/s390.c
```diff
@@ -290,6 +290,62 @@ gboolean bd_s390_dasd_is_ldl (const gchar *dasd, GError **error) {
...
+
+    close(f);
+
+    /* check if we're on an FBA DASD */
+    if (strncmp(dasd_info.type, "FBA", 3) == 0) {
+        return TRUE;
+    }
+    else {
+        return FALSE;
```
vpodzime: Please replace this ``if`` with ``return strncmp(dasd_info.type, "FBA", 3) == 0``.

poncovka: Fixed.

Issue #178, opened 2017-03-30
Linking fails, missing gobject-2.0 and libm?


andhe:
Trying to build libblockdev 2.6 (on Debian) results in the following error:

```
/bin/bash ../../libtool  --tag=CC   --mode=link gcc -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -g -O2 -fdebug-prefix-map=/build/libblockdev-2.6=. -fstack-protector-strong -Wformat -Werror=format-security -L./../utils/ -version-info 2:0:0 -Wl,--no-undefined -Wl,-z,relro -o libblockdev.la -rpath /usr/lib/x86_64-linux-gnu libblockdev_la-blockdev.lo libblockdev_la-plugins.lo -lglib-2.0 ./../utils/libbd_utils.la -ldl 
libtool: link: gcc -shared  -fPIC -DPIC  .libs/libblockdev_la-blockdev.o .libs/libblockdev_la-plugins.o   -Wl,-rpath -Wl,/build/libblockdev-2.6/src/utils/.libs -L./../utils/ -lglib-2.0 ./../utils/.libs/libbd_utils.so -ldl  -g -O2 -fstack-protector-strong -Wl,--no-undefined -Wl,-z -Wl,relro   -Wl,-soname -Wl,libblockdev.so.2 -o .libs/libblockdev.so.2.0.0
.libs/libblockdev_la-blockdev.o: In function `bd_lvm_pvdata_get_type':
./src/lib/plugin_apis/lvm.c:61: undefined reference to `g_boxed_type_register_static'
.libs/libblockdev_la-blockdev.o: In function `bd_lvm_vgdata_get_type':
./src/lib/plugin_apis/lvm.c:115: undefined reference to `g_boxed_type_register_static'
.libs/libblockdev_la-blockdev.o: In function `bd_lvm_lvdata_get_type':
./src/lib/plugin_apis/lvm.c:186: undefined reference to `g_boxed_type_register_static'
.libs/libblockdev_la-blockdev.o: In function `bd_lvm_cache_stats_get_type':
./src/lib/plugin_apis/lvm.c:244: undefined reference to `g_boxed_type_register_static'
.libs/libblockdev_la-blockdev.o: In function `bd_btrfs_device_info_get_type':
./src/lib/plugin_apis/btrfs.c:42: undefined reference to `g_boxed_type_register_static'
[...]

PR #180, opened 2017-03-31
libparted-fs-resize.pc not available in 3.2

andhe:
Bump the version of libparted that's expected to have a .pc file for
libparted-fs-resize since that was not yet shipped in 3.2 it seems.
(Atleast not available in debian packages of libparted 3.2)

See #178

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -177,7 +177,7 @@ AS_IF([test "x$with_fs" != "xno"],
               [Define as neutral value if libblkid doesn't provide the definition])])
 
        # older versions of parted don't provide the libparted-fs-resize.pc file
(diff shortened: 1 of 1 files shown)


andhe:
The libparted-fs-resize.pc seems to exist in parted master branch so I assume it will be part of next (3.3?) release......

vpodzime:
> The libparted-fs-resize.pc seems to exist in parted master branch so I assume it will be part of next (3.3?) release...... Atleast as far as I remember when I looked at this.

Probably, but I think the best way for this is to use ``AS_IF([pkg-config libparted-fs-resize]`` and switch the branches.

vpodzime:
This is resolved by d6e00cb8d9c23e184dc4d6adf1eb97eb593242bb

PR #182, opened 2017-04-03
Make sure the returned thpool MD size is valid

vpodzime:
It may be too small as given by the thin_metadata_size tool which
obviously doesn't guard for too small values.


vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

PR #183, opened 2017-04-12
Do not try to parse 'raid_spec' for 'bd_md_activate' (#1439111)

vojtechtrefny:
It is possible to call 'mdadm --assemble' using the '/dev/md...'
path so do not try to parse/check given raid specification --
even non-existing path is a valid input for this function.


vojtechtrefny:
Jenkins, test this please.

PR #185, opened 2017-04-20
Fixes to make tests passing on CentOS/RHEL 7

vpodzime:
These are the fixes and adaptations of both the code and tests to allow us running tests on CentOS/RHEL 7. The only remaining failing test is the one for getting *zram* stats where the failure is caused by incompatibilities between the 3.10 and almost-vanilla kernels in el7 and Fedora respectively.

Apart from disabling tests related to bcache (which is not available on el7) there are bugs in the el7 versions of ``volume_key`` and ``libparted`` that make it impossible to run some tests properly. Those will be reported.


vpodzime:
@vojtechtrefny needs this merged.

PR #189, opened 2017-04-25
Skip btrfs min size tests on Fedora 25

vojtechtrefny:
Btrfs has different min size limits on Fedora 25.


vpodzime:
One has to love *Btrfs*

PR #190, opened 2017-04-28
Add functions for resolving device (symlinks) and getting device symlinks

vpodzime:
These often come handy. Is the fact that the *utils* library now requires *libudev* a problem? I think it should be fine since that library is required by a number of core system components and the *libblockdv-dm* plugin already requires it too.

### DIFF
--- a/src/utils/dev_utils.c
+++ b/src/utils/dev_utils.c
@@ -18,8 +18,17 @@
  */
 
 #include <glib.h>
+#include <libudev.h>
 
 #include "dev_utils.h"
+
+/**
+ * bd_utils_dev_utils_error_quark: (skip)
+ */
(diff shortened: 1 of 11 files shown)


src/utils/dev_utils.c
```diff
@@ -23,7 +23,8 @@
 
 /**
  * bd_utils_resolve_device:
- * @dev_spec: specification of the device (e.g. "/dev/sda", any symlink, or the name of a file under "/dev")
+ * @dev_spec: specification of the device (e.g. "/dev/sda", any symlink, or the name of a file
+ *            under "/dev")
```
vojtechtrefny: This should probably be part of the previous commit.

PR #191, opened 2017-05-02
First try to read the 'autoclear' flag from /sys/

vpodzime:
The 'autoclear' flag of loop devices can either be read from /sys
or by doing an ioctl() on the device. While the latter one is
simpler and cheaper it may not work 100 % of time due to
interactions with udev. Acquiring an flock() for the time doing
the ioctl() often causes udev to not update its database (in
time). So reading the flag from /sys is safer and thus should be
preferred.


vpodzime:
We need this on our testing machines ASAP.

PR #195, opened 2017-05-29
Add a function for cleaning a device

vojtechtrefny:
Added bd_fs_clean() function that removes all signatures from given
device.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -207,6 +207,20 @@ GType bd_fs_vfat_info_get_type () {
 gboolean bd_fs_wipe (const gchar *device, gboolean all, GError **error);
 
 /**
+ * bd_fs_clean:
+ * @device: the device to clean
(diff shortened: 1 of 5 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -207,6 +207,19 @@ GType bd_fs_vfat_info_get_type () {
 gboolean bd_fs_wipe (const gchar *device, gboolean all, GError **error);
 
 /**
+ * bd_fs_clean:
+ * @device: the device to clean
+ * @error: (out): place to store error (if any)
+ *
+ * Difference between this and bd_fs_wipe() is that this function doesn't
```
vpodzime: This should start with something like ``Clean all signatures from the device.`` so that people don't have to go and read what ``bd_fs_wipe()`` does first.

vojtechtrefny: Fixed.

PR #199, opened 2017-06-02
Test fixes

vpodzime:
This is an attempt to stabilize *libblockdev* tests once again because otherwise they are just failing horribly every time they are run.


vpodzime:
Merging this ASAP so that running our tests once again makes sense.

PR #209, opened 2017-06-09
Do not link libraries required by plugins to the GI files

vpodzime:
Only the plugins require those not the library itself.


vpodzime:
Jenkins, test this, please.

PR #212, opened 2017-06-12
Fix hardcoded reference to gcc

tgurr:
Fixes build failures in cross environments where the tools may be
prefixed with the host triplet, e.g. x86_64-pc-linux-gnu-gcc.


vpodzime:
Jenkins, this is ok to test.

PR #215, opened 2017-06-13
kbd.c: Make bd_kbd_bcache_create work without abort

tasleson:
When you create a cached device with a backing device at the same
time you do not need to separately register and attach devices.
This is done by the udev rules.  This fixes an issue where the
code was doing an attach for a device that was already attached.
The bcache sysfs attach kernel entry will return -2 in an unsigned,
which yields 2**32-2 returned to user space.  This will cause an
abort if you are running G_DEBUG=fatal-criticals in glib as
g_io_channel_flush does a g_string_erase of the write buffer for
the number of bytes written, which is ~4G, and we only actually
have a 1K buffer that we wrote 36 bytes of.

Signed-off-by: Tony Asleson <tasleson@redhat.com>

### DIFF
--- a/src/plugins/kbd.c
+++ b/src/plugins/kbd.c
@@ -749,6 +749,19 @@ BDKBDZramStats* bd_kbd_zram_get_stats (const gchar *device, GError **error) {
 
 
 #ifdef WITH_BD_BCACHE
+
+gboolean wait_for_file (const char *filename) {
+    gint count = 500;
+    while (count > 0) {
+        g_usleep (100000); /* microseconds */
+        if (g_file_test (filename, G_FILE_TEST_EXISTS)) {
+            break;
+        }
+        --count;
+    }
+    return (count > 0) ? TRUE : FALSE;
+}
+
 /**
  * bd_kbd_bcache_create:
  * @backing_device: backing (slow) device of the cache
@@ -767,17 +780,16 @@ gboolean bd_kbd_bcache_create (const gchar *backing_device, const gchar *cache_d
     gchar **lines = NULL;
     GRegex *regex = NULL;
     GMatchInfo *match_info = NULL;
-    gchar *set_uuid = NULL;
     guint i = 0;
-    gboolean found = FALSE;
     glob_t globbuf;
     gchar *pattern = NULL;
     gchar *path = NULL;
     gchar *dev_name = NULL;
     gchar *dev_name_end = NULL;
-    guint n_retry = 5;
     guint64 progress_id = 0;
     gchar *msg = NULL;
+    guint n = 0;
+    gchar device_uuid[2][64];
 
     msg = g_strdup_printf ("Started creation of bcache on '%s' and '%s'", backing_device, cache_device);
     progress_id = bd_utils_report_started (msg);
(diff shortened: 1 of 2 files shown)


vpodzime:
Jenkins, this is ok to test.

src/plugins/kbd.c
```diff
@@ -749,6 +749,19 @@ BDKBDZramStats* bd_kbd_zram_get_stats (const gchar *device, GError **error) {
...
+gboolean wait_for_file (const char *filename) {
+    gint count = 500;
+    while (count > 0) {
+        g_usleep (100000); /* microseconds */
+        if (g_file_test (filename, G_FILE_TEST_EXISTS)) {
+            break;
+        }
+        --count;
+    }
+    return (count > 0) ? TRUE : FALSE;
```
vpodzime: I think it would be nicer to do ``return TRUE`` instead of the ``break`` above and just ``return FALSE`` here.

src/plugins/kbd.c
```diff
@@ -805,56 +817,46 @@ gboolean bd_kbd_bcache_create (const gchar *backing_device, const gchar *cache_d
         return FALSE;
     }
 
-    for (i=0; !found && lines[i]; i++) {
+    for (i=0; lines[i]; i++) {
         success = g_regex_match (regex, lines[i], 0, &match_info);
         if (success) {
-            found = TRUE;
-            set_uuid = g_match_info_fetch (match_info, 1);
+            strcpy(device_uuid[n++], g_match_info_fetch (match_info, 1));
```
vpodzime: Please add a space between the function's name.

src/plugins/kbd.c
```diff
@@ -805,56 +817,46 @@ gboolean bd_kbd_bcache_create (const gchar *backing_device, const gchar *cache_d
...
+            g_match_info_free (match_info);
+            if (n == 2) {
+                break;
```
vpodzime: I'd prefer ``n < 2`` being used as part of the ``for`` cycle's condition.

src/plugins/kbd.c
```diff
@@ -805,56 +817,46 @@ gboolean bd_kbd_bcache_create (const gchar *backing_device, const gchar *cache_d
...
         g_free (output);
         bd_utils_report_finished (progress_id, (*error)->message);
         return FALSE;
     }
     g_free (output);
 
+    /* Wait for the symlinks to show up, would it be better to do a udev settle? */
+    for (i=0; i < 2; i++) {
+        const char *uuid_file = g_strdup_printf ("/dev/disk/by-uuid/%s", device_uuid[i]);
+        gboolean present = wait_for_file(uuid_file);
```
vpodzime: Again a space after the function name.

vpodzime:
Just one more question here -- do you know how the *udev* rules know which cache to attach to which backing device?

vpodzime:
I tried running the *bcache* tests with these changes and they seem to work great. Could you please try to remove the ``not feeling lucky`` guards on them so that they are run by the CI on this PR?

tasleson:
Thanks for the review, all suggestions done!

tasleson:
> Just one more question here -- do you know how the udev rules know which cache to attach to which 
> backing device?

I can only conclude that the necessary information is stored in the superblocks of the devices which belong to the set.

vpodzime:
> I can only conclude that the necessary information is stored in the superblocks of the devices which belong to the set.

Oh, I now see that we are creating the backing and cache devices' signatures in one step. We/I used to do that in two steps which was probably also the reason why the following attach was necessary and working. Thanks a lot for fixing this!

PR #216, opened 2017-06-13
Ignore parted warnings if possible

### LINKED ISSUE #214: Parted warning is treated as an error
Currently any parted exception is treated as an error (which leads to abort). A parted exception may be only a warning and it should be treated as such in that case.

I would love to work on a PR to fix this.

I'm not sure how warnings should be treated in this library and therefore suggestions are welcome.

Furthermore this issue is a question on whether PRs on this are welcome at all.

squimrel:
Ignore parted exceptions of type warning and information if parted marks them as ignorable. Fixes #214.

~~I haven't tested this yet \*cough*. Plus I should probably add at least one test case in which a parted warning is ignored. So please don't merge yet.~~

### DIFF
--- a/src/plugins/part_err.c
+++ b/src/plugins/part_err.c
@@ -17,6 +17,7 @@
  * Author: Vojtech Trefny <vtrefny@redhat.com>
  */
 
+#include <glib.h>
 #include <parted/parted.h>
 
 #include "part_err.h"
@@ -24,6 +25,10 @@
 static __thread gchar *error_msg = NULL;
 
 PedExceptionOption bd_exc_handler (PedException *ex) {
+    if (ex->type <= PED_EXCEPTION_WARNING && (ex->options & PED_EXCEPTION_IGNORE) != 0) {
+      g_warning ("[parted] %s", ex->message);
+      return PED_EXCEPTION_IGNORE;
+    }
     error_msg = g_strdup (ex->message);
     return PED_EXCEPTION_UNHANDLED;
 }


src/plugins/part_err.c
```diff
@@ -24,8 +24,16 @@
 static __thread gchar *error_msg = NULL;
 
 PedExceptionOption bd_exc_handler (PedException *ex) {
-    error_msg = g_strdup (ex->message);
-    return PED_EXCEPTION_UNHANDLED;
+    /*
+     * Only treat exceptions as errors if they aren't at least errors or can't
+     * be ignored.
+     */
```
squimrel: This comment seems pretty superfluous to me.

squimrel:
This seems to work but I have a hard time creating a test that throws a parted warning so I'd think about giving up on it. Current tests still pass.

src/plugins/part_err.c
vpodzime: ``aren't`` should be ``are``, shouldn't it?

src/plugins/part_err.c
```diff
@@ -17,15 +17,24 @@
...
+    if (ex->type >= PED_EXCEPTION_ERROR || (ex->options & PED_EXCEPTION_IGNORE) == 0) {
+      error_msg = g_strdup (ex->message);
+      return PED_EXCEPTION_UNHANDLED;
+    }
+    g_warning ("[parted] %s", ex->message);
+    return PED_EXCEPTION_IGNORE;
```
vpodzime: I'd prefer if the logic was inverted here. So something like ``if THIS_CAN_BE_IGNORED -> ignore; else raise``

squimrel:
There you go.

vpodzime:
Jenkins, this is ok to test.

PR #217, opened 2017-06-14
Query functions for FS resize and repair support

pothos:
The FS plugin functions bd_fs_can_resize/check/repair
can be used to test for installed utility and FS support
before starting the action or exposing it in an UI.

Maybe you can give me feedback on this - I intend to use it in UDisks. It already has a value e.g. because it could be called to see if resize is supported by the generic resize method.

But to make it more meaningful the check_deps() would need to not already fail if e.g. xfs_db is missing. Currently this requires UDisks to depend on xfsprogs and if other filesystems like ntfs would be supported in FS plugin they all would need to be dependencies if check_deps() will remain strict. Through these functions the caller could query for support and there is no need for a hard dependency. But this would be another PR.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -362,6 +362,63 @@ gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
 gboolean bd_fs_resize (const gchar *device, guint64 new_size, GError **error);
 
 /**
+ * BDFsResizeFlags:
+ * Flags indicating whether a filesystem resize action supports growing and/or
+ * shrinking if mounted or unmounted.
+ */
+typedef enum {
+    BD_FS_OFFLINE_SHRINK = 1 << 1,
+    BD_FS_OFFLINE_GROW = 1 << 2,
+    BD_FS_ONLINE_SHRINK = 1 << 3,
+    BD_FS_ONLINE_GROW = 1 << 4
+} BDFsResizeFlags;
+
+/**
+ * bd_fs_can_resize:
+ * @type: the filesystem type to be tested for installed resize support
+ * @mode: (out): flags for allowed resizing (i.e. growing/shrinking support for online/offline)
+ * @required_utility: (out) (transfer full): the utility binary which is required for resizing (if missing i.e. returns FALSE but no error)
+ * @error: (out): place to store error (if any)
+ *
+ * Searches for the required utility to resize the given filesystem and returns whether
+ * it is installed. The mode flags indicate if growing and/or shrinking resize is available if
+ * mounted/unmounted.
+ * Unknown filesystems or filesystems which do not support resizing result in errors.
+ *
+ * Returns: whether filesystem resize is available
+ */
+gboolean bd_fs_can_resize (const gchar *type, BDFsResizeFlags *mode, gchar **required_utility, GError **error);
+
+/**
+ * bd_fs_can_check:
+ * @type: the filesystem type to be tested for installed consistency check support
+ * @required_utility: (out) (transfer full): the utility binary which is required for checking (if missing i.e. returns FALSE but no error)
+ * @error: (out): place to store error (if any)
+ *
+ * Searches for the required utility to check the given filesystem and returns whether
+ * it is installed.
+ * Unknown filesystems or filesystems which do not support checking result in errors.
+ *
+ * Returns: whether filesystem check is available
+ */
+gboolean bd_fs_can_check (const gchar *type, gchar **required_utility, GError **error);
+
+/**
+ * bd_fs_can_repair:
+ * @type: the filesystem type to be tested for installed repair support
+ * @required_utility: (out) (transfer full): the utility binary which is required for repairing (if missing i.e. return FALSE but no error)
+ * @error: (out): place to store error (if any)
+ *
+ * Searches for the required utility to repair the given filesystem and returns whether
+ * it is installed.
+ * Unknown filesystems or filesystems which do not support reparing result in errors.
+ *
+ * Returns: whether filesystem repair is available
+ */
+gboolean bd_fs_can_repair (const gchar *type, gchar **required_utility, GError **error);
+
+
+/**
  * bd_fs_ext2_mkfs:
  * @device: the device to create a new ext2 fs on
  * @extra: (allow-none) (array zero-terminated=1): extra options for the creation (right now
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -361,6 +361,32 @@ gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
  */
 gboolean bd_fs_resize (const gchar *device, guint64 new_size, GError **error);
 
+ /**
+ * bd_fs_repair:
+ * @device: the device the file system of which to repair
+ * @error: (out): place to store error (if any)
+ *
+ * Repair filesystem on @device. This calls other fs repair functions from this
+ * plugin based on detected filesystem (e.g. bd_fs_xfs_repair for XFS). This
+ * function will return an error for unknown/unsupported filesystems.
+ *
+ * Returns: whether the file system on @device was successfully repaired or not
+ */
+gboolean bd_fs_repair (const gchar *device, GError **error);
+
+/**
+ * bd_fs_check:
+ * @device: the device the file system of which to check
+ * @error: (out): place to store error (if any)
+ *
+ * Check filesystem on @device. This calls other fs check functions from this
+ * plugin based on detected filesystem (e.g. bd_fs_xfs_check for XFS). This
+ * function will return an error for unknown/unsupported filesystems.
+ *
+ * Returns: whether the file system on @device passed the consistency check or not
+ */
+gboolean bd_fs_check (const gchar *device, GError **error);
+
 /**
  * BDFsResizeFlags:
  * Flags indicating whether a filesystem resize action supports growing and/or
--- a/src/plugins/fs.c
+++ b/src/plugins/fs.c
@@ -1517,20 +1517,8 @@ static gboolean xfs_resize_device (const gchar *device, guint64 new_size, const 
     return success;
 }
 
-/**
- * bd_fs_resize:
- * @device: the device the file system of which to resize
- * @new_size: new requested size for the file system (if 0, the file system is
- *            adapted to the underlying block device)
- * @error: (out): place to store error (if any)
- *
- * Resize filesystem on @device. This calls other fs resize functions from this
- * plugin based on detected filesystem (e.g. bd_fs_xfs_resize for XFS). This
- * function will return an error for unknown/unsupported filesystems.
- *
- * Returns: whether the file system on @device was successfully resized or not
- */
-gboolean bd_fs_resize (const gchar *device, guint64 new_size, GError **error) {
+static gboolean device_operation (const gchar *device, BDFsOpType op, guint64 new_size, GError **error) {
+    const gchar* op_name = NULL;
     g_autofree gchar* fstype = NULL;
 
     fstype = bd_fs_get_fstype (device, error);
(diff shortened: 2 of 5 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -361,6 +361,62 @@ gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
...
+typedef enum {
+    BD_FS_NO_SUPPORT,
+    BD_FS_OFFLINE,
+    BD_FS_OFFLINEGROW,
+    BD_FS_ONLINE,
+    BD_FS_ONLINEGROW,
+    BD_FS_ONLINE_AND_OFFLINE,
+    BD_FS_ONLINEGROW_AND_OFFLINE,
+    BD_FS_ONLINEGROW_AND_OFFLINEGROW,
+} BDFsOnlineOffline;
```
vpodzime: These need to have values assigned. And I think clever values can be chosen (powers of 2) that would allow the flags to be combined with bitwise OR. That way not so many would be needed.

vpodzime: They should also be documented.

src/lib/plugin_apis/fs.api
```diff
@@ -361,6 +361,62 @@ gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
...
+
+/**
+ * bd_fs_can_resize:
+ * @type: the filesystem type to be tested for installed resize support
+ * @mode: (out): how the filesystem allows resizing (e.g. online, offline, growing or not at all)
+ * @required_utility: (out): the utility binary which is required for resizing (if any)
```
vpodzime: What is the purpose of returning the utility back? I hope one day we won't need a utility for everything and then this will cause problems.

vpodzime: And if we really want to return it, it needs to have ownership transfer defined (and possibly being ``strdup()``ed on return).

src/plugins/fs.c
```diff
@@ -1484,6 +1484,116 @@ gboolean bd_fs_resize (const gchar *device, guint64 new_size, GError **error) {
...
+        *required_utility = "";
+        *mode = BD_FS_OFFLINE;
+        return TRUE;
+    }
+
+    if (*required_utility == NULL) {
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_NOT_SUPPORTED,
+                     "Resizing filesystem '%s' is not supported.", type);
+        return FALSE;
+    }
```
vpodzime: I think this should be a final ``else`` part of the above ``if``.

src/lib/plugin_apis/fs.api
pothos: Hello, thanks for the review.
My idea was to only return the utility binary if it is required (i.e. empty for vfat) so that if FALSE is returned and no error is set, it is clear that only the binary is missing and could be installed with PackageKit etc.

I thought since the default is (transfer none) and only static strings needed it would be ok like that. But yes, ownership transfer is what people would expect and return NULL instead of an empty string for vfat.

vpodzime: > My idea was to only return the utility binary if it is required (i.e. empty for vfat) so that if FALSE is returned and no error is set, it is clear that only the binary is missing and could be installed with PackageKit etc.

I see now. Good idea! I just somehow didn't "derive" that from the documentation. Then I think the utility should only be returned if it's actually missing (and the documentation should say that).

pothos:
The resize conditions are changed. But I did not make up a lookup table for the utilities yet because this would also affect the other functions which use them. I am also not sure if this is more like an overoptimization given the fact that the binary names are unlikely to change?

src/lib/plugin_apis/fs.api
```diff
@@ -362,6 +362,63 @@ gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
...
+
+/**
+ * bd_fs_can_resize:
+ * @type: the filesystem type to be tested for installed resize support
+ * @mode: (out): flags for allowed resizing (i.e. growing/shrinking support for online/offline)
+ * @required_utility: (out) (transfer full): the utility binary which is required for resizing (if missing i.e. returns FALSE but no error)
+ * @error: (out): place to store error (if any)
+ *
+ * Searches for the required utility to resize the given filesystem and returns whether
+ * it is installed. The mode flags indicate if growing and/or shriking resize is available if
```
vpodzime: *shriking*

vpodzime:
> But I did not make up a lookup table for the utilities yet because this would also affect the other functions which use them. I am also not sure if this is more like an overoptimization given the fact that the binary names are unlikely to change?

it's not about changing names of utilities, but rather adding support for more file systems and sharing code between the functions. They are all like:
"Based on the given *FS type* and *checked functionality*, check whether the particular utility is available".

I think the code should be generic and just combined with data like:
```
{{"xfs", "xfs_growfs", "xfs_repair"},
{"ext2", "resize2fs", "e2fsck"},
...}
```

pothos:
Hi,
maybe the struct is better because it can be extended to be also useful for other functions in future.
Since only two of the surrounding functions could share a small amount of code I just left this.
Regards,
Kai

vpodzime:
> Since only two of the surrounding functions could share a small amount of code I just left this.

If you move the *mode* resolving in ``can_resize()`` to the end, all three functions are doing exactly the same with the only difference ``resize->check->repair``.

vpodzime:
We definitely need tests for these new functions.

vpodzime:
@pothos, I don't really want to torture you with all the nitpicks and details. Would you mind if we created some special branch for these changes upstream so that they could be merged as PRs to that new branch and then I would go and clean the changes a bit so that they can be merged to *master*? That said, I'm not going to write all the tests, they have to be part of the PR.

pothos:
@vpodzime It's ok like this, no problem for me. Maybe just for this case of sharing code I'm not sure how to solve it properly when they need to return different things.

pothos:
Ah, G_STRUCT_OFFSET might help…

vpodzime:
> Maybe just for this case of sharing code I'm not sure how to solve it properly when they need to return different things.

Have a function like:
```
static gboolean check_fs_operation (const gchar *fs_type, OpType op, gchar **required_utility, GError **error);
```
where ``OpType`` is an internal *enum* type (``RESIZE``, ``REPAIR``, ``CHECK``). ``bd_fs_can_resize()`` would then do:
```
success = check_fs_operation (type, RESIZE_OP, required_utility, error);
/* determine and set mode here */
return success;
```
and the other two functions would just call ``check_fs_operation()``.

pothos:
Should I force push here if there are no further suggestions or make a new PR?

src/plugins/fs.c
```diff
@@ -47,6 +47,29 @@
...
+};
+
+const BDFSInfo *
+bd_fs_get_info (const gchar *type)
+{
+    guint n;
+
+    g_return_val_if_fail (type != NULL, NULL);
+
+    for (n = 0; n < sizeof(bd_fs_info)/sizeof(BDFSInfo); n++) {
```
vpodzime: You can declare ``n`` here just for the cycle.

vpodzime: Also I think it would be better and more reliable to terminate the ``bd_fs_info`` array with ``NULL`` and make use of it in this cycle. The compiler can do some alignments possibly making the ``sizeof(bd_fs_info)/sizeof(BDFSInfo)`` calculation wrong.

src/plugins/fs.c
```diff
@@ -1486,39 +1571,142 @@ static gboolean xfs_resize_device (const gchar *device, guint64 new_size, const
...
-    } else {
+    if (required_utility != NULL)
+        *required_utility = NULL;
+
+    if (mode != NULL)
+        *mode = 0;
+
+    fsinfo = bd_fs_get_info (fs_type);
+    if (fsinfo != NULL)
+        exec_util = G_STRUCT_MEMBER (const gchar *, fsinfo, exec_util_field);
```
vpodzime: I believe this function could be much less "magical" and avoid using the ``G_STRUCT`` macros if the steps were reordered -- i.e. get the ``fs_info`` first and then extract the utility from it in a ``switch``.

src/plugins/fs.c
```diff
@@ -47,6 +47,29 @@
...
+const BDFSInfo bd_fs_info[] = {
+    {"xfs", "xfs_db", "xfs_repair", "xfs_growfs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW},
+    {"ext2", "e2fsck", "e2fsck", "resize2fs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK},
+    {"ext3", "e2fsck", "e2fsck", "resize2fs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK},
+    {"ext4", "e2fsck", "e2fsck", "resize2fs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK},
+    {"vfat", "fsck.vfat", "fsck.vfat", "", BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK}
+};
+
+const BDFSInfo *
+bd_fs_get_info (const gchar *type)
```
vpodzime: These should be static. And please note that it's not necessary to use the ``bs_fs`` prefix for internal symbols (actually, it's not desirable because using the prefix may lead to a confusion).

src/plugins/fs.c
```diff
@@ -1486,39 +1571,142 @@ static gboolean xfs_resize_device (const gchar *device, guint64 new_size, const
...
+ * plugin based on detected filesystem (e.g. bd_fs_xfs_check for XFS). This
+ * function will return an error for unknown/unsupported filesystems.
+ *
+ * Returns: whether the file system on @device passed the consistency check or not
+ */
+gboolean bd_fs_check (const gchar *device, GError **error) {
+    return _do_operation (device, BD_FS_CHECK, 0, error);
+}
+
+static gboolean _check_fs_operation (const gchar *fs_type, BDFsOpType op, gchar **required_utility, BDFsResizeFlags *mode, GError **error) {
```
vpodzime: No need to use ``_`` at the beginning of the function's name, it's ``static``.

src/plugins/fs.h
```diff
@@ -82,6 +82,42 @@ gboolean bd_fs_mount (const gchar *device, const gchar *mountpoint, const gchar
...
+    const gchar *resize_util;
+    BDFsResizeFlags resize_mode;
+} BDFSInfo;
+
+typedef enum {
+    BD_FS_RESIZE,
+    BD_FS_REPAIR,
+    BD_FS_CHECK
+} BDFsOpType;
+
```
vpodzime: Now I'm confused. Are the above ``struct`` and ``enum`` supposed to be public/part of the API? I don't think they should be.

tests/fs_test.py
```diff
@@ -844,6 +844,69 @@ def test_vfat_resize(self):
...
+        avail, util = BlockDev.fs_can_check("xfs")
+        os.environ["PATH"] = old_path
+        self.assertFalse(avail)
+        self.assertEqual(util, "xfs_db")
+
+        avail = BlockDev.fs_can_check("vfat")
+        self.assertTrue(avail)
+
+        with self.assertRaises(GLib.GError):
+            BlockDev.fs_can_check("nilfs2")
```
vpodzime: Please add at least some simple tests for the generic check and repair functions too.

vpodzime:
> Should I force push here if there are no further suggestions or make a new PR?

Please (reasonably) squash and force-push the changes once the changes are actually approved.

vpodzime:
Jenkins, this is ok to test.

pothos:
Ok, thanks, I've updated the two commits.

src/plugins/fs.c
```diff
@@ -47,6 +47,51 @@
...
+    {"vfat", "fsck.vfat", "fsck.vfat", "", BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK},
+    {NULL, NULL, NULL, NULL, 0}
+};
+
+static const BDFSInfo *
+get_fs_info (const gchar *type)
+{
+    g_return_val_if_fail (type != NULL, NULL);
+
+    for (guint n = 0; fs_info[n].type != NULL; n++) {
```
pothos: The test fails with the hint that this is only allowed in C99 - should I revert back or will you adjust the CI?

vpodzime: I noticed that, but it only happens on RHEL/CentOS7. Please keep this in place, I'll change how we compile libblockdev so that this works fine.

vpodzime: https://github.com/storaged-project/libblockdev/pull/247

vpodzime:
Jenkins, test this, please.

vpodzime:
I think we also need #250, @pothos, please take a look.

PR #238, opened 2017-06-19
Add partition resize function

pothos:
The new function bd_part_resize_part allows to
change the size of a partition and guarantees
that it does not move and has at least the requested
size allocated. It also supports resizing to maximum.

Should I write tests as well or is it anyway better if they are not written by the one writing the code?

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -702,10 +702,98 @@ BDPartSpec* bd_part_get_best_free_region (const gchar *disk, BDPartType type, gu
     return ret;
 }
 
+static PedConstraint* prepare_alignment_constraint (PedDevice *dev, PedDisk *disk, BDPartAlign align, gint *orig_flag_state) {
+    if (align == BD_PART_ALIGN_OPTIMAL) {
+        /* cylinder alignment does really weird things when turned on, let's not
+           deal with it in 21st century (the flag is reset back in the end) */
+        if (ped_disk_is_flag_available (disk, PED_DISK_CYLINDER_ALIGNMENT)) {
+            *orig_flag_state = ped_disk_get_flag (disk, PED_DISK_CYLINDER_ALIGNMENT);
+            ped_disk_set_flag (disk, PED_DISK_CYLINDER_ALIGNMENT, 0);
+        }
+        return ped_device_get_optimal_aligned_constraint (dev);
+    } else if (align == BD_PART_ALIGN_MINIMAL)
+        return ped_device_get_minimal_aligned_constraint (dev);
+    else
+        return NULL;
+}
+
+static void finish_alignment_constraint (PedDisk *disk, gint orig_flag_state) {
+    if (ped_disk_is_flag_available (disk, PED_DISK_CYLINDER_ALIGNMENT)) {
+        ped_disk_set_flag (disk, PED_DISK_CYLINDER_ALIGNMENT, orig_flag_state);
+    }
+}
+
+static gboolean resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk, guint64 size, BDPartAlign align, GError **error) {
+    PedConstraint *constr = NULL;
+    PedGeometry *geom;
+    gint orig_flag_state = 0;
+    PedSector start;
+    PedSector end;
+    PedSector new_size = 0;
+    gint status = 0;
+
+    constr = prepare_alignment_constraint (dev, disk, align, &orig_flag_state);
+    start = part->geom.start;
+
+    if (!constr)
+        constr = ped_constraint_any (dev);
+
+    geom = ped_disk_get_max_partition_geometry (disk, part, constr);
+    if (!ped_geometry_set_start (geom, start)) {
+        set_parted_error (error, BD_PART_ERROR_FAIL);
+        g_prefix_error (error, "Failed to set partition start on device '%s'", dev->path);
+        ped_constraint_destroy (constr);
+        ped_geometry_destroy (geom);
+        finish_alignment_constraint (disk, orig_flag_state);
+        return FALSE;
+    }
+    if (size == 0) {
+        new_size = geom->length;
+    } else {
+        new_size = (size + dev->sector_size - 1) / dev->sector_size;
+    }
+
+    ped_geometry_destroy (geom);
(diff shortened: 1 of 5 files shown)


vpodzime:
> Should I write tests as well or is it anyway better if they are not written by the one writing the code?

Yes, add the tests, please.

src/plugins/part.c
```diff
@@ -984,6 +984,171 @@ gboolean bd_part_delete_part (const gchar *disk, const gchar *part, GError **err
...
+        ped_disk_set_flag (ped_disk, PED_DISK_CYLINDER_ALIGNMENT, orig_flag_state);
+    }
+
+    ped_disk_destroy (ped_disk);
+    ped_device_destroy (dev);
+    ped_geometry_destroy (geom);
+    bd_utils_report_finished (progress_id, "Completed");
+
+    return ret;
+}
```
vpodzime: I think this function should use the ``static add_part_to_disk`` function (possibly enhanced). And similarly, it should use a new ``static`` function doing what ``bd_part_delete_part()`` does except for committing changes to disk.

tests/part_test.py
```diff
@@ -912,6 +912,69 @@ def test_get_part_by_pos(self):
...
+class PartCreateResizePartCase(PartTestCase):
+    def test_create_resize_part_simple(self):
+        """Verify that it is possible to create and resize a parition"""
+
+        # we first need a partition table
+        succ = BlockDev.part_create_table (self.loop_dev, BlockDev.PartTableType.MSDOS, True)
+        self.assertTrue(succ)
+
+        # create a maximal partition
+        ps = BlockDev.part_create_part (self.loop_dev, BlockDev.PartTypeReq.NORMAL, 2048*512, 0, BlockDev.PartAlign.OPTIMAL)
```
vpodzime: Please use a size that can be easily recognized -- i.e. a multiple of (a power of) 1024.

tests/part_test.py
```diff
@@ -912,6 +912,69 @@ def test_get_part_by_pos(self):
...
+        initial_start = ps.start
+        initial_size = ps.size
+
+        new_size = 20 * 1000**2  # resize to 20 MB
```
vpodzime: ``20 * 1024**2`` would be better here.

tests/part_test.py
```diff
@@ -912,6 +912,69 @@ def test_get_part_by_pos(self):
...
+        succ = BlockDev.part_resize_part (self.loop_dev, ps.path, new_size, BlockDev.PartAlign.OPTIMAL)
+        self.assertTrue(succ)
+        ps = BlockDev.part_get_part_spec(self.loop_dev, ps.path)
+        self.assertEqual(initial_start, ps.start)  # offset must not be moved
+        self.assertGreaterEqual(ps.size, new_size)  # at least the requested size,
+        self.assertLess(ps.size, new_size + 8 * 1024**2)  # but also not too big
```
vpodzime: 8 MiB is quite a big difference from 20 MiB. AFAIK, we have all the test disks set to have optimal alignment ~1 MiB so there should never be such a big difference.

tests/part_test.py
```diff
@@ -912,6 +912,69 @@ def test_get_part_by_pos(self):
...
+
+        # resize to maximum with no alignment
+        succ = BlockDev.part_resize_part (self.loop_dev, ps.path, 0, BlockDev.PartAlign.NONE)
+        self.assertTrue(succ)
+        ps = BlockDev.part_get_part_spec(self.loop_dev, ps.path)
+        new_size = ps.size
+        self.assertEqual(initial_start, ps.start)
```
vpodzime: This should also check ``new_size``/``ps.size``.

tests/part_test.py
pothos: The idea was that the input could not be a multiple of the block size, so I think it's good to not only test for MiB but also MB.

vpodzime: That's a good idea. Just please add a comment about that then.

tests/part_test.py
pothos: The problem is in libparted which can give a smaller size for maximizing a partition with no alignment than with optimal.

src/plugins/part.c
pothos: It was not possible for me to find a proper solution that way and updating the geometry is I think more reliable to preserve the partition properties than copying flags and labels.
Also the way add_part_to_disk added maximum sized partitions was problematic and buggy if a maximal partition should be placed in front of existing one (all my tries to fix it still failed my test_create_resize_part_two).
From my point of view this is corrected in a good way now by using ped_disk_get_max_partition_geometry which requires an existing partition, so add_part_to_disk makes a new partition with temporary size and then calls the resize function. That way the partition is guaranteed to have the requested start position (plus possible alignment in only one way during creation) and at least the requested size.
(I chose 1 MiB as initial size for partitions as needed for logical partitions.)

vpodzime: Fair enough, this makes sense. Thanks for the details.

src/plugins/part.c
```diff
@@ -702,74 +702,152 @@ BDPartSpec* bd_part_get_best_free_region (const gchar *disk, BDPartType type, gu
...
+        return NULL;
+}
+
+static void _finish_alignment_constraint (PedDisk *disk, gint orig_flag_state) {
+    if (ped_disk_is_flag_available (disk, PED_DISK_CYLINDER_ALIGNMENT)) {
+        ped_disk_set_flag (disk, PED_DISK_CYLINDER_ALIGNMENT, orig_flag_state);
+    }
+}
+
+static gboolean _resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk, guint64 size, BDPartAlign align, GError **error) {
```
vpodzime: Please remove the ``_`` from the start of the functions' names. They are static, there's no point in marking them as internal with any other mechanism.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Feel free to ignore the *ci/x86_64/f_rawhide* test failure here, it's a magically failing test not related to the changes being done here.

vpodzime:
Jenkins, test this, please.

PR #239, opened 2017-06-19
Fixed include for libvolume_key.h

gtristan:
This is volume_key/libvolume_key.h, not libvolume_key.h.

configure.ac searches for the header in the correct location `volume_key/libvolume_key.h`, and then crypto.c tries to include it as `#include <libvolume_key.h>`, so compilation fails even after installing volume_key.

### DIFF
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -21,7 +21,7 @@
 #include <glib.h>
 #include <libcryptsetup.h>
 #include <nss.h>
-#include <libvolume_key.h>
+#include <volume_key/libvolume_key.h>
 #include <sys/fcntl.h>
 #include <sys/ioctl.h>
 #include <linux/random.h>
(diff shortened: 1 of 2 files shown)


src/plugins/crypto.c
```diff
@@ -21,7 +21,7 @@
 #include <glib.h>
 #include <libcryptsetup.h>
 #include <nss.h>
-#include <libvolume_key.h>
+#include <volume_key/libvolume_key.h>
```
vpodzime: This is covered by ``-I/usr/include/volume_key`` added to ``CFLAGS`` in *src/plugins/Makefile.am*. Please remove that from there.

gtristan: Ah sorry I missed that bit, pushed an additional commit to remove that to this branch.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

PR #241, opened 2017-06-21
Size in bytes for xfs_resize_device

pothos:
The bd_fs_resize function uses the xfs_resize_device
helper function with size in bytes, so instead of
forwarding this to bd_fs_xfs_resize which takes the
size in blocks, it has to be converted correctly
within xfs_resize_device using the block size.


vpodzime:
Jenkins, this is ok to test.

PR #242, opened 2017-06-21
Fix read-only device mounting (with new libmount)

vpodzime:
This should fix failing *udisks* tests.


vpodzime:
Merging this ASAP in the hope a new build with this fixes the failing *udisks* tests.

PR #250, opened 2017-06-29
Do not strictly require all FS utilities

vpodzime:
We now have code for checking if FS operations are
supported/available or not. Having the plugin to fail to load
when not everything it could possibly need is available is
pointless. We need a much better way to do the checks in plugins
in general. But for now let's just make the FS plugin usable.

Related: #217


pothos:
Should we include an additional way to test if setting the label is supported (+ generic label set function?) or is it okay to try it with the current label and in case it fails it means that setting is simply not supported.

vpodzime:
> Should we include an additional way to test if setting the label is supported (+ generic label set function?)

Yes, that would definitely be nice and consistent with the rest of the functionality.

pothos:
https://github.com/storaged-project/libblockdev/pull/251

PR #251, opened 2017-06-29
Query setting FS label support and generic relabeling

pothos:
Introduces bd_fs_can_set_label to query for installed
tooling. Adds the generic function bd_fs_set_label to
set the filesystem label of a device.

The test case for XFS fails if a space is contained like "new label" - what's wrong there?

### DIFF
--- a/docs/libblockdev-sections.txt
+++ b/docs/libblockdev-sections.txt
@@ -430,9 +430,11 @@ bd_fs_get_mountpoint
 bd_fs_resize
 bd_fs_repair
 bd_fs_check
+bd_fs_set_label
 bd_fs_can_resize
 bd_fs_can_check
 bd_fs_can_repair
+bd_fs_can_set_label
 bd_fs_ext2_check
 bd_fs_ext2_get_info
 bd_fs_ext2_info_copy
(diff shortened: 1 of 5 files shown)


pothos:
Ah, xfs_admin seems broken in this regard. Creating a filesystem through mkfs.xfs with "new label" is no problem.

pothos:
(I've changed it to underscore like in the other tests for now.)

vpodzime:
Jenkins, this is ok to test.

vpodzime:
I'm sorry, but I don't follow

> Ah, xfs_admin seems broken in this regard. Creating a filesystem through mkfs.xfs with "new label" is no problem.

and

> (I've changed it to underscore like in the other tests for now.)

We have tests for XFS (re)labeling and they seem to pass just fine.

pothos:
xfs_admin does not accept labels containing the space character while mkfs.xfs does, so I had to change the test from "new label" to "new_label".

vpodzime:
> xfs_admin does not accept labels containing the space character while mkfs.xfs does, so I had to change the test from "new label" to "new_label".

Oh, that's okay.

PR #252, opened 2017-06-30
Allow non-source directory builds and no-plugin builds

pothos:
Hello,

what do you think about that?
This enables to build with e.g. jhbuild/gnome-continuous. The autogen.sh comes from GNOME Disks and could be improved I guess(?). I also ran into a linking error if all plugins were disabled.

Best regards,
Kai

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,7 +1,16 @@
 #!/bin/bash -e
-[ -d m4 ] || mkdir m4
-libtoolize --copy --force
-aclocal -I m4
-autoconf
-automake --foreign --add-missing --copy
-rm -rf autom4te.cache
+test -n "$srcdir" || srcdir=`dirname "$0"`
+test -n "$srcdir" || srcdir=.
+
+pushd $srcdir
+
+(test -f configure.ac) || {
+        echo "*** ERROR: Directory "\`$srcdir\'" does not look like the top-level project directory ***"
+        exit 1
+}
+
+PKG_NAME=`autoconf --trace 'AC_INIT:$1' configure.ac`
+
+aclocal --install || exit 1
+autoreconf --verbose --force --install -Wno-portability || exit 1
+popd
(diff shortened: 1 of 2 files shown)


vpodzime:
Jenkins, this is ok to test.

autogen.sh
```diff
@@ -1,7 +1,33 @@
...
+        echo "*** WARNING: I am going to run \`configure' with no arguments." >&2
+        echo "*** If you wish to pass any to it, please specify them on the" >&2
+        echo "*** \`$0\' command line." >&2
+        echo "" >&2
+fi
+
+# libtoolize --copy --force --install
+aclocal --install || exit 1
+autoreconf --verbose --force --install -Wno-portability || exit 1
+cd $olddir
```
vpodzime: Can't this use ``pushd`` and ``popd`` instead of ``cd`` and remembering the ``$olddir``?

autogen.sh
```diff
@@ -1,7 +1,33 @@
...
+if [ "$NOCONFIGURE" = "" ]; then
```
vpodzime: This shouldn't be the default.

vpodzime:
@pothos, could you please update this so that it can be merged?

pothos:
(Same additional jhbuild declaration needed as the respective PR in libbytesize)

autogen.sh
```diff
@@ -1,7 +1,17 @@
...
-libtoolize --copy --force
-aclocal -I m4
-autoconf
-automake --foreign --add-missing --copy
-rm -rf autom4te.cache
+test -n "$srcdir" || srcdir=`dirname "$0"`
+test -n "$srcdir" || srcdir=.
+
+pushd .
+cd $srcdir
```
vpodzime: This should be ``pushd $srcdir``, which does the same as the two lines above.

PR #253, opened 2017-07-03
Specify tolerance for partition size

pothos:
If the caller of the partition creation and
resize functions is not aware of size reductions
due to alignment it should still be able to give
the theoretical free space as size argument.
This should be possible by a 3 MiB tolerance which
the final partition is allowed to be smaller.

This is to make it easier for an application that should rather use '0' to still be able to pass e.g. the device size to create one big partition.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -729,8 +729,14 @@ static gboolean resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk, 
     gint orig_flag_state = 0;
     PedSector start;
     PedSector end;
+    PedSector max_end;
     PedSector new_size = 0;
     gint status = 0;
+    /* it should be possible to pass the whole drive size a partition size,
+     * so -1 MiB for the first partition alignment,
+     * -1 MiB for creating this here as a logial partition
+     * and -1 MiB for end alingment */
+    const PedSector tolerance = (PedSector) (4 MiB /  dev->sector_size);
 
     constr = prepare_alignment_constraint (dev, disk, align, &orig_flag_state);
     start = part->geom.start;
@@ -753,6 +759,7 @@ static gboolean resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk, 
         new_size = (size + dev->sector_size - 1) / dev->sector_size;
     }
 
+    max_end = geom->end;
     ped_geometry_destroy (geom);
     geom = ped_geometry_new (dev, start, new_size);
     if (!geom) {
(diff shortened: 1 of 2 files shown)


vpodzime:
Jenkins, this is ok to test.

vpodzime:
> This is to make it easier for an application that should rather use '0' to still be able to pass e.g. the device size to create one big partition.

The problem is that this still doesn't work with the fixes here. We need a test for this.

vpodzime:
Jenkins, this is ok to test.

src/plugins/part.c
```diff
@@ -729,8 +729,10 @@ static gboolean resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk,
     gint orig_flag_state = 0;
     PedSector start;
     PedSector end;
+    PedSector max_end;
     PedSector new_size = 0;
     gint status = 0;
+    const PedSector tolerance = (PedSector) (4 MiB /  dev->sector_size);
```
vpodzime: We need a comment here why ``4 MiB``.

src/plugins/part.c
```diff
@@ -779,7 +793,7 @@ static gboolean resize_part (PedPartition *part, PedDevice *dev, PedDisk *disk,
         ped_constraint_destroy (constr);
         finish_alignment_constraint (disk, orig_flag_state);
         return FALSE;
-    } else if (part->geom.start != start || part->geom.length < new_size) {
+    } else if (part->geom.start != start || part->geom.length < new_size - tolerance) {
```
vpodzime: Would be great if we had different error messages for wrong start and size. Otherwise it's hard to see what went wrong from the error message.

tests/part_test.py
```diff
@@ -212,6 +212,22 @@ def test_create_part_minimal_start(self):
...
+        ps = BlockDev.part_create_part (self.loop_dev, BlockDev.PartTypeReq.NEXT, 0, 100 * 1024**2, BlockDev.PartAlign.OPTIMAL)
+
+    def test_full_device_auto(self):
+        # we first need a partition table
+        succ = BlockDev.part_create_table (self.loop_dev, BlockDev.PartTableType.GPT, True)
+        self.assertTrue(succ)
+
+        # create a maximal partition
+        ps = BlockDev.part_create_part (self.loop_dev, BlockDev.PartTypeReq.NEXT, 0, 0, BlockDev.PartAlign.OPTIMAL)
+
```
vpodzime: Please also try this with ``1`` as the partition start which should yield the same results (after deleting the partitions created with ``0``, of course :))

src/plugins/part.c
pothos: It's a conservative measure, actually it would also be possible to calculate 2-3 MiB, assuming that in a logical partition the shift will be + 1 MiB offset and maybe -1 MiB end alingment, maybe 3-4 if there is already an extended partition and the caller still passes the whole disk size.

vpodzime:
> There's still a problem on CentOS7 though. :(

Resolved in #256 which contains changes from this PR. Let's move there.

Issue #243, opened 2017-06-22
The resize tests should check if the resulting size is correct


vpodzime:
Otherwise they just verify that the functions don't crash. Not that they work as expected.

(hit in #241)

PR #254, opened 2017-07-03
Check resulting FS size in tests for generic FS resize

vpodzime:
Otherwise we don't catch issues in how the generic resize passes
everything to specificy resize functions.

Resolves: #243


vpodzime:
Jenkins, test this, please.

PR #255, opened 2017-07-05
Do not verify vfat FS' size after generic resize

vpodzime:
Otherwise trying to get the information about the file system may
fail randomly.


vpodzime:
Changed to:

Getting information about the file system after the resize
sometimes fails. Totally randomly. Which is a pity, but we
definitely don't want our CI results to become almost useless due
to this.

Also, other file systems do the resize precisely so we don't have
to be benevolent when comparing the requested and resulting
sizes. Which means better tests for the other file systems.

vpodzime:
Jenkins, test this, please.

vpodzime:
Merging to make CI results relevant again.

PR #256, opened 2017-07-05
Specify tolerance for partition size

vpodzime:
*Note: This PR was created as an extension of the PR #253 with two extra commits fixing the behavior on RHEL/CentOS7 and making things a bit more strict.*

If the caller of the partition creation and
resize functions is not aware of size reductions
due to alignment it should still be able to give
the theoretical free space as size argument.
This should be possible by a 3 MiB tolerance which
the final partition is allowed to be smaller.

This is to make it easier for an application that should rather use '0' to still be able to pass e.g. the device size to create one big partition.


vpodzime:
@pothos, what do you think about this?

pothos:
Nice, looks good and I hope that this is finally settled :)

Issue #203, opened 2017-06-05
Warning for missing command line tools


tasleson:
When running `make test` I see the following warnings repeated over and over again:

``` bash
** (process:25560): WARNING **: Cannot load the LVM plugin: The 'lvm' utility is not available

** (process:25560): WARNING **: Cannot load the LVM plugin: The 'thin_metadata_size' utility is not available
** (process:25560): WARNING **: Cannot load the BTRFS plugin: The 'btrfs' utility is not available
** (process:25560): WARNING **: Cannot load the swap plugin: The 'mkswap' utility is not available
** (process:25560): WARNING **: Cannot load the mpath plugin: The 'multipath' utility is not available
** (process:25560): WARNING **: Cannot load the DM plugin: The 'dmsetup' utility is not available
** (process:25560): WARNING **: Cannot load the MDRAID plugin: The 'mdadm' utility is not available
** (process:25560): WARNING **: Cannot load the kbd plugin: The 'make-bcache' utility is not available
** (process:25560): WARNING **: Cannot load the part plugin: The 'sgdisk' utility is not available
** (process:25560): WARNING **: Cannot load the part plugin: The 'sfdisk' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'mke2fs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'e2fsck' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'tune2fs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'dumpe2fs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'resize2fs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'mkfs.xfs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'xfs_db' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'xfs_repair' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'xfs_admin' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'xfs_growfs' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'mkfs.vfat' utility is not available
** (process:25560): WARNING **: Cannot load the FS plugin: The 'fatlabel' utility is not available
[...]

PR #259, opened 2017-07-21
Keep most utilities available for tests

vpodzime:
We only need to make sure specific utilities are not available in
the tests that require so. Otherwise we are getting many warnings
for missing utilities in the tests.

Resolves: #203


vpodzime:
Jenkins, test this, please.

PR #264, opened 2017-08-04
Wait for resized partition

pothos:
After the new partition table was committed
it takes some time until the partition block
device appears again. If the resize would not
have any effect this is skipped and otherwise
it is waited until the resize result is available.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -25,6 +25,8 @@
 #include <inttypes.h>
 #include <unistd.h>
 #include <sys/file.h>
+#include <sys/ioctl.h>
+#include <linux/fs.h>
 #include <blockdev/utils.h>
 #include <part_err.h>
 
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/plugins/part.c
```diff
@@ -1172,14 +1176,39 @@ gboolean bd_part_resize_part (const gchar *disk, const gchar *part, guint64 size
...
-    ret = disk_commit (ped_disk, disk, error);
+    new_size = ped_part->geom.length * dev->sector_size;
+    if (old_size != new_size) {
+        gint fd = 0;
+        gint wait = 0;
+        guint64 block_size = 0;
+
+        ret = disk_commit (ped_disk, disk, error);
+        /* wait for partition to appear with new size */
+        while (wait < 150) {
```
vpodzime: I take it this means 15 seconds? Would be nice to add a comment or something so that people don't have to do the maths every time they look at this code. Plus I think it's a very long timeout for a partition creation.

vojtechtrefny:
Jenkins, test this please.

PR #269, opened 2017-08-11
script for running tests

vojtechtrefny:


tests/run_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        os.environ['LD_LIBRARY_PATH'] = LIBDIRS
+        os.environ['GI_TYPELIB_PATH'] = GIDIR
+        os.environ['LIBBLOCKDEV_CONFIG_DIR'] = os.path.join(testdir, 'default_config')
+
+        try:
+            pyver = 'python3' if six.PY3 else 'python'
+            os.execv(sys.executable, [pyver] + sys.argv)
+        except OSError as e:
+            print('Failed re-exec with a new LD_LIBRARY_PATH and GI_TYPELIB_PATH: %s' % str(e))
+            sys.exit(1)
```
vpodzime: I think this should be done right at the beginning because everything above it is repeated again after the re-exec, right?

Makefile.am
```diff
@@ -61,44 +61,25 @@ pylint:
...
 test: all pylint
-	@rm -f tests/error_occurred;
-	@rm -f $(TEST_SUITE_LOG)
-	@touch $(TEST_SUITE_LOG)
-	for plugin in ${PLUGINS} ${LIBRARY_TESTS} ${UTILS_TESTS}; do \
-		$(MAKE) test-plugin-$$plugin || touch tests/error_occurred; \
-	done
-	$(MAKE) test-lvm-dbus || touch tests/error_occurred
-	@test ! -f tests/error_occurred
+	sudo $(TEST_PYTHON) tests/run_tests.py --lucky
```
vpodzime: ``--lucky`` should only be used in ``test-all``. Or at least that's how it was supposed to work before.

tests/run_tests.py
vojtechtrefny: Fixed

PR #270, opened 2017-08-14
Make sure the device is opened for libparted

vpodzime:
Sometimes it magically happens that libparted kills the whole
process trying to create a partition due to the

  assert(disk->dev->open_count > 0)

Which should obviously never happen because we are never opening
or closing the device explicitly just by means of
ped_device_get() and ped_device_destroy(). So this is probably a
bug in libparted, but we have no reproducer for it.

Let's just try to make sure the device is opened while we are
trying to commit the new partition table to it.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -196,6 +196,14 @@ static gboolean disk_commit (PedDisk *disk, const gchar *path, GError **error) {
        chance things will just work. If not, an error will be reported
        anyway with no harm. */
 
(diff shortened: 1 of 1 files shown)


vpodzime:
This is an attempt to fix the failures we have been seeing from time to time in udisks tests and which have also been reported by @martinpitt as https://github.com/storaged-project/udisks/issues/354

vpodzime:
Jenkins, test this, please.

PR #272, opened 2017-08-30
libfdisk-based part plugin - basics

vpodzime:
These commit implement the basic functionality of the *part* plugin using *libfdisk*.


vpodzime:
Reworked this to not introduce an alternative *part* plugin, but rather move a few first pieces to *libfdisk* instead.

vojtechtrefny:
Jenkins, test this please.

PR #277, opened 2017-09-08
Dynamic dependency checking and tech querying

vpodzime:
This PR will eventually do two things:

- [x] make checking for required utilities dynamic/on demand
- [x] implement the technology availability querying

But it also needs:

- [x] new tests
- [x] docs updates (functions and types)
- [x] assign functions to tech-mode combinations

PR #281, opened 2017-09-22
Skip btrfs tests on CentOS 7 aarch64

vojtechtrefny:
btrfs module is not available there.


vpodzime:
We should do a better job here. I created  the issue #282 for this to be improved when things from #277 are merged and usable here.

vojtechtrefny:
Jenkins, test this please.

PR #283, opened 2017-09-25
Better handle old and new zram sysfs api in tests

vojtechtrefny:
Use kernel version to pick right test instead of checking
distribution name/version.


vojtechtrefny:
Jenkins, test this please.

PR #289, opened 2017-10-11
Use system values in KbdTestBcacheStatusTest

vojtechtrefny:
Bcache stats in sysfs are sometimes wrong or take long time to
get updatedi after bcache creation. This test shouldn't fail just
because bcache is broken -- so instead of comparing stats returned
by "kbd_bcache_status" with some "expected" values, just compare
them from stats we read from sysfs. It isn't our problem if bcache
is broken.

### DIFF
--- a/src/plugins/kbd.c
+++ b/src/plugins/kbd.c
@@ -25,6 +25,7 @@
 #include <locale.h>
 #include <blockdev/utils.h>
 #include <stdio.h>
+#include <bs_size.h>
 
 #include "kbd.h"
 #include "check_deps.h"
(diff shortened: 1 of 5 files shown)


vojtechtrefny:
Update: The problem is not only in the tests -- sysfs sometimes says that `block_size` is `0.5k` instead of `512` so we need to parse this value with libblockdev.

tests/kbd_test.py
```diff
@@ -1,9 +1,11 @@
 import unittest
 import os
```
vpodzime: *updatedi* in the commit message.

tests/kbd_test.py
```diff
@@ -487,6 +489,14 @@ def test_bcache_get_set_mode(self):
         wipe_all(self.loop_dev, self.loop_dev2)
 
 class KbdTestBcacheStatusTest(KbdBcacheTestCase):
+
+    def _get_size(self, bcache_name):
+        cache_dir = '/sys/block/%s/bcache/cache' % bcache_name
+
+        # sum sizes from all caches
+        caches = ['%s/%s' % (cache_dir, d) for d in os.listdir(cache_dir) if re.match('cache[0-9]*$', d)]
```
vpodzime: I think it's better to use ``glob.glob()`` here.

src/plugins/kbd.c
```diff
@@ -1415,13 +1450,12 @@ BDKBDBcacheStats* bd_kbd_bcache_status (const gchar *bcache_device, GError **err
...
-    ret->block_size = get_number_from_file (path, error);
-    g_free (path);
-    if (*error) {
-        g_clear_error (error);
+    ret->block_size = get_bcache_block_size (bcache_device, &loc_error);
+    if (loc_error) {
         g_set_error (error, BD_KBD_ERROR, BD_KBD_ERROR_BCACHE_INVAL,
-                     "Failed to get 'block_size' for '%s' Bcache device", bcache_device);
+                     "Failed to get 'block_size' for '%s' Bcache device: %s",
+                     bcache_device, loc_error->message);
```
vpodzime: This can use the ``g_prefix_error()`` call, AFAICT.

tests/kbd_test.py
vojtechtrefny: Fixed

tests/kbd_test.py
vojtechtrefny: I took this from udisks tests -- https://github.com/storaged-project/udisks/blob/master/src/tests/dbus-tests/test_bcache.py#L56 -- glob would be better, but I think this is ok for tests.

src/plugins/kbd.c
vojtechtrefny: Changed to `g_propagate_prefixed_error`

PR #294, opened 2017-10-22
Support NTFS

pothos:
Hello,

this adds NTFS to the filesystem plugin.

Since I don't know if you have NTFS stuff installed on the CI, I just added these three basic test cases in a second commit which can be left out if not wanted. Resizing always needs the repairing before and afterwards, otherwise even basic things will refuse to run on a dirty-marked device.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -268,6 +268,51 @@ GType bd_fs_vfat_info_get_type () {
     return type;
 }
 
+/**
+ * BDFSNtfsInfo:
+ * @size: size of the filesystem in bytes
+ * @free_space: number of free space in the filesystem in bytes
+ */
+typedef struct BDFSNtfsInfo {
+    guint64 size;
+    guint64 free_space;
+} BDFSNtfsInfo;
+
+/**
+ * bd_fs_ntfs_info_copy: (skip)
+ *
+ * Creates a new copy of @data.
+ */
+BDFSNtfsInfo* bd_fs_ntfs_info_copy (BDFSNtfsInfo *data) {
+    BDFSNtfsInfo *ret = g_new0 (BDFSNtfsInfo, 1);
+
+    ret->size = data->size;
+    ret->free_space = data->free_space;
+
+    return ret;
+}
+
+/**
+ * bd_fs_ntfs_info_free: (skip)
+ *
+ * Frees @data.
+ */
+void bd_fs_ntfs_info_free (BDFSNtfsInfo *data) {
+    g_free (data);
+}
+
+GType bd_fs_ntfs_info_get_type () {
+    static GType type = 0;
+
+    if (G_UNLIKELY(type == 0)) {
+        type = g_boxed_type_register_static("BDFSNtfsInfo",
+                                            (GBoxedCopyFunc) bd_fs_ntfs_info_copy,
+                                            (GBoxedFreeFunc) bd_fs_ntfs_info_free);
+    }
+
+    return type;
+}
+
 typedef enum {
     BD_FS_TECH_GENERIC = 0,
     BD_FS_TECH_MOUNT,
@@ -276,6 +321,7 @@ typedef enum {
     BD_FS_TECH_EXT4,
     BD_FS_TECH_XFS,
     BD_FS_TECH_VFAT,
+    BD_FS_TECH_NTFS
 } BDFSTech;
 
 typedef enum {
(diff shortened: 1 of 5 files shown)


src/plugins/fs.c
```diff
@@ -2966,3 +3017,217 @@ gboolean bd_fs_vfat_resize (const gchar *device, guint64 new_size, GError **erro
...
+
+    ret = g_new0 (BDFSNtfsInfo, 1);
+    lines = g_strsplit (output, "\n", 0);
+    g_free (output);
+    line_p = lines;
+    /* find the beginning of the (data) section we are interested in */
+    while (*line_p && !g_str_has_prefix (*line_p, "bytes per volume"))
+        line_p++;
+    if (!line_p) {
+        /* error is already populated */
```
vpodzime: A copy-paste leftover?

src/plugins/fs.c
```diff
@@ -2966,3 +3017,217 @@ gboolean bd_fs_vfat_resize (const gchar *device, guint64 new_size, GError **erro
...
+        g_strfreev (lines);
+        bd_fs_ntfs_info_free (ret);
+        return FALSE;
+    }
+
+    /* extract data from something like this: "bytes per volume        : 998240256" */
+    val_start = strchr (*line_p, ':');
+    val_start++;
+    while (isspace (*val_start))
+        val_start++;
```
vpodzime: I think this is not necessary, spaces before the number should be fine. But please verify if that's really the case. :)

src/plugins/fs.c
```diff
@@ -2966,3 +3017,217 @@ gboolean bd_fs_vfat_resize (const gchar *device, guint64 new_size, GError **erro
...
+        g_strfreev (lines);
+        bd_fs_ntfs_info_free (ret);
+        return FALSE;
+    }
+
+    /* extract data from something like this: "bytes of free space     : 992759808" */
+    val_start = strchr (*line_p, ':');
+    val_start++;
+    while (isspace (*val_start))
+        val_start++;
```
vpodzime: The same applies here, of course.

src/lib/plugin_apis/fs.api
```diff
@@ -997,4 +1042,91 @@ BDFSVfatInfo* bd_fs_vfat_get_info (const gchar *device, GError **error);
...
+ *
+ * Tech category: %BD_FS_TECH_NTFS-%BD_FS_TECH_MODE_SET_LABEL
+ */
+gboolean bd_fs_ntfs_set_label (const gchar *device, const gchar *label, GError **error);
+
+/**
+ * bd_fs_ntfs_resize:
+ * @device: the device the file system of which to resize
+ * @new_size: new requested size for the file system (if 0, the file system is
+ *            adapted to the underlying block device)
```
vpodzime: This should probably specify a unit.

tests/fs_test.py
```diff
@@ -1219,6 +1231,16 @@ def test_ext4_generic_resize(self):
...
 
+    def test_ntfs_generic_resize(self):
+        """Test generic resize function with an ntfs file system"""
+        def mkfs_prepare(drive, l):
+            return BlockDev.fs_ntfs_mkfs(drive, l) and BlockDev.fs_repair(drive)
+        def info_prepare(drive):
+            return BlockDev.fs_repair(drive) and BlockDev.fs_ntfs_get_info(drive)
+        self._test_generic_resize(mkfs_function=mkfs_prepare,
+                                  fs_info_func=info_prepare,
+                                  info_size_func=lambda fi: fi.size + 4096)
```
vpodzime: This ``+ 4096`` requires an explanation in a comment.

pothos:
Thanks for the feedback, I have updated it.

tests/fs_test.py
```diff
@@ -1219,6 +1231,21 @@ def test_ext4_generic_resize(self):
...
+    def test_ntfs_generic_resize(self):
+        """Test generic resize function with an ntfs file system"""
+        def mkfs_prepare(drive, l):
+            return BlockDev.fs_ntfs_mkfs(drive, l) and BlockDev.fs_repair(drive)
+        def info_prepare(drive):
+            return BlockDev.fs_repair(drive) and BlockDev.fs_ntfs_get_info(drive)
+        def expected_size(fi):
+            # man ntfsresize says "The filesystem size is set to be at least one
+            # sector smaller" (maybe depending on alignment as well?), thus on a
+            # loop device as in this test it is 4096 bytes smaller than requested
```
pothos: Is "loop device" ok as description or should it be called LIO device?

vpodzime: It's a LIO loop device so I don't see a problem with this. :)

src/plugins/fs.c
```diff
@@ -290,6 +325,12 @@ static UtilDep deps[DEPS_LAST] = {
...
     {"fsck.vfat", NULL, NULL, NULL},
+
+    {"mkntfs", NULL, NULL, NULL},
+    {"ntfsfix", NULL, NULL, NULL},
+    {"ntfsresize", NULL, NULL, NULL},
+    {"ntfslabel", NULL, NULL, NULL},
+    {"ntfscluster", NULL, NULL, NULL},
 };
 
 static guint32 fs_mode_util[][FS_MODE_LAST+1] = {
```
vpodzime: This array needs an extra row too.

vpodzime:
Jenkins, this is ok to test.

pothos:
Maybe I missed something: locally I get `** (process:8810): CRITICAL **: The function 'bd_fs_is_tech_avail' called, but not implemented!` when trying to run the tests

pothos:
Where should I put the `BlockDev.ensure_init()` in the tests?

pothos:
Actually I don't understand why it is throwing an error and not just returning False in ci/*/centos_7:
```
@unittest.skipIf(not BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.MKFS)
Error: g-bd-utils-exec-error-quark: The 'mkntfs' utility is not available (6)
```

pothos:
OK, now it should be better as it requires only the FS plugin. The way PluginSpec and is_tech_avail work in Python is a bit strange.

vpodzime:
That's because the tests only (re)inialize the libblockdev library in ``setUpClass()`` class methods of the test case classes. Which happens later than the evaluation of the decorators.

tests/fs_test.py
```diff
@@ -11,6 +11,18 @@
...
 
+fs_spec = BlockDev.PluginSpec()
+fs_spec.name = BlockDev.Plugin.FS
+BlockDev.ensure_init(require_plugins=[fs_spec])
+try:
+    ntfs_avail = BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.MKFS) and \
+                 BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.RESIZE) and \
+                 BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.REPAIR) and \
+                 BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.CHECK) and \
+                 BlockDev.fs_is_tech_avail(BlockDev.FSTech.NTFS, BlockDev.FSTechMode.SET_LABEL)
```
vpodzime: The modes can be combined with ``|`` to create a bitmask. So just a single call of the ``fs_is_tech_avail()`` function is needed here.

tests/fs_test.py
```diff
@@ -11,6 +11,18 @@
 
 from gi.repository import BlockDev, GLib
 
+fs_spec = BlockDev.PluginSpec()
+fs_spec.name = BlockDev.Plugin.FS
+BlockDev.ensure_init(require_plugins=[fs_spec])
```
vpodzime: We have intentionally moved library (re)initialization into the test case classes. Please move this to ``setUpClass`` below and skip the tests by adding something like this to the beginning of the tests (methods):
```
if not self.ntfs_avail:
    self.skipTest("skipping NTFS: not available")
```

pothos:
Ah, thanks - I've searched for the wrong function names and overlooked it. It's corrected now.

PR #296, opened 2017-10-26
More fixes for issues discovered by coverity

vojtechtrefny:
Running `covscan` with `--profile=multi` (instead of `errata`) shows some more warnings.

### DIFF
--- a/scripts/boilerplate_generator.py
+++ b/scripts/boilerplate_generator.py
@@ -191,12 +191,12 @@ def get_includes_str(includes):
     return ret
 
 def get_funcs_info(fn_infos, module_name):
-    ret = "static const gchar const * const {0}_functions[] = {{\n".format(module_name)
+    ret = "static const gchar* const {0}_functions[] = {{\n".format(module_name)
     for info in fn_infos:
         ret += '    "{0.name}",\n'.format(info)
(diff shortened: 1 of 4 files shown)


scripts/boilerplate_generator.py
```diff
@@ -191,12 +191,12 @@ def get_includes_str(includes):
     return ret
 
 def get_funcs_info(fn_infos, module_name):
-    ret = "static const gchar const * const {0}_functions[] = {{\n".format(module_name)
+    ret = "static const gchar* {0}_functions[] = {{\n".format(module_name)
```
vpodzime: I believe this should have two ``const`` not just one.

scripts/boilerplate_generator.py
```diff
@@ -191,12 +191,12 @@ def get_includes_str(includes):
...
     for info in fn_infos:
         ret += '    "{0.name}",\n'.format(info)
     ret += "    NULL};\n\n"
 
-    ret += ("gchar const * const * get_{0}_functions (void) {{\n".format(module_name) +
+    ret += ("const gchar** get_{0}_functions (void) {{\n".format(module_name) +
```
vpodzime: This should have two ``const`` too just in a better order.

PR #299, opened 2017-11-06
Mark unstable tests as such

vpodzime:
So that they are not making results of our tests hard to work
with.


vpodzime:
See also https://github.com/storaged-project/udisks/pull/434

PR #302, opened 2017-11-16
Mark bcache tests as unstable

vpodzime:
They are quite often failing due to some issues in
kernel/udev/something that is not our business.


vojtechtrefny:
Jenkins, test this please.

PR #304, opened 2017-11-21
mpath plugin: Use libdmmp API.

cathay4t:
* Use libdmmp for these functions of mpath plugin:
    * `bd_mpath_flush_mpaths()`
    * `bd_mpath_reconfig()`
    * `bd_mpath_is_mpath_member()`
    * `bd_mpath_get_mpath_members()`

### DIFF
--- a/src/plugins/mpath.c
+++ b/src/plugins/mpath.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014  Red Hat, Inc.
+ * Copyright (C) 2014-2018  Red Hat, Inc.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -15,17 +15,55 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
  * Author: Vratislav Podzimek <vpodzime@redhat.com>
+ *         Gris Ge <fge@redhat.com>
+ */
+
+/*
+ * Notes on Code style:
+ *  * 4 spaces instead of tab as indention.
+ *  * Use this format of function block:
+ *      const char *something (int arg1) {
+ *      }
+ *  * Space before '(' unless it's in macro define.
+ *  * Use this format for if/switch/for block:
+ *      if (foo != boo) {
+ *          do_something ();
+ *      }
  */
 
 #include <glib.h>
-/* provides MAJOR, MINOR macros */
-#include <linux/kdev_t.h>
-#include <libdevmapper.h>
 #include <unistd.h>
 #include <blockdev/utils.h>
+#include <libdmmp/libdmmp.h>
+#include <string.h>
+#include <stdio.h>
 
 #include "mpath.h"
 #include "check_deps.h"
+
+#define _ERR_MSG_BUFF_SIZE    4096
+
+/*
+ * Just don't print to stdout or stderr, and save last error message
+ */
+static void _log_func (struct dmmp_context *ctx, int priority,
+                       const char *file, int line, const char *func_name,
+                       const char *format, va_list args);
+
+static const char *_dmmp_last_err_msg (struct dmmp_context *ctx);
+
+#define _dmmp_ctx_setup(ctx, err_msg_buff) \
+    do { \
+        ctx = dmmp_context_new (); \
+        if (ctx != NULL) \
+          { \
+            memset (err_msg_buff, 0, _ERR_MSG_BUFF_SIZE); \
+            dmmp_context_userdata_set (ctx, err_msg_buff); \
+            dmmp_context_log_func_set (ctx, _log_func); \
+          } \
+    } while (0)
+
+#define _UNUSED(x) (void)(x)
 
 /**
  * SECTION: mpath
@@ -39,25 +77,20 @@
 /**
  * bd_mpath_error_quark: (skip)
  */
-GQuark bd_mpath_error_quark (void)
-{
+GQuark bd_mpath_error_quark (void) {
     return g_quark_from_static_string ("g-bd-mpath-error-quark");
 }
 
 static volatile guint avail_deps = 0;
 static GMutex deps_check_lock;
 
-#define DEPS_MPATH 0
-#define DEPS_MPATH_MASK (1 << DEPS_MPATH)
-#define DEPS_MPATHCONF 1
+#define DEPS_MPATHCONF 0
 #define DEPS_MPATHCONF_MASK (1 << DEPS_MPATHCONF)
-#define DEPS_LAST 2
+#define DEPS_LAST 1
 
 static UtilDep deps[DEPS_LAST] = {
-    {"multipath", MULTIPATH_MIN_VERSION, NULL, "multipath-tools v([\\d\\.]+)"},
     {"mpathconf", NULL, NULL, NULL},
 };
-
 
 /**
  * bd_mpath_check_deps:
@@ -75,7 +108,8 @@ gboolean bd_mpath_check_deps () {
 
     for (i=0; i < DEPS_LAST; i++) {
         status = bd_utils_check_util_version (deps[i].name, deps[i].version,
-                                              deps[i].ver_arg, deps[i].ver_regexp, &error);
+                                              deps[i].ver_arg,
+                                              deps[i].ver_regexp, &error);
         if (!status)
             g_warning ("%s", error->message);
         else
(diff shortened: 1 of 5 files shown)


src/plugins/mpath.c
```diff
@@ -15,18 +15,57 @@
...
+        ctx = dmmp_context_new (); \
+        if (ctx != NULL) \
+          { \
+            memset(err_msg_buff, 0, _ERR_MSG_BUFF_SIZE); \
+            dmmp_context_userdata_set (ctx, err_msg_buff); \
+            dmmp_context_log_func_set (ctx, _log_func); \
+          } \
+    } while(0)
+
+#define _UNUSED(x) (void)(x)
```
vpodzime: Why do all these things start with an underscore?

src/plugins/mpath.c
```diff
@@ -69,24 +106,15 @@ static UtilDep deps[DEPS_LAST] = {
...
-        g_warning("Cannot load the mpath plugin");
+    gboolean ret = FALSE;
 
+    /* mpathconf doesn't report its version */
+    ret = bd_utils_check_util_version ("mpathconf", NULL, NULL, NULL, &error);
+    if (!ret && error)
+      {
+        g_warning ("Cannot load the mpath plugin: %s", error->message);
+        g_clear_error (&error);
+      }
```
vpodzime: This needs to preserve the ``avail_deps`` atomic manipulation in place.

src/plugins/mpath.c
```diff
@@ -110,7 +138,7 @@ gboolean bd_mpath_init () {
  *
  */
 void bd_mpath_close () {
-    /* nothing to do here */
+      /* nothing to do here */
```
vpodzime: Why this change?

src/plugins/mpath.c
```diff
@@ -156,360 +184,295 @@ gboolean bd_mpath_is_tech_avail (BDMpathTech tech, guint64 mode, GError **error)
  * Tech category: %BD_MPATH_TECH_BASE-%BD_MPATH_TECH_MODE_MODIFY
  */
 gboolean bd_mpath_flush_mpaths (GError **error) {
-    const gchar *argv[3] = {"multipath", "-F", NULL};
-    gboolean success = FALSE;
-    gchar *output = NULL;
+    struct dmmp_mpath **mps = NULL;
+    struct dmmp_context *ctx = NULL;
+    uint32_t mp_count = 0;
+    uint32_t i = 0;
```
vpodzime: Please use the *GLib* types here.

src/plugins/mpath.c
```diff
@@ -156,360 +184,295 @@ gboolean bd_mpath_is_tech_avail (BDMpathTech tech, guint64 mode, GError **error)
...
+    gboolean rc = FALSE;
+    const char *name = NULL;
+    int dmmp_rc = DMMP_OK;
+    char err_msg_buff[_ERR_MSG_BUFF_SIZE];
 
-    if (!check_deps (&avail_deps, DEPS_MPATH_MASK, deps, DEPS_LAST, &deps_check_lock, error))
-        return FALSE;
+    assert (error != NULL);
```
vpodzime: Please don't use ``assert()`` in a library.

vpodzime: We need some error to be set and ``FALSE`` be returned instead.

src/plugins/mpath.c
```diff
@@ -156,360 +184,295 @@ gboolean bd_mpath_is_tech_avail (BDMpathTech tech, guint64 mode, GError **error)
...
-        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_INVAL,
-                     "Failed to determine device name for '%s'",
-                     major_minor);
-        g_free (link);
-        return NULL;
+    _dmmp_ctx_setup (ctx, err_msg_buff);
+    if (ctx == NULL) {
+        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_NO_MEMORY,
+                     "No enough memory to create dmmp context");
+        goto out;
```
vpodzime: Please avoid the ``goto`` statement. I know it make things more complicated, but we have always been sacrificing that for avoiding ``goto``s in this library and this doesn't look like any special case that should end such efforts.

src/plugins/mpath.c
```diff
@@ -156,360 +184,295 @@ gboolean bd_mpath_is_tech_avail (BDMpathTech tech, guint64 mode, GError **error)
...
 
-    if (dm_task_get_info (task, &info) == 0) {
-        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_DM_ERROR,
-                     "Failed to get task info");
-        dm_task_destroy (task);
-        return FALSE;
+out:
+    if (dmmp_rc != DMMP_OK) {
+        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_FLUSH,
+                     "%d: %s", dmmp_rc, _dmmp_last_err_msg(ctx));
```
vpodzime: This should provide some more descriptive error messages based on what (step) exactly failed.

src/plugins/mpath.c
```diff
@@ -156,360 +184,295 @@ gboolean bd_mpath_is_tech_avail (BDMpathTech tech, guint64 mode, GError **error)
...
-    if (geteuid () != 0) {
-        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_NOT_ROOT,
-                     "Not running as root, cannot query DM maps");
-        bd_utils_report_finished (progress_id, (*error)->message);
-        return NULL;
+    GList *path_blk_names = NULL;
+    gchar *dup = NULL;
+    guint len = 0;
+    guint m = 0;
+    GList *cur_node = NULL;
```
vpodzime: Please use a ``GPtrArray`` here. And note that ``g_ptr_array_free (arr, FALSE)`` can be used to free the container while preserving the content which means the content doesn't have to be copied over elsewhere.

src/plugins/mpath.h
```diff
@@ -25,7 +28,6 @@ typedef enum {
     BD_MPATH_TECH_MODE_MODIFY = 1 << 1,
 } BDMpathTechMode;
 
-
```
vpodzime: no need for this change

cathay4t:
Fedora 27 and later has the `libdmmp` function required by this PR.

src/plugins/mpath.c
cathay4t: Just don't want to override existing macros (from other header) by accident.
Just a habit of keep internal functions start with an underscore.

src/plugins/mpath.c
cathay4t: Just a trivial miss type. Will remove it.

src/plugins/mpath.c
cathay4t: These are for `libdmmp` function only. Why we need `GLib`ize them?

src/plugins/mpath.c
cathay4t: Error cannot be set due to NULL pointer. Will just return `FALSE`.

src/plugins/mpath.c
cathay4t: Sure. Will do.

src/plugins/mpath.c
cathay4t: Will fix this by removing `goto` statement.

src/plugins/mpath.c
cathay4t: Just FYI: https://peerj.com/preprints/826v1/ 
```
We conclude that developers limit themselves to using 
goto appropriately in most cases, and not in an unrestricted 
manner like Dijkstra feared, thus suggesting that goto does 
not appear to be harmful in practice.
```

src/plugins/mpath.c
cathay4t: The memory holding the path device string is managed/owned by `struct dmmp_mpath` where we cannot only partly free(free all memory except the path device string). Hence `strdup()` is required.

cathay4t:
V2 patch uploaded.

Changes:
 * Remove `goto` statement.
 * Remove requirement of `multipath` tool. Always return `TRUE` for `BD_MPATH_TECH_BASE`. The libdmmp only require a running `multipathd`.
 * Fix issue found by vpodzime

Patch was tested by https://gist.github.com/cathay4t/7788a8dcaa3ba59bea1a862702cf07cf

vpodzime:
This PR is abandoned.

PR #305, opened 2017-11-23
Move the fs.h file to its original place

vpodzime:
Otherwise builds using it break.


vpodzime:
Merging this without review, it can hardly make things worse for projects using *libblockdev*.

PR #306, opened 2017-11-27
Use only sgdisk to set flags on GPT

vojtechtrefny:
libparted has some "custom flags" on GPT and some these "flags"
are actually not flags but GUIDs and unsetting these "flags"
chages previously set GUID.

### DIFF
--- a/src/lib/plugin_apis/part.api
+++ b/src/lib/plugin_apis/part.api
@@ -27,6 +27,20 @@ typedef enum {
     BD_PART_DISK_FLAG_GPT_PMBR_BOOT = 1
 } BDPartDiskFlag;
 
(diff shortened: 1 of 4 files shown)


src/lib/plugin_apis/part.api
```diff
@@ -27,6 +27,20 @@ typedef enum {
     BD_PART_DISK_FLAG_GPT_PMBR_BOOT = 1
 } BDPartDiskFlag;
```
vpodzime: ...some *of* these "flags"... (in the commit message)

vojtechtrefny: Fixed

vojtechtrefny:
Test added

PR #317, opened 2018-01-05
Progress for e2fsck

jtulak:
These two commits add a progress reporting capabilities for check/repair of ext2/3/4 fs. It utilizes the existing infrastructure including `bd_utils_init_prog_reporting()`. An example of usage is here: https://github.com/jtulak/fd-issue, but it is the same as for any other of the few operations that currently support progress report.

### DIFF
--- a/src/plugins/fs/ext.c
+++ b/src/plugins/fs/ext.c
@@ -314,6 +314,7 @@ static gboolean ext_check (const gchar *device, const BDExtraArg **extra, GError
     /* Force checking even if the file system seems clean. AND
      * Open the filesystem read-only, and assume an answer of no to all
      * questions. */
+    const gchar *args_progress[7] = {"e2fsck", "-f", "-n", "-C", "1", device, NULL};
     const gchar *args[5] = {"e2fsck", "-f", "-n", device, NULL};
     gint status = 0;
     gboolean ret = FALSE;
@@ -321,7 +322,12 @@ static gboolean ext_check (const gchar *device, const BDExtraArg **extra, GError
     if (!check_deps (&avail_deps, DEPS_E2FSCK_MASK, deps, DEPS_LAST, &deps_check_lock, error))
         return FALSE;
 
-    ret = bd_utils_exec_and_report_status_error (args, extra, &status, error);
+    if (bd_utils_prog_reporting_initialized ()) {
+        ret = bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
+    } else {
+        ret = bd_utils_exec_and_report_status_error (args, extra, &status, error);
+    }
+
     if (!ret && (status == 4)) {
         /* no error should be reported for exit code 4 - File system errors left uncorrected */
         g_clear_error (error);
@@ -378,12 +384,18 @@ static gboolean ext_repair (const gchar *device, gboolean unsafe, const BDExtraA
     /* Force checking even if the file system seems clean. AND
      *     Automatically repair what can be safely repaired. OR
      *     Assume an answer of `yes' to all questions. */
+    const gchar *args_progress[7] = {"e2fsck", "-f", "-n", "-C", "1", device, NULL};
     const gchar *args[5] = {"e2fsck", "-f", unsafe ? "-y" : "-p", device, NULL};
+    gint status = 0;
 
     if (!check_deps (&avail_deps, DEPS_E2FSCK_MASK, deps, DEPS_LAST, &deps_check_lock, error))
         return FALSE;
 
-    return bd_utils_exec_and_report_error (args, extra, error);
+    if (bd_utils_prog_reporting_initialized ()) {
+        return bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
+    } else {
+        return bd_utils_exec_and_report_status_error (args, extra, &status, error);
+    }
 }
 
 /**
--- a/src/plugins/fs/ext.c
+++ b/src/plugins/fs/ext.c
@@ -59,6 +59,76 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
 
 #define UNUSED __attribute__((unused))
 
+static gint8 compute_percents (guint8 pass_cur, guint8 pass_total, gint val_cur, gint val_total) {
+    gint perc;
+    gint one_pass;
+    /* first get a percentage in the current pass/stage */
+    perc = (val_cur * 100) / val_total;
+
+    /* now map it to the total progress, splitting the stages equally */
+    one_pass = 100 / pass_total;
+    perc = ((pass_cur - 1) * one_pass) + (perc / pass_total);
+
+    return perc;
+}
+
+/**
+ * filter_line_fsck: (skip)
+ * Filter one line - decide what to do with it.
+ *
+ * Returns: Zero or positive number as a percentage, -1 if not a percentage, -2 on an error
+ */
+static gint8 filter_line_fsck (const gchar * line, guint8 total_stages, GError **error) {
+    static GRegex *output_regex = NULL;
+    GMatchInfo *match_info;
+    gint8 perc = -1;
+
+    if (output_regex == NULL) {
+        /* Compile regular expression that matches to e2fsck progress output */
+        output_regex = g_regex_new ("^([0-9][0-9]*) ([0-9][0-9]*) ([0-9][0-9]*) (/.*)", 0, 0, error);
+        if (output_regex == NULL) {
+            return -2;
+        }
+    }
+
+    /* Execute regular expression */
+    if (g_regex_match (output_regex, line, 0, &match_info)) {
+        guint8 stage;
+        gint64 val_cur;
+        gint64 val_total;
+
+        /* The output_regex ensures we have a number in these matches, so we can skip
+         * tests for conversion errors.
+         */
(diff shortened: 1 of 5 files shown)


vpodzime:
@jtulak would you mind submitting a PR for https://github.com/storaged-project/api-examples with that example? I think it could be useful for other people too. Thanks!

src/plugins/fs/ext.c
```diff
@@ -19,6 +19,7 @@
 
 #include <blockdev/utils.h>
 #include <check_deps.h>
+#include <regex.h>
```
vpodzime: Please use the [regular expressions implementation from GLib](https://developer.gnome.org/glib/2.54/glib-Perl-compatible-regular-expressions.html)

src/plugins/fs/ext.c
```diff
@@ -59,6 +60,81 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
 
 #define UNUSED __attribute__((unused))
 
+static gint8 compute_percents(guint8 pass_cur, guint8 pass_total, int val_cur, int val_total)
+{
+    int perc;
+    int one_pass;
```
vpodzime: Use GLib types here. And also in the parameters.

src/plugins/fs/ext.c
```diff
@@ -59,6 +60,81 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+        return -2;
+    }
+    return perc;
+}
+
+static gboolean extract_e2fsck_progress (const gchar *line, guint8 *completion) {
+    /* A magic number 5, e2fsck has 5 stages, but this can't be read from the output in advance. */
+    *completion = filter_line_fsck(line, 5, (GError **)NULL);
+    if (*completion < 0)
+        return 0;
```
vpodzime: This should be ``FALSE``

src/plugins/fs/ext.c
```diff
@@ -59,6 +60,81 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+    return TRUE;
+}
+
+
+
```
vpodzime: Two blank lines should be enough here.

jtulak:
I updated the commits, so the issues should be fixed now. And I'm going to send the PR to the api examples too. :)

I did some other changes to this PR: I added one other commit and updated the last commit to make it easier to use (before, the user had to set -C1 argument for e2fsck manually, because ext.c plugin file didn't knew if we are reporting progress or not. I added a detection for this...

vojtechtrefny:
Jenkins, ok to test.

src/utils/exec.h
```diff
@@ -60,6 +60,7 @@ gint bd_utils_version_cmp (const gchar *ver_string1, const gchar *ver_string2, G
 gboolean bd_utils_check_util_version (const gchar *util, const gchar *version, const gchar *version_arg, const gchar *version_regexp, GError **error);
 
 gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **error);
+gboolean bd_utils_prog_reporting_initialized();
```
vojtechtrefny: Please add space between the function name and parentheses (and in the definition too). Thanks

jtulak:
Ok. Into which test I should add it, `utils_test.py`? I can't find progress reporting tested anywhere in `tests` directory: `grep -ri "prog_report" tests/`

vojtechtrefny:
I think we don't have a test for progress reporting yet. The best place is probably `fs_test.py` -- you can add a new test case similar to `ExtTestCheck` just with the progress reporting enabled.

jtulak:
I pushed a new version with the space in the function header and tests. At the end, I made two of them: one to see if the progress reporting initialization works in Utils, and one in fs to see if ext4 check reports some progress.

src/utils/exec.c
```diff
@@ -727,6 +727,17 @@ gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **e
...
 
+/**
+ * bd_utils_prog_reporting_initialized:
+ *
+ * Returns: TRUE if progress reporting has been initialized.
+ */
+gboolean bd_utils_prog_reporting_initialized () {
+    if (prog_func)
+        return TRUE;
+    return FALSE;
```
vpodzime: This can be reduced to ``return prog_func != NULL;``

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,75 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
 
 #define UNUSED __attribute__((unused))
 
+static gint8 compute_percents(guint8 pass_cur, guint8 pass_total, gint val_cur, gint val_total)
+{
```
vpodzime: Put the brace to the end of the previous line, please.

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,75 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+
+/**
+ * filter_line_fsck: (skip)
+ * Filter one line - decide what to do with it.
+ *
+ * Returns: Zero or positive number as a percentage, -1 if not a percentage, -2 on an error
+ */
+static gint8 filter_line_fsck(const gchar * line, guint8 total_stages, GError **error) {
+    static GRegex *regex;
+    static int reginit = 0;
```
vpodzime: Please rename ``regex`` to something that suggests it's specific for this function. Also, either change ``reginit`` to ``gboolean`` or get rid of it altogether (use ``regex != NULL``).

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,75 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+            return -2;
+        }
+        reginit = 1;
+    }
+
+    /* Execute regular expression */
+    if (g_regex_match (regex, line, 0, &match_info)) {
+        guint8 stage;
+        long val_cur;
+        long val_total;
```
vpodzime: ``g_ascii_strtoll()`` returns a ``gint64``. And I think ``g_ascii_strtoull()`` would be better here, right?

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,75 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+    }
+    g_match_info_free (match_info);
+    return perc;
+}
+
+static gboolean extract_e2fsck_progress (const gchar *line, guint8 *completion) {
+    /* A magic number 5, e2fsck has 5 stages, but this can't be read from the output in advance. */
+    gint8 perc;
+
+    perc = filter_line_fsck(line, 5, (GError **)NULL);
```
vpodzime: Please pass a real ``GError**`` here and use it above to report the error from ``g_regex_new()``

jtulak:
UPDATE sections removed and the other issues are fixed as well.

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,76 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+        if (output_regex == NULL) {
+            return -2;
+        }
+    }
+
+    /* Execute regular expression */
+    if (g_regex_match (output_regex, line, 0, &match_info)) {
+        guint8 stage;
+        long val_cur;
+        long val_total;
```
vpodzime: These two should be ``gint64``

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,76 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
 
 #define UNUSED __attribute__((unused))
 
+static gint8 compute_percents(guint8 pass_cur, guint8 pass_total, gint val_cur, gint val_total) {
```
vpodzime: Please add a space between function names an parentheses. In definitions as well as where the functions are called.

jtulak:
Hmmm, I'm not sure what is wrong with the last changes, why some Jenkins tests failed. Could the RPM build just get stuck?

vojtechtrefny:
> Hmmm, I'm not sure what is wrong with the last changes, why some Jenkins tests failed. Could the RPM build just get stuck?

I can ignore it, tests passed but Jenkins failed to copy the results to fedorapeople.

src/plugins/fs/ext.c
```diff
@@ -59,6 +59,76 @@ static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
...
+        guint8 stage;
+        gint64 val_cur;
+        gint64 val_total;
+
+        /* The output_regex ensures we have a number in these matches, so we can skip
+         * tests for conversion errors.
+         */
+        stage = (guint8) g_ascii_strtoull(g_match_info_fetch (match_info, 1), (char **)NULL, 10);
+        val_cur = g_ascii_strtoll(g_match_info_fetch (match_info, 2), (char **)NULL, 10);
+        val_total = g_ascii_strtoll(g_match_info_fetch (match_info, 3), (char **)NULL, 10);
```
vpodzime: Still some missing spaces between function names and ``(``s in the above and also in a couple of places below. @jtulak, please fix this so that the PR can be merged.

jtulak: OK, I tried to find any other case with a `[a-zA-Z0-9_](` regex in my files and it should be done.

vpodzime:
Thanks!

PR #320, opened 2018-02-10
Add support for VeraCrypt volumes and extend support for TCRYPT volumes

segfault3:
This adds support for:
* VeraCrypt volumes
* TCRYPT keyfiles
* TCRYPT hidden containers
* TCRYPT system volumes
* VeraCrypt PIM


segfault3:
The corresponding udisks pull request: https://github.com/storaged-project/udisks/pull/495.

vojtechtrefny:
This is unfortunately an API change -- if you need the new options for `bd_crypto_tc_open`, you'll need to create a new function -- similar to `bd_crypto_luks_open` and `bd_crypto_luks_open_blob`.

segfault3:
I see. I moved the new code into a new function in 074814d.

vojtechtrefny:
Thank you. Maybe `bd_crypto_tc_open2` is not the best name for the new function, but I don't have better idea, so I guess it's ok.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
The `veracrypt_pim` parameter is available only in cryptsetup 2, please add some checks/ifdefs for this parameter so it works on older version of cryptsetup too.

germangoergs:
How can I merge your changes on my system?

src/lib/plugin_apis/crypto.api
```diff
@@ -309,6 +309,27 @@ gboolean bd_crypto_luks_resize (const gchar *luks_device, guint64 size, GError *
  */
 gboolean bd_crypto_tc_open (const gchar *device, const gchar *name, const guint8* pass_data, gsize data_len, gboolean read_only, GError **error);
 
+/**
+ * bd_crypto_tc_open2:
```
vpodzime: I suggest ``bd_crypto_tc_open_full`` as a better name for this function.

src/plugins/crypto.c
```diff
@@ -959,7 +959,73 @@ gboolean bd_crypto_luks_resize (const gchar *luks_device, guint64 size, GError *
...
  * @data_len: length of the @pass_data buffer
  * @read_only: whether to open as read-only or not (meaning read-write)
- * @keyfiles: (array length=keyfiles_count) paths to the keyfiles for the TrueCrypt/VeraCrypt volume
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the @device was successfully opened or not
+ *
+ * Tech category: %BD_CRYPTO_TECH_TRUECRYPT-%BD_CRYPTO_TECH_MODE_OPEN_CLOSE
+ */
+gboolean bd_crypto_tc_open (const gchar *device, const gchar *name, const guint8* pass_data, gsize data_len, gboolean read_only, GError **error) {
```
vpodzime: I think the old function should just call the new more universal one.

src/plugins/crypto.c
```diff
@@ -959,14 +959,16 @@ gboolean bd_crypto_luks_resize (const gchar *luks_device, guint64 size, GError *
  * @pass_data: (array length=data_len): a passphrase for the TrueCrypt/VeraCrypt volume (may contain arbitrary binary data)
  * @data_len: length of the @pass_data buffer
  * @read_only: whether to open as read-only or not (meaning read-write)
+ * @keyfiles: (array length=keyfiles_count) paths to the keyfiles for the TrueCrypt/VeraCrypt volume
+ * @keyfiles_count: length of the @keyfiles array
```
vpodzime: We do zero/NULL terminated arrays where possible.

src/lib/plugin_apis/crypto.api
segfault3: Done in 94aca5c

src/plugins/crypto.c
segfault3: Makes sense. Done in 12841cf.

src/plugins/crypto.c
segfault3: Fixed in 8f0f434.

segfault3:
> The veracrypt_pim parameter is available only in cryptsetup 2, please add some checks/ifdefs for this parameter so it works on older version of cryptsetup too.

Done in 08a09b7.

segfault3:
> First of all -- THANKS a lot for these changes! 

I'm glad it's appreciated :)

> There are a couple small things that can be made better, but other than that the changes are nice.

I think I addressed all of them now.

> It would be nice if these changes didn't break git bisect. IOW, if the API wasn't broken even in the middle of the changes. @segfault3 please try to do that. If the rebase gets ridiculously complicated, feel free to give it up, though. :)

I rewrote git history and think that now none of the commit breaks the API. I squashed some commits to make the rebasing easier.

src/lib/plugin_apis/crypto.api
segfault3: I squashed commit 94aca5c into 9a7f054 during the rebasing.

src/plugins/crypto.c
segfault3: I also squahed 12841cf into 9a7f054 during the rebasing.

src/plugins/crypto.c
segfault3: That is now done in commit 69a3223 after the rebasing.

vojtechtrefny:
Jenkins, ok to test.

src/plugins/crypto.h
```diff
@@ -6,6 +6,10 @@
 
 #define BD_CRYPTO_LUKS_METADATA_SIZE (2 MiB)
 
+#define BD_CRYPTO_CHI_SQUARE_LOWER_LIMIT 136
+#define BD_CRYPTO_CHI_SQUARE_UPPER_LIMIT 426
+#define BD_CRYPTO_CHI_SQUARE_BYTES_TO_CHECK 512
```
vojtechtrefny: This looks like something that might be better as parameters for `bd_crypto_device_seems_encrypted`. I don't know if there is a change that these will change in feature, but it might be better to not have to recompile libblockdev just because you want to change these limits.

segfault3: Mmh, I don't really see the use case for changing these values - especially not in the higher level applications which will use this function. Why do you think they should be changeable without recompiling?

vojtechtrefny: I don't know, but the limits just look like something that users of libblockdev might want to change eventually. But if you think this isn't something that makes sense to change from e.g. udisks, we can keep it this way.

segfault3: I don't see why these values should be changed from udisks. We explained at https://tails.boum.org/blueprint/veracrypt/#detection why, with these values, the chi-squared test should only produce false negatives with negligible probability (1 in 10 billion). And devices with (non-encrypted) filesystem headers will never pass the test. The only reason I can think of to change these values is to handle strange filesystems, which don't have their header at least partly in the first 512 Bytes (I don't know of any such filesystem), but then I think we should handle this in libblockdev and not udisks or similar.

vojtechtrefny: Ok, you convinced me, we can keep it this way.

src/lib/plugin_apis/crypto.api
```diff
@@ -294,6 +294,31 @@ gboolean bd_crypto_luks_change_key_blob (const gchar *device, const guint8 *pass
...
+ * and 426 as candidates for being encrypted.
+ * For the reasoning, see: https://tails.boum.org/blueprint/veracrypt/
+ *
+ * Returns: %TRUE if the given @device seems to be encrypted or %FALSE if not or
+ * failed to determine (the @error) is populated with the error in such
+ * cases)
+ *
+ * Tech category: %BD_CRYPTO_TECH_TRUECRYPT-%BD_CRYPTO_TECH_MODE_QUERY
+ */
+gboolean bd_crypto_device_seems_encrypted (const gchar *device, GError **error);
```
vojtechtrefny: Please add the new functions to `docs/libblockdev-sections.txt` too.

segfault3: Done in b8d8bd7.

segfault3:
I fixed a URL in 2d15e2d.

PR #324, opened 2018-02-22
Fix build against musl libc

karlguy:
This patch fixes POSIX compliance by changing all references to "sys/fcntl.h"
to "fcntl.h", and changes the use of the "__PRI64_PREFIX" to the
equivalent "PRIx64".

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -25,6 +25,7 @@
 #include <inttypes.h>
 #include <unistd.h>
 #include <sys/file.h>
+#include <fcntl.h>
 #include <sys/ioctl.h>
 #include <linux/fs.h>
 #include <blockdev/utils.h>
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/plugins/part.c
```diff
@@ -25,6 +25,7 @@
 #include <inttypes.h>
 #include <unistd.h>
 #include <sys/file.h>
+#include <fcntl.h>
```
vojtechtrefny: Why is this new include needed here?

karlguy: That is required.
```
part.c:232:14: error: implicit declaration of function 'open' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
    dev_fd = open (disk->dev->path, O_RDONLY|O_CLOEXEC);
             ^
part.c:232:37: error: use of undeclared identifier 'O_RDONLY'
    dev_fd = open (disk->dev->path, O_RDONLY|O_CLOEXEC);
                                    ^
part.c:232:46: error: use of undeclared identifier 'O_CLOEXEC'
    dev_fd = open (disk->dev->path, O_RDONLY|O_CLOEXEC);
                                             ^
```

vojtechtrefny: Interesting, this doesn't happen for me. But you are right, `open` is defined in `fcntl.h`. Probably something else includes it on Linux.

PR #325, opened 2018-02-23
Fix build with clang

karlguy:
Fixed multiple things such as improper initializatons of structs,
and type conversion issues.

### DIFF
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -35,6 +35,8 @@
 #define CRYPT_LUKS NULL
 #endif
 
+#define ZERO_INIT {}
+
 /**
  * SECTION: crypto
  * @short_description: plugin for operations with encrypted devices
@@ -970,7 +972,7 @@ gboolean bd_crypto_tc_open (const gchar *device, const gchar *name, const guint8
     gint ret = 0;
     guint64 progress_id = 0;
     gchar *msg = NULL;
-    struct crypt_params_tcrypt params = {0};
+    struct crypt_params_tcrypt params = ZERO_INIT;
 
     msg = g_strdup_printf ("Started opening '%s' TrueCrypt/VeraCrypt device", device);
     progress_id = bd_utils_report_started (msg);
@@ -1090,7 +1092,7 @@ static gchar *replace_char (gchar *str, gchar orig, gchar new) {
     return str;
 }
 
-static gboolean write_escrow_data_file (struct libvk_volume *volume, struct libvk_ui *ui, enum libvk_packet_format format, const gchar *out_path,
+static gboolean write_escrow_data_file (struct libvk_volume *volume, struct libvk_ui *ui, enum libvk_secret secret_type, const gchar *out_path,
                                         CERTCertificate *cert, GError **error) {
     gpointer packet_data = NULL;
     gsize packet_data_size = 0;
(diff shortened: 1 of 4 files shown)


src/plugins/crypto.c
```diff
@@ -970,7 +970,7 @@ gboolean bd_crypto_tc_open (const gchar *device, const gchar *name, const guint8
     gint ret = 0;
     guint64 progress_id = 0;
     gchar *msg = NULL;
-    struct crypt_params_tcrypt params = {0};
+    struct crypt_params_tcrypt params = {NULL,0,NULL,0,NULL,NULL,NULL,0,0,0};
```
vpodzime: I will say this right away here: *I don't like this change.* The compiler should be clever enough to figure this out. Maybe *clang* needs to be told to not produce warnings about this?

karlguy: I completely agree with you on this, and there should be a better way to do that. It's possible upstream may have thought of this already and provided a compiler flag or macro for this. I'll investigate.

vpodzime: > I'll investigate.

Thanks!

karlguy:
@vpodzime 
Ok, so after a bit of digging I discovered a few things. First of all, POSIX does not provide a clear, standardized way in C to initialize a struct to default values, which causes this problem in the first place. The best we can get then is what POSIX defines as standard in C++ for struct initialization (Option 3).

A relevant link: http://clang-developers.42468.n3.nabble.com/Wmissing-field-initializers-td3761260.html

**Option 1**

This is currently the only struct initialization supported by POSIX C99, but is bad for reasons mentioned.
```
struct crypt_params_tcrypt params = {NULL,0,NULL,0,NULL,NULL,NULL,0,0,0};
```

**Option 2**

This is what you guys currently do, but clang dislikes this because it is a GNU extension, and because the default initalized values do not necessarily have to have any connection with the value 0.
```
struct crypt_params_tcrypt params = {0};
```
**Option 3**

Instead, clang prefers this, which is POSIX in C++, and also supported by GCC. I recommend this option, and will be updating the PR with this option.
```
struct crypt_params_tcrypt params = {};
```

src/plugins/crypto.c
```diff
@@ -970,7 +970,7 @@ gboolean bd_crypto_tc_open (const gchar *device, const gchar *name, const guint8
     gint ret = 0;
     guint64 progress_id = 0;
     gchar *msg = NULL;
-    struct crypt_params_tcrypt params = {0};
+    struct crypt_params_tcrypt params = {};
```
vpodzime: Could you please add and use something like ``#define ZERO_INIT {}``? That way we can easily change it in the future and also the intention would be clear.

vpodzime:
Jenkins, this is ok to test.

src/plugins/crypto.c
karlguy: Done

vpodzime:
Hmm, looks like the macro is causing issues, unfortunately. :(

karlguy:
Hmm, it worked when I tried it with GCC on my machine.

vojtechtrefny:
```
exec.c: In function 'bd_utils_exec_and_report_progress':
exec.c:359:5: error: missing initializer for field 'fd' of 'GPollFD' [-Werror=missing-field-initializers]
```

It happens only on CentOS 7 -- it has old gcc 4.8.5, so we'll probably need some ifdef for this.

vpodzime:
@karlguy which options are you using when building with *clang*? I just tried it on my system and it's not complaining about the  ``{0}`` initializers at all.

karlguy:
I added no options at all, and all I did was ./configure and make.
```
exec.c:357:25: error: missing field 'events' initializer
      [-Werror,-Wmissing-field-initializers]
    GPollFD fds[2] = {{0}, {0}};
                        ^
exec.c:357:30: error: missing field 'events' initializer
      [-Werror,-Wmissing-field-initializers]
    GPollFD fds[2] = {{0}, {0}};
```
It's because by default, clang calls -Werror for missing field initializers.

karlguy:
@vpodzime The ifdefs are actually pretty ugly if we test for a gcc < 5.
```
#if __GNUC__ < 5
#ifndef __clang__
#define ZERO_INIT {0}
#else
#define ZERO_INIT {}
#endif
#else
#define ZERO_INIT {}
#endif
```
I decided to just use this to keep it less messy
```
#ifdef __clang__
#define ZERO_INIT {}
#else
#define ZERO_INIT {0}
#endif
```

vpodzime:
Thanks @karlguy!

PR #326, opened 2018-02-26
s390: don't hardcode paths, search PATH

flokli:
This removes hardcoded binary paths from src/plugins/s390.c.

### DIFF
--- a/src/plugins/s390.c
+++ b/src/plugins/s390.c
@@ -154,7 +154,7 @@ gboolean bd_s390_is_tech_avail (BDS390Tech tech, guint64 mode, GError **error) {
  */
 gboolean bd_s390_dasd_format (const gchar *dasd, const BDExtraArg **extra, GError **error) {
     gboolean rc = FALSE;
-    const gchar *argv[8] = {"/sbin/dasdfmt", "-y", "-d", "cdl", "-b", "4096", NULL, NULL};
+    const gchar *argv[8] = {"dasdfmt", "-y", "-d", "cdl", "-b", "4096", NULL, NULL};
 
     if (!check_deps (&avail_deps, DEPS_DASDFMT_MASK, deps, DEPS_LAST, &deps_check_lock, error))
         return FALSE;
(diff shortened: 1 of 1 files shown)


jtojnar:
Actually, would it be possible to make the paths adjustable in `./configure`? That would allow us to use the library without setting the path in each of its consumers.

flokli:
I'd agree it would be useful to keep track of all the used binaries, and warn (or error) if those binaries were not found at configure time.

We could use `AC_PATH_PROG` and `AC_DEFINE_UNQUOTED` to do that, similar to how it's done in [libvirt](https://github.com/libvirt/libvirt/blob/master/configure.ac#L874). Not sure if we want to just warn (and hope its in $PATH at runtime) or fail configure phase if a binary wasn't found (and the feature is enabled). @vpodzime any opinion on that?

But in general, this would be out of the scope of this PR, but a separate one.

vpodzime:
> Not sure if we want to just warn (and hope its in $PATH at runtime) or fail configure phase if a binary wasn't found (and the feature is enabled). @vpodzime any opinion on that?

I don't see any reason why the binaries should have to be required at configure phase. The only case would be if somebody wanted to run the tests, but that's not the usual case for this library when just building it.

And to be honest, I think searching for the binaries in ``$PATH`` is the best way to go. Why should the library care where something is installed? What if the user has their own version of something installed somewhere else than in the standard location? I believe that's what the ``$PATH`` variable was created for and I don't see a reason to not use it. Last but not least, this library is used during OS installations which may potentially run in very non-standard environments.

flokli:
@vpodzime using `AC_{CHECK,PATH}_PROG` at configure time would help in listing required and discovering missing binaries, instead of simply failing at runtime if it's is missing.

This can be used to only warn at configure phase if the binary is missing, and still rely on `$PATH` lookups at runtime.

vpodzime:
> @vpodzime using AC_{CHECK,PATH}_PROG at configure time would help in listing required and discovering missing binaries, instead of simply failing at runtime if it's is missing.

Failing only at runtime if a binary is missing is one of the goals here. That way other things can still be used.

vpodzime:
> This can be used to only warn at configure phase if the binary is missing, and still rely on $PATH lookups at runtime.

That sounds like a plan to me.

flokli:
Great! Will file a followup PR for that. Anything missing on this one here?

vpodzime:
> Great! Will file a followup PR for that. Anything missing on this one here?

No, I'm just going to merge it. Thanks!

PR #330, opened 2018-03-21
lvm-dbus: Check returned job object for error

vojtechtrefny:
It is possible that calling a function won't return an error
immediately but will return a job object with 'GetError' property
set so we need to check it too.

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -605,6 +605,8 @@ static void call_lvm_method_sync (const gchar *obj, const gchar *intf, const gch
     gdouble progress = 0.0;
     gchar *log_msg = NULL;
     gboolean completed = FALSE;
+    gint64 error_code = 0;
+    gchar *error_msg = NULL;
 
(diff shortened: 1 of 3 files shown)


src/plugins/lvm-dbus.c
```diff
@@ -706,18 +709,41 @@ static void call_lvm_method_sync (const gchar *obj, const gchar *intf, const gch
...
+                    log_task_status (log_task_id, "No result");
+            }
             bd_utils_report_finished (prog_id, "Completed");
             g_free (obj_path);
 
             /* remove the job object and clean after ourselves */
             ret = g_dbus_connection_call_sync (bus, LVM_BUS_NAME, task_path, JOB_INTF, "Remove", NULL,
-                                               NULL, G_DBUS_CALL_FLAGS_NONE, METHOD_CALL_TIMEOUT, NULL, error);
+                                               NULL, G_DBUS_CALL_FLAGS_NONE, METHOD_CALL_TIMEOUT, NULL,
+                                               &loc_error);
```
vpodzime: ``NULL`` can be passed here if we don't care about the error.

vojtechtrefny: Fixed, thanks.

Issue #333, opened 2018-03-25
libblockdev fails to build, 'libvolume_key.h not available'


TimB87:
Hey, 
I'm trying to build libblockdev 2.16 (or git) with crypto support, I have installed volume_key 0.3.9 in the same prefix (=/usr), but configure aborts with libvolume_key.h not available.
Can somebody help me out here?

Log:
+ cd libblockdev-2.16
+ ./configure --prefix=/usr --without-lvm_dbus --without-dm --without-lvm --without-bcache --without-gtk-doc
[..]
*** Libblockdev encountered the following issues during configuration:

libvolume_key.h not available

*** Libblockdev will not successfully build without these missing dependencies
=======> ERROR: Building '/home/pkgmk/packages/libblockdev#2.16-1.pkg.tar.gz' failed.

# ls -la /usr/include/volume_key/libvolume_key.h 
-rw-r--r-- 1 root root 14890 Mär 25 13:19 /usr/include/volume_key/libvolume_key.h
# ls -la /usr/lib/libvolume_key.*                                                 
-rwxr-xr-x 1 root root   1231 Mär 25 13:19 /usr/lib/libvolume_key.la
lrwxrwxrwx 1 root root     22 Mär 25 13:19 /usr/lib/libvolume_key.so -> libvolume_key.so.1.2.2
lrwxrwxrwx 1 root root     22 Mär 25 13:19 /usr/lib/libvolume_key.so.1 -> libvolume_key.so.1.2.2
-rwxr-xr-x 1 root root 101392 Mär 25 13:19 /usr/lib/libvolume_key.so.1.2.2

vpodzime:
Doesn't the output from ``./configure`` tell something more? I mean, in the part where it checks for the library/headers.

TimB87:
Hey,
this is the complete output: https://pastebin.com/2dsde4kw

vpodzime:
Could you please test if #334 fixes the issue for you, @TimB87? Thanks!

TimB87:
I am not that familiar with git yet, so I created patch file with your change and patched it against 2.16, no change so far.
https://pastebin.com/JpA39rKw
I'll try with the git version next.

TimB87:
Seems to work!
=======> New files found:
NEW       -rw-r--r--      root/root       usr/include/blockdev/crypto.h
NEW       -rwxr-xr-x      root/root       usr/lib/libbd_crypto.la
[...]

PR #334, opened 2018-03-26
Use '=' instead of '==' to compare using 'test'

vpodzime:
'=' is correct, '==' may work with some versions of 'test' (or
shells that don't actually use the utility).

Fixes: #333


vpodzime:
Small change, all tests passed, issue reported to be addressed by this change. Merging.

PR #338, opened 2018-04-11
New function for luks metadata size

japokorn:
- added new crypto plugin API function that returns offset where the
disk data starts (~ luks metadata size)


src/lib/plugin_apis/crypto.api
```diff
@@ -89,6 +89,19 @@ gboolean bd_crypto_device_is_luks (const gchar *device, GError **error);
  */
 gchar* bd_crypto_luks_uuid (const gchar *device, GError **error);
 
+/**
+ * bd_crypto_luks_metadata_size:
```
vpodzime: I would prefer ``bd_crypto_luks_get_metadata_size``, it's a function, after all. I know there's a similar ``bd_crypto_luks_status`` right below, but that's a remnant of the original *blivet.devicelibs* code legacy.

src/plugins/crypto.c
```diff
@@ -243,6 +243,43 @@ gchar* bd_crypto_luks_uuid (const gchar *device, GError **error) {
     return ret;
 }
 
+/**
+ * bd_crypto_luks_metadata_size:
+ * @device: the queried device
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: luks device metadata size of the @device
+ * or %NULL if failed to determine (@error) is populated
```
vpodzime: ``NULL`` is not of type ``guint64``

src/plugins/crypto.c
```diff
@@ -243,6 +243,43 @@ gchar* bd_crypto_luks_uuid (const gchar *device, GError **error) {
...
+
+    ret_num = crypt_load (cd, CRYPT_LUKS, NULL);
+    if (ret_num != 0) {
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_DEVICE,
+                     "Failed to load device: %s", strerror_l(-ret_num, c_locale));
+        crypt_free (cd);
+        return 0;
+    }
+
+    ret = 512 * crypt_get_data_offset (cd);
```
vpodzime: I think we have a macro/constant for 512. If we don't, there should be one. Magic numbers are bad.

src/lib/plugin_apis/crypto.api
japokorn: Function renamed.

src/plugins/crypto.c
japokorn: Fixed.

src/plugins/crypto.c
japokorn: Fixed.

src/plugins/crypto.c
```diff
@@ -243,13 +245,50 @@ gchar* bd_crypto_luks_uuid (const gchar *device, GError **error) {
...
+}
+
 /**
  * bd_crypto_luks_status:
  * @luks_device: the queried LUKS device
  * @error: (out): place to store error (if any)
  *
  * Returns: (transfer none): one of "invalid", "inactive", "active" or "busy" or
- * %NULL if failed to determine (@error is populated with the error in
+ * 0 if failed to determine (@error is populated with the error in
```
vpodzime: This actually returns NULL in case of failure.

tests/crypto_test.py
```diff
@@ -388,6 +389,49 @@ def test_luks_get_uuid(self):
...
+
+        ret, out, err = run_command("cryptsetup luksDump %s" % self.loop_dev)
+        if ret != 0:
+            self.fail("Failed to get LUKS 2 header from %s:\n%s %s" % (self.loop_dev, out, err))
+
+        m = re.search(r"offset:\s*([0-9]+)\s*\[bytes\]", out)
+        if m is None:
+            self.fail("Failed to get LUKS 2 offset information from %s:\n%s %s" % (self.loop_dev, out, err))
+        offset = m.group(1)
+        self.assertEquals(str(meta_size), offset, "LUKS 2 metadata sizes differ")
```
vpodzime: I think it would be better to compare integers here too (just like in the other test below).

japokorn: Fixed.

src/plugins/crypto.c
japokorn: Fixed.

japokorn:
Documentation updated.

PR #347, opened 2018-05-03
Always check for error when (un)mounting

vojtechtrefny:
Both `mnt_context_mount` and `mnt_context_umount` can return zero
even if the (un)mount failed. This happens when using the (un)mount
helper program (e.g. mount.ntfs), in this case we need to get the
return code using `mnt_context_get_helper_status`.

### DIFF
--- a/src/plugins/fs/mount.c
+++ b/src/plugins/fs/mount.c
@@ -49,12 +49,15 @@ typedef gboolean (*MountFunc) (MountArgs *args, GError **error);
 static gboolean do_mount (MountArgs *args, GError **error);
 
 #ifndef LIBMOUNT_NEW_ERR_API
-static void parse_unmount_error_old (struct libmnt_context *cxt, int rc, const gchar *spec, GError **error) {
+static gboolean get_unmount_error_old (struct libmnt_context *cxt, int rc, const gchar *spec, GError **error) {
     int syscall_errno = 0;
+    int helper_status = 0;
 
-    if (mnt_context_syscall_called (cxt)) {
+    if (mnt_context_syscall_called (cxt) == 1) {
         syscall_errno = mnt_context_get_syscall_errno (cxt);
         switch (syscall_errno) {
+            case 0:
+                return TRUE;
             case EBUSY:
                 g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
                              "Target busy.");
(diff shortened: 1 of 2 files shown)


src/plugins/fs/mount.c
```diff
@@ -72,6 +73,14 @@ static void parse_unmount_error_old (struct libmnt_context *cxt, int rc, const g
...
                 break;
         }
+    } else if (mnt_context_helper_executed (cxt) == 1) {
+        helper_status = mnt_context_get_helper_status (cxt);
+        if (helper_status != 0) {
+            g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
+                         "Unmount helper program failed: %d.", helper_status);
+            return FALSE;
+        } else
+            return TRUE;
```
vpodzime: Are you sure this is the only place where this function should return ``TRUE``?

src/plugins/fs/mount.c
```diff
@@ -259,6 +263,14 @@ static gboolean parse_mount_error_old (struct libmnt_context *cxt, int rc, Mount
...
                 break;
         }
+    } else if (mnt_context_helper_executed (cxt) == 1) {
+        helper_status = mnt_context_get_helper_status (cxt);
+        if (helper_status != 0) {
+            g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
+                         "Mount helper program failed: %d.", helper_status);
+            return FALSE;
+        } else
+            return TRUE;
```
vpodzime: Same here.

tests/fs_test.py
```diff
@@ -86,6 +86,16 @@ def _clean_up(self):
...
 
+    def setro(self, device):
+        ret, _out, _err = utils.run_command("blockdev --setro %s" % device)
+        if ret != 0:
+            self.fail("Failed to set %s read-only" % device)
+
+    def setrw(self, device):
+        ret, _out, _err = utils.run_command("blockdev --setrw %s" % device)
+        if ret != 0:
+            self.fail("Failed to set %s read-only" % device)
```
vpodzime: *read-write*

tests/fs_test.py
```diff
@@ -1125,6 +1135,67 @@ def test_mount_fstab_user(self):
...
+
+        # set the device read-only
+        self.setro(self.loop_dev)
+        self.addCleanup(self.setrw, self.loop_dev)
+
+        # standard mount (rw) should fail
+        with self.assertRaises(GLib.GError):
+            BlockDev.fs_mount(self.loop_dev, tmp, "ntfs", None)
+
+        # read-only mount shoudl work
```
vpodzime: *should*

vojtechtrefny:
Jenkins, test this please.

PR #348, opened 2018-05-08
Fix off-by-one error when counting TCRYPT keyfiles

segfault3:
When opening a TCRYPT volume, the keyfiles count is off by one.

### DIFF
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -1072,9 +1072,8 @@ gboolean bd_crypto_tc_open_full (const gchar *device, const gchar *name, const g
     g_free (msg);
 
     if (keyfiles) {
-        for (i=0; *(keyfiles + i); i++) {
-            keyfiles_count = i;
-        }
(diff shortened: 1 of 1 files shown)


src/plugins/crypto.c
```diff
@@ -1073,7 +1073,7 @@ gboolean bd_crypto_tc_open_full (const gchar *device, const gchar *name, const g
 
     if (keyfiles) {
         for (i=0; *(keyfiles + i); i++) {
-            keyfiles_count = i;
+            keyfiles_count = i + 1;
```
vpodzime: Why is ``keyfiles_count`` not assigned just once to ``i`` after the cycle?

vpodzime: Yeah, to me that would look more like counting. This looks to me like there is something missing in the loop that should use the updated value of ``keyfiles_count``.

vpodzime:
Jenkins, test this, please.

PR #352, opened 2018-05-17
Correct arguments for ext4 repair with progress

pothos:
The ext4 repair function with progress reporting
did pass "-n" which is the read-only mode.

Use same arguments as without progress reporting.


vojtechtrefny:
Jenkins, ok to test.

PR #353, opened 2018-05-17
Introduce reporting function per thread

pothos:
Having only one global reporting function
requires the client to filter for the events
of the current thread and would involve
knowledge about the task_id.

The new function bd_utils_init_prog_reporting_thread
can now be used to set up a reporting function only
for the current thread which then takes precedence
over the global report function.


vojtechtrefny:
Jenkins, ok to test.

docs/libblockdev-sections.txt
```diff
@@ -156,6 +156,7 @@ bd_utils_exec_and_report_error_no_progress
 bd_utils_exec_and_report_progress
 bd_utils_init_logging
 bd_utils_init_prog_reporting
+bd_utils_init_prog_reporting_thread
```
vpodzime: Should we also add a function to unset the progress reporting handler for a thread?

vpodzime: Yes, but I think it would be a nice convenience function.

pothos: Ok, I have added it by using static empty function which allows to detect muting in `bd_utils_prog_reporting_initialized`. Checking in `bd_utils_report_progress` is not necessary because the code will usually test if reporting is initialized and otherwise not call this function, but anyway it would only make an empty call which is not much overhead.

src/utils/exec.c
```diff
@@ -733,13 +734,54 @@ gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **e
...
+       function gets more complicated */
+
+    thread_prog_func = new_prog_func;
+
+    return TRUE;
+}
+
+static void thread_progress_muted (guint64 task_id __attribute__((unused)), BDUtilsProgStatus status __attribute__((unused)), guint8 completion __attribute__((unused)), gchar *msg __attribute__((unused))) {
+    /* This function serves as a special value for the progress reporting
+     * function to detect that nothing is done here. If the client uses an own
```
vpodzime: *...their own...*

src/utils/exec.c
```diff
@@ -733,13 +734,54 @@ gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **e
...
+}
+
 /**
  * bd_utils_prog_reporting_initialized:
  *
  * Returns: TRUE if progress reporting has been initialized.
  */
 gboolean bd_utils_prog_reporting_initialized () {
-    return prog_func != NULL;
+    return (thread_prog_func != NULL || prog_func != NULL) && thread_prog_func != thread_progress_muted;
```
vpodzime: I think explicitly muted thread progress reporting should be considered initialized.

vpodzime: > So this means that e.g. fsck would try to use the reporting function and parse the output.

Oh that's a good point, we should avoid the parsing overhead.

vpodzime: > Should muting and disabling be considered differently? Then I could implement both, once for local muting (still initialized) and once for disabling (not initialized anymore).

Yes, that's what I head in mind. I think disabling should be possible like before. However, I see a benefit in being able to mute just specific thread(s).

pothos:
Ok, thanks for the clarification, this should cover both use cases now ;)

src/utils/exec.c
```diff
@@ -733,13 +734,77 @@ gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **e
...
+    return TRUE;
+}
+
+static void thread_progress_deinitialized (guint64 task_id __attribute__((unused)), BDUtilsProgStatus status __attribute__((unused)), guint8 completion __attribute__((unused)), gchar *msg __attribute__((unused))) {
+    /* This function serves as a special value for the progress reporting
+     * function to detect that nothing is done here. If clients use their own
+     * empty function then bd_utils_prog_reporting_initialized will return TRUE
+     * but with this function here it returns FALSE.
+     */
+}
```
vpodzime: Can't we just use ``NULL`` instead of this special function?

pothos: `NULL` is the default and means that no local reporting function is used and thus the global one is called.

pothos: But local deinitialization is needed to avoid the parsing overhead. Reversing the process of setting up a reporting function is similar in global and local state: passing `NULL` to `bd_utils_init_prog_reporting_thread` as it is done with `bd_utils_init_prog_reporting`.

pothos: `NULL` for `bd_utils_init_prog_reporting_thread` resets the behavior to the global initialization which might or might not have been set up.

vpodzime: Which is exactly what ``thread_initialize()`` followed by ``thread_deinitialize()`` should do I think.

vpodzime: ``thread_deinitialize()`` should guarantee that **thread** reporting is deinitialized/not used. So I think it should just set the thread-local reporting function to ``NULL``. So the same as ``thread_initialize(NULL)``, but it provides some level of abstraction and can do different things in the future.

pothos: Ok, good, was not sure what naming is best. Then what name should the function have that deinitializes reporting for a thread? How about `bd_utils_disable_any_prog_reporting_thread`?

vpodzime: Well, that's what ``mute()`` should do, ideally also avoiding the parsing overhead, but we can probably live with it, if necessary.

pothos: Ah, then I misunderstood you in the beginning with the 'convenience function', sorry… Ok, empty callbacks can be set up by the client but a special value which disables it not, so only this will get a function here called `mute`. I'll update it ;)

vpodzime: Do you plan to do some more changes here? Because right now ``thread_deinitialize()`` is not a reverse process to ``thread_initialize()``, it's the same as ``mute()``, right?

pothos: Hi, sorry, that function was just a leftover and not used anymore. I think this is it now.

src/utils/exec.c
```diff
@@ -733,13 +734,58 @@ gboolean bd_utils_init_prog_reporting (BDUtilsProgFunc new_prog_func, GError **e
...
+gboolean bd_utils_init_prog_reporting_thread (BDUtilsProgFunc new_prog_func, GError **error __attribute__((unused))) {
+    /* XXX: the error attribute will likely be used in the future when this
+       function gets more complicated */
+
+    thread_prog_func = new_prog_func;
+
+    return TRUE;
+}
+
+static void thread_progress_deinitialized (guint64 task_id __attribute__((unused)), BDUtilsProgStatus status __attribute__((unused)), guint8 completion __attribute__((unused)), gchar *msg __attribute__((unused))) {
```
vpodzime: I think ``thread_progress_muted()`` was a better name for this function.

pothos: Ok, changed. It was just because it's used in the test of `bd_utils_prog_reporting_initialized`.

vojtechtrefny:
Jenkins, test this please.

PR #356, opened 2018-06-20
RFC - crypto: Add bd_crypto_luks_get_policies

mvollmer:
This is to support the UDisks2 API discussed in https://github.com/storaged-project/udisks/pull/539

Only the "tang" and "sss" pins are supported for now.  There needs to be explicit support for each pin since this code needs to know how to convert the JWE header stored in the luksmeta slot back to the config object that could be passed to "clevis luks bind".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,6 +193,8 @@ LIBBLOCKDEV_PKG_CHECK_MODULES([KMOD], [libkmod >= 19])
 
 AS_IF([test "x$with_crypto" != "xno"],
       [LIBBLOCKDEV_PKG_CHECK_MODULES([CRYPTSETUP], [libcryptsetup >= 1.6.7])
+      LIBBLOCKDEV_PKG_CHECK_MODULES([LUKSMETA], [luksmeta])
+      LIBBLOCKDEV_PKG_CHECK_MODULES([JSON_C], [json-c])
       AS_IF([$PKG_CONFIG --atleast-version=2.0 libcryptsetup],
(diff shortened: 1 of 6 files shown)


mvollmer:
@npmccallum, fyi.  I just wanted write this all down, although it makes a lot of sense to have this code in clevis itself.  I think it's not a good idea to spawn a shell script to do the processing, so maybe there could be a small libclevis.so that does the job of `get_clevis_config` here.

mvollmer:
It's probably not a good idea to return the array inside the JSON string, but I couldn't quickly figure out how to make Python bindings for a function that returns a array of structs...

vojtechtrefny:
Just some general notes:
- The luksmetada dependency needs to be optional. Something similar to the `volume_key` dependency. Packagers/users from various distributions are already asking us to reduce the dependecies so we shouldn't add a new mandatory one. And it looks like luksmeta is not available for some popular distributions like [SUSE ](https://software.opensuse.org/search?utf8=%E2%9C%93&q=luksmeta) or [Arch](https://www.archlinux.org/packages/?sort=&q=luksmeta).
- It looks like this is adding a lot logic that should be part of the luksmeta library, not libblockdev. The decoding of the header is really complicated and it could be really hard to maintain in libblockdev.
- It should be possible to return json directly (even in python bindings), I'll try to take a look at this.

mvollmer:
> It looks like this is adding a lot logic that should be part of the luksmeta library, not libblockdev. 

Yes, very much so.  I'll close this PR to underscore that point.

PR #372, opened 2018-07-18
vdo: Implement bd_vdo_get_statistics()

tbzatek:
This exposes statistics of the specified VDO volume. Instead of calling either
`vdostats` or `vdo info` whose output keys are human-readable strings with
no stability guarantees the statistics are actually taken from /sys/kvdo structure
which seems to be reasonably stable.

There are some keys that are computed by the vdo tools from existing values
so those are backported.

The fact that no calls to the vdo tools are involved makes this function suitable
for reuse within potential VDO support integration in the LVM module. Also since
there's no real I/O involved this function should be reasonably fast.

### DIFF
--- a/src/plugins/vdo.c
+++ b/src/plugins/vdo.c
@@ -864,3 +864,141 @@ gboolean bd_vdo_grow_physical (const gchar *name, const BDExtraArg **extra, GErr
 
     return ret;
 }
+
+
+static gboolean get_stat_val (GHashTable *stats, const gchar *key, gint64 *val) {
+    const gchar *s;
+    gchar *endptr = NULL;
+
+    s = g_hash_table_lookup (stats, key);
+    if (s == NULL)
+        return FALSE;
+
+    *val = g_ascii_strtoll (s, &endptr, 0);
+    if (endptr == NULL || *endptr != '\0')
+        return FALSE;
+
+    return TRUE;
+}
+
+static void add_write_ampl_r_stats (GHashTable *stats) {
+    gint64 bios_meta_write, bios_out_write, bios_in_write;
+
+    if (! get_stat_val (stats, "bios_meta_write", &bios_meta_write) ||
+        ! get_stat_val (stats, "bios_out_write", &bios_out_write) ||
+        ! get_stat_val (stats, "bios_in_write", &bios_in_write))
+        return;
+
+    if (bios_in_write <= 0)
+        g_hash_table_replace (stats, g_strdup ("writeAmplificationRatio"), g_strdup ("0.00"));
+    else
+        g_hash_table_replace (stats,
+                              g_strdup ("writeAmplificationRatio"),
+                              g_strdup_printf ("%.2f", (gfloat) (bios_meta_write + bios_out_write) / (gfloat) bios_in_write));
+}
+
+static void add_block_stats (GHashTable *stats) {
+    gint64 physical_blocks, block_size, data_blocks_used, overhead_blocks_used, logical_blocks_used;
+    gint64 savings;
+
+    if (! get_stat_val (stats, "physical_blocks", &physical_blocks) ||
+        ! get_stat_val (stats, "block_size", &block_size) ||
+        ! get_stat_val (stats, "data_blocks_used", &data_blocks_used) ||
+        ! get_stat_val (stats, "overhead_blocks_used", &overhead_blocks_used) ||
+        ! get_stat_val (stats, "logical_blocks_used", &logical_blocks_used))
+        return;
+
+    g_hash_table_replace (stats, g_strdup ("oneKBlocks"), g_strdup_printf ("%ld", physical_blocks * block_size / 1024));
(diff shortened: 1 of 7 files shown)


src/plugins/vdo.c
```diff
@@ -864,3 +864,141 @@ gboolean bd_vdo_grow_physical (const gchar *name, const BDExtraArg **extra, GErr
...
+
+    if (! get_stat_val (stats, "journal_entries_committed", &journal_entries_committed) ||
+        ! get_stat_val (stats, "journal_entries_started", &journal_entries_started) ||
+        ! get_stat_val (stats, "journal_entries_written", &journal_entries_written) ||
+        ! get_stat_val (stats, "journal_blocks_committed", &journal_blocks_committed) ||
+        ! get_stat_val (stats, "journal_blocks_started", &journal_blocks_started) ||
+        ! get_stat_val (stats, "journal_blocks_written", &journal_blocks_written))
+        return;
+
+    g_hash_table_replace (stats, g_strdup ("journal_entries_batching"), g_strdup_printf ("%ld", journal_entries_started - journal_entries_written));
```
vojtechtrefny: You need to use `G_GINT64_FORMAT` macro here instead of `ld`.

tests/vdo_test.py
```diff
@@ -277,6 +278,20 @@ def test_grow_physical(self):
...
+        self.addCleanup(self._remove_vdo, self.vdo_name)
+        self.assertTrue(ret)
+
+        with six.assertRaisesRegex(self, GLib.GError, "No such file or directory"):
+            stats = BlockDev.vdo_get_statistics("nonexistingxxx")
+
+        stats = BlockDev.vdo_get_statistics(self.vdo_name)
+        self.assertIsNotNone(stats)
+        self.assertGreater(len(stats), 0)
+
```
vojtechtrefny: We should check at least some of the returned data to be sure it is the same what vdo tools report.

src/plugins/vdo.c
```diff
@@ -864,3 +864,141 @@ gboolean bd_vdo_grow_physical (const gchar *name, const BDExtraArg **extra, GErr
...
+ */
+GHashTable* bd_vdo_get_statistics (const gchar *name, const BDExtraArg **extra G_GNUC_UNUSED, GError **error) {
+    GHashTable *stats;
+    GDir *dir;
+    gchar *stats_dir;
+    const gchar *direntry;
+    gchar *s;
+    gchar *val;
+
+    if (!check_deps (&avail_deps, DEPS_VDO_MASK, deps, DEPS_LAST, &deps_check_lock, error) ||
```
vojtechtrefny: The statistics are provided by the kernel module, so this function should work without the vdo cli tools, right?

vpodzime:
I'm not really a fan of returning a hash table like this. I guess the reason is that VDO is still quite unstable as far as these things are concerned?

tbzatek:
Would you be more comfortable with an array of custom struct instead of this GLib data type?

Stability of VDO tools output or this `/sys/kvdo/` structure is quite a question. I've actually tried asking the VDO team directly but haven't received any reasonable answer other than that any files in `/etc` are considered private and should not be parsed.

I'm expecting degree of instability with regards to the attributes presented by the `kvdo` module, that's the reason I wanted to allow the output some degree of flexibility. We can't rely on specific attribute presence, nor its values.

vpodzime:
I think the best way would be to add two functions:

* `get_stats()` returning the few "stable" and most important statistics

* `get_stats_full()` returning a hash table, but with the minimum set of keys in the hash table defined 
  in the documentation of the function and no keys ever being removed in newer versions.

Does that make sense?

tbzatek:
Makes sense as discussed with @vojtechtrefny, thanks for the answer. We can pick some of the most useful keys.

Just that the hashtable contents would always be kinda free-form depending on the kernel `kvdo` module. So no guarantees about presence of all known attributes. In such cases the struct returned by `get_stats()` would just carry either `NULL` or `-1` values when certain attribute is not present. The `vdo` tool seems to be working the same way, printing only available statistics.

If you really want to have stable set of keys in the hashtable, we would have to explicitly add NULL values for the well-known keys.

vojtechtrefny:
> `get_stats_full()` returning a hash table, but with the minimum set of keys in the hash table defined
in the documentation of the function and no keys ever being removed in newer versions.

The problem is that VDO can decide to remove or rename some of the stats and there is nothing we can do about it. I think we should just say the results of the "full" function are "unstable" and users of the plugin have to check if the key they want is present. It's better than creating some artificial results (like -1 or NULL) for missing stats.

vpodzime:
> The problem is that VDO can decide to remove or rename some of the stats and there is nothing we can do about it.

And one would think they themselves should care about backwards-compatibility. Alright, I guess then there's nothing better to do than have a best-effort hash table.

tbzatek:
Oh well Github is unable to track re-created branch nor update the head branch. Can't even reopen this PR manually. Seems like there's no other option to rebase existing PR than doing a force push that is generally discouraged.

So please continue in #381

Issue #373, opened 2018-07-25
python module not fully compatible with python-3.4+


Polynomial-C:
From downstream [bug report](https://bugs.gentoo.org/662066):
```
>>> from gi.repository import BlockDev
>>> dir(BlockDev.btrfs)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/site-packages/gi/overrides/BlockDev.py", line 915, in __dir__
    items = set(dir(self._mod) + locals().keys())
TypeError: can only concatenate list (not "dict_keys") to list
```

PR #374, opened 2018-07-25
BlockDev.py Convert dictionary keys to list before using them

vojtechtrefny:
"dict_keys" is a special type in Python 3, not a list-like object.

Fixes: #373

### DIFF
--- a/src/python/gi/overrides/BlockDev.py
+++ b/src/python/gi/overrides/BlockDev.py
@@ -942,7 +942,7 @@ class ErrorProxy(object):
         """Let's make TAB-TAB in ipython work!"""
 
         if self._use_local:
-            items = set(dir(self._mod) + locals().keys())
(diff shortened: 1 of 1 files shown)


mgorny:
Thanks. This fixes the problem for me.

src/python/gi/overrides/BlockDev.py
```diff
@@ -942,7 +942,7 @@ def __dir__(self):
         """Let's make TAB-TAB in ipython work!"""
 
         if self._use_local:
-            items = set(dir(self._mod) + locals().keys())
+            items = set(dir(self._mod) + list(locals().keys()))
```
vpodzime: Using `set()` and then union on the two sets would make this look more professional. :wink:

vojtechtrefny: Ok...

Issue #309, opened 2017-12-01
Fix bd_lvm_is_tech_avail()


vpodzime:
It currently doesn't check if the *lvm* utility is available or not:
https://github.com/storaged-project/libblockdev/blob/master/src/plugins/lvm.c#L237

PR #375, opened 2018-07-26
Fix dependency checks in "bd_lvm_is_tech_avail"

vojtechtrefny:
Fixes: #309

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -54,8 +54,6 @@ static gchar *global_config_str = NULL;
 #define SNAP_INTF LVM_BUS_NAME".Snapshot"
 #define THPOOL_INTF LVM_BUS_NAME".ThinPool"
 #define CACHE_POOL_INTF LVM_BUS_NAME".CachePool"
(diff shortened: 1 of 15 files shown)


src/plugins/lvm-dbus.c
```diff
@@ -314,13 +296,43 @@ gboolean bd_lvm_check_deps () {
...
     ret = g_dbus_connection_call_sync (bus, LVM_BUS_NAME, LVM_OBJ_PREFIX, DBUS_INTRO_IFACE,
                                        "Introspect", NULL, NULL, G_DBUS_CALL_FLAGS_NONE,
-                                       -1, NULL, &error);
-    if (ret)
+                                       -1, NULL, error);
+    if (!ret) {
+        if (*error)
+            return FALSE;
+        else
+            return TRUE;
```
vpodzime: Is this really OK? If so, I think it at least deserves a comment why.

vojtechtrefny:
I've created a new "generic" function `bd_utils_dbus_service_available` and moved the check from lvm dbus plugin to `check_deps.c`. Also `call_lvm_method` now checks for the dbus service availability.

PR #376, opened 2018-07-30
Fix lots of compiler warnings

MaxKellermann:
Those warnings are switched off in the `libblockdev` build, but some of them affect programs which include `libblockdev`'s headers.
And maybe it would be a good idea to enable those warnings. Do you wish to have a PR which enables more compiler warnings?

### DIFF
--- a/src/lib/blockdev.c.in
+++ b/src/lib/blockdev.c.in
@@ -183,7 +183,7 @@ static gboolean load_config (GSequence *config_files, GSList **plugins_sonames, 
     return TRUE;
 }
 
-static void unload_plugins () {
+static void unload_plugins (void) {
     if (plugins[BD_PLUGIN_LVM].handle && !unload_lvm (plugins[BD_PLUGIN_LVM].handle))
         g_warning ("Failed to close the lvm plugin");
     plugins[BD_PLUGIN_LVM].handle = NULL;
@@ -378,7 +378,7 @@ static gboolean load_plugins (BDPluginSpec **require_plugins, gboolean reload, g
     return requested_loaded;
 }
 
-GQuark bd_init_error_quark ()
+GQuark bd_init_error_quark (void)
 {
     return g_quark_from_static_string ("g-bd-init-error-quark");
 }
@@ -694,7 +694,7 @@ gboolean bd_try_reinit (BDPluginSpec **require_plugins, gboolean reload, BDUtils
  * was/were called and either at least one plugin is loaded or 0 plugins are
  * loaded after an explicit call that requested 0 plugins to be loaded.
  */
-gboolean bd_is_initialized () {
+gboolean bd_is_initialized (void) {
     gboolean is = FALSE;
     g_mutex_lock (&init_lock);
     is = initialized;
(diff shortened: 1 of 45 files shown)


MaxKellermann:
> ** (process:13503): CRITICAL **: The function 'bd_lvm_get_max_lv_size' called, but not implemented!

Hmm.... I don't get this problem. I assume it's unrelated to this PR?

vojtechtrefny:
>> ** (process:13503): CRITICAL **: The function 'bd_lvm_get_max_lv_size' called, but not implemented!

> Hmm.... I don't get this problem. I assume it's unrelated to this PR?

This is expected "result" of the `test_not_implemented` test cases -- it checks that calling function from a plugin that isn't loaded returns error (and it also prints warning to stderr).

The problem is with one of the tests after this one -- either `test_plugin_fallback` or `test_plugin_priority` -- we somehow managed to kill python :-)

Coredump from the x86_64 rawhide machine (same problem happens on all distros):

```
Process 13077 (python3) of user 0 dumped core.
                                                           
Stack trace of thread 13077:
 #0  0x00007f160b45b86a check_match (ld-linux-x86-64.so.2)
 #1  0x00007f160b45bda4 do_lookup_x (ld-linux-x86-64.so.2)
 #2  0x00007f160b45c63f _dl_lookup_symbol_x (ld-linux-x86-64.so.2)
 #3  0x00007f160ae9004c do_sym (libc.so.6)
 #4  0x00007f160b0a63e8 dlsym_doit (libdl.so.2)
 #5  0x00007f160ae90677 _dl_catch_exception (libc.so.6)
 #6  0x00007f160ae90713 _dl_catch_error (libc.so.6)
 #7  0x00007f160b0a69d9 _dlerror_run (libdl.so.2)
 #8  0x00007f160b0a6453 dlsym (libdl.so.2)
 #9  0x00007f15fcc99215 n/a (/home/jenkins/workspace/libblockdev-PR/arch/x86_64/distro/f_rawhide/type/libblockdev/src/lib/.libs/libblockdev.so.2.0.0)
```

MaxKellermann:
Ok - so is this crash related to my PR? Do I need to do/debug anything? I still don't seem to get it, it looks unrelated.

vojtechtrefny:
> Ok - so is this crash related to my PR? Do I need to do/debug anything? I still don't seem to get it, it looks unrelated.

All the distro-architecture combinations failed the same way, so this is definitely related to these changes.
The problem is somewhere in the `fix -Wstrict-prototypes` commit. The code that loads plugins is quite complicated and it is possible there is some bug that was discovered because of these changes. I'll continue with debugging this tomorrow.

vpodzime:
> @MaxKellermann please add these two commits to the PR -- vojtechtrefny@d03bd02 and vojtechtrefny@4a54217 -- it should fix the failin

@vojtechtrefny you can push these to the PR. https://stackoverflow.com/questions/15530510/how-do-i-push-to-a-pull-request-on-github

MaxKellermann:
@vojtechtrefny OK, cherry-picked to my branch and pushed to this PR.

vojtechtrefny:
Thank you and thank you for the changes.

(The failed test on Debian is not related to these changes, it happens sometimes, there is something broken in mdadm on debian stable.)

PR #381, opened 2018-08-08
vdo: Implement bd_vdo_get_stats()

tbzatek:
This exposes statistics of the specified VDO volume. Instead of calling either
`vdostats` or `vdo info` whose output keys are human-readable strings with
no stability guarantees the statistics are actually taken from /sys/kvdo structure
which seems to be reasonably stable.

There are some keys that are computed by the vdo tools from existing values
so those are backported.

The fact that no calls to the vdo tools are involved makes this function suitable
for reuse within potential VDO support integration in the LVM module. Also since
there's no real I/O involved this function should be reasonably fast.

### DIFF
--- a/src/lib/plugin_apis/vdo.api
+++ b/src/lib/plugin_apis/vdo.api
@@ -113,6 +113,66 @@ GType bd_vdo_info_get_type () {
     return type;
 }
 
+#define BD_VDO_TYPE_STATS (bd_vdo_stats_get_type ())
+GType bd_vdo_stats_get_type();
+
+/**
+ * BDVDOStats:
+ * @block_size: block size of a VDO volume, in bytes
+ * @logical_block_size: logical block size, in bytes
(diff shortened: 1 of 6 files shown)


tbzatek:
Fixed, changed to the `g_memdup()` call.

src/lib/plugin_apis/vdo.api
```diff
@@ -113,6 +113,66 @@ GType bd_vdo_info_get_type () {
...
+ */
+typedef struct BDVDOStats {
+    gint64 block_size;
+    gint64 logical_block_size;
+    gint64 physical_blocks;
+    gint64 data_blocks_used;
+    gint64 overhead_blocks_used;
+    gint64 logical_blocks_used;
+    gint64 used_percent;
+    gint64 saving_percent;
```
vpodzime: Why are these not `guint64`?

PR #385, opened 2018-08-14
Allow building without libkmod if no plugin uses it

MaxKellermann:


vojtechtrefny:
I think this needs a different approach. This patch basically "removes" three public functions from utils based on plugin support and it can be quite confusing. Creating a new configure option (something like `--without-kmod`) that removes dependency on libkmod would be better.

We did similar changes for libdmraid (https://github.com/storaged-project/libblockdev/commit/45737e9d0b12b447a23efdd568207910ae5eaf17) and libvolume_key (https://github.com/storaged-project/libblockdev/commit/7f0d584c4722feecfe3e65de6c5b03c13e3cfd36).

vpodzime:
> Creating a new configure option (something like --without-kmod) that removes dependency on libkmod would be better.

But what should the plugins that require it do then?

vpodzime:
> We could also "fix" this in configure -- if you specify --without-kmod you must disable all plugins that need the "kmod functions".

Yeah, but that's exactly what I would hate to see happening. Most of those plugins are useful even without *libkmod*.

vpodzime:
> The dependency checks should cover this -- check_module_deps will fail (with error message from bd_utils_have_kernel_module saying that libblockdev was build without kmod support) and plugins that need kmod will fail to load. Or with runtime dependency checks enabled only functions that need to check for module dependency will fail -- it should still be possible to use functions that don't need to check or (un)load modules.

Sounds good to me. I believe this is the way it should be done.

MaxKellermann:
> This patch basically "removes" three public functions from utils

That's only an internal static library, and each plugin carries a copy of it... I thought changing its API would not be a problem at all, because it doesn't have an ABI anybody relies on.

@vojtechtrefny's idea could be implemented, but I think this is more complex than the problem deserves a solution to be.

Anyway, when I read the libblockdev source code, I wondered why libblockdev's plugins care about kernel modules at all. Either some feature is present, or the plugin fails with an error message, which may make the admin load that module, but it would be surprising for a library to load modules. (Maybe I'm so ignorant of this feature because all of my kernels have their module support disabled.)

vpodzime:
> Either some feature is present, or the plugin fails with an error message, which may make the admin load that module, but it would be surprising for a library to load modules.

That's because for example setting up `zram` devices is done by loading the module (at least on some older kernel version).

vpodzime:
And in general, we are trying to make *libblockdev* being able to work with (runtime) dependencies missing with only the affected parts not being able to provide the standard functionality. And in those cases, proper and informative errors should be provided. So not just seemingly random failures, proper error messages clearly saying what's missing.

vpodzime:
@MaxKellermann is there still some intention to push this through?

MaxKellermann:
Test

MaxKellermann:
No, I've given up on udisks and libblockdev. But if you wish, I can rebase this PR for you. If not, just close it.

vpodzime:
Let's not spend time on this for now. If somebody wants to push it further, the PR can always be reopened.

vpodzime:
Thanks @MaxKellermann anyways!!!

Issue #391, opened 2018-08-28
When formatting luks1, libblockdev also needs to wipe the gap ("luksmeta nuke")


mvollmer:
See https://bugzilla.redhat.com/show_bug.cgi?id=1594217

There is `luksmeta_nuke` in libluksmeta that can do this.

An alternative is for cryptsetup itself to clear the gap.  I'll try to argue a bit for that as well in https://bugzilla.redhat.com/show_bug.cgi?id=1594217.

mvollmer:
`luksmeta init` sounds reasonable, too, imo.

PR #392, opened 2018-08-29
crypto: Initialize a LUKSv1 disk for luksmeta after formatting

mvollmer:
Fixes: #391

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,6 +193,7 @@ LIBBLOCKDEV_PKG_CHECK_MODULES([KMOD], [libkmod >= 19])
 
 AS_IF([test "x$with_crypto" != "xno"],
       [LIBBLOCKDEV_PKG_CHECK_MODULES([CRYPTSETUP], [libcryptsetup >= 1.6.7])
+      LIBBLOCKDEV_PKG_CHECK_MODULES([LUKSMETA], [luksmeta])
       AS_IF([$PKG_CONFIG --atleast-version=2.0.3 libcryptsetup],
             [AC_DEFINE([LIBCRYPTSETUP_2])], [])
       AS_IF([test "x$with_escrow" != "xno"],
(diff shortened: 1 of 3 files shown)


mvollmer:
I didn't find tests for bd_crypto_format et al.  Are there some that I can extend?

vpodzime:
> I didn't find tests for bd_crypto_format et al. Are there some that I can extend?

Yes, please check the *tests/crypto_test.py* file.

vojtechtrefny:
I still don't understand why libblockdev should be doing this. I understand why you want to remove the "stale" metadata but this needs a better solution -- something "system wide". Ideally blkid/wipefs should be able to tell that the device is not empty after removing the luks header and it should be possible to remove it using wipefs. Wipefs already can "detect" and remove nested metadata for some formats, why not in this case?

I also don't like the idea of changing behaviour of `luks_format` like this -- if we really need this, then it should be a separate function that must be called "manually" after luks format. Our functions should be as close to cryptsetup functionality as possible.

From my point of view this is like telling us to run `pvremove` after luks format just to be sure there is no stale lvm metadata.

There must be a better way to fix this. Solving this issue only in libblockdev doesn't make sense. If someone calls "cryptsetup luksFormat" manually you have the same issue.

mvollmer:
> Yes, please check the tests/crypto_test.py file.

Ouch, how could I miss this...

mvollmer:
> I still don't understand why libblockdev should be doing this.

My only argument is "because cryptsetup doesn't want to", see https://bugzilla.redhat.com/show_bug.cgi?id=1594217.

I think the best place is in cryptsetup.  When it writes the header, it should write zeroes into the gap.  (Or directly initialize it for luksmeta, because that is convenient.)

mvollmer:
Let's not do this here.  I still don't understand why he cryptsetup people don't want to wipe the gap as well when formatting LUKS1, but hopefully nobody gets hurt.

PR #397, opened 2018-09-18
Fix build of plugins by changing linking order

devurandom:
Link against libbd_utils.la before any *_LIBS can change the linker path.

Related: 12b2506d4853e515e9914d5c761b5501df1242fe

Signed-off-by: Dennis Schridde <devurandom@gmx.net>


vojtechtrefny:
Jenkins, ok to test.

vpodzime:
Thanks for this fix!

PR #404, opened 2018-10-01
Use libblkid to check swap status before swapon

vojtechtrefny:
libblkid probe is more reliable than our custom check.

### DIFF
--- a/src/plugins/swap.c
+++ b/src/plugins/swap.c
@@ -21,6 +21,8 @@
 #include <string.h>
 #include <unistd.h>
 #include <sys/swap.h>
+#include <fcntl.h>
+#include <blkid.h>
 #include <blockdev/utils.h>
 
 #include "swap.h"
@@ -179,13 +181,14 @@ gboolean bd_swap_mkswap (const gchar *device, const gchar *label, const BDExtraA
  * Tech category: %BD_SWAP_TECH_SWAP-%BD_SWAP_TECH_MODE_ACTIVATE_DEACTIVATE
  */
 gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
-    GIOChannel *dev_file = NULL;
-    GIOStatus io_status = G_IO_STATUS_ERROR;
-    GError *tmp_error = NULL;
-    gsize num_read = 0;
-    gchar dev_status[11];
-    dev_status[10] = '\0';
-    gint page_size;
+    blkid_probe probe = NULL;
+    gint fd = 0;
+    gint status = 0;
+    guint n_try = 0;
+    const gchar *value = NULL;
+    gint64 status_len = 0;
+    gint64 swap_pagesize = 0;
+    gint64 sys_pagesize = 0;
     gint flags = 0;
     gint ret = 0;
     guint64 progress_id = 0;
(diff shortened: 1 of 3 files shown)


src/plugins/swap.c
```diff
@@ -198,56 +201,150 @@ gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
...
         bd_utils_report_finished (progress_id, (*error)->message);
+        blkid_free_probe (probe);
         return FALSE;
     }
 
-    page_size = getpagesize ();
-    page_size = MAX (2048, page_size);
-    io_status = g_io_channel_seek_position (dev_file, page_size - 10, G_SEEK_SET, &tmp_error);
-    if (io_status != G_IO_STATUS_NORMAL) {
+    /* we may need to try mutliple times with some delays in case the SWAPDEVICE is
```
vpodzime: `the device`, no `SWAPDEVICE` in this codebase ;)

src/plugins/swap.c
```diff
@@ -198,56 +201,150 @@ gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
...
         return FALSE;
     }
 
-    io_status = g_io_channel_read_chars (dev_file, dev_status, 10, &num_read, &tmp_error);
-    if ((io_status != G_IO_STATUS_NORMAL) || (num_read != 10)) {
+    blkid_probe_enable_partitions (probe, 1);
+    blkid_probe_set_partitions_flags (probe, BLKID_PARTS_MAGIC);
+    blkid_probe_enable_superblocks (probe, 1);
+    blkid_probe_set_superblocks_flags (probe, BLKID_SUBLKS_USAGE | BLKID_SUBLKS_TYPE |
+                                              BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_BADCSUM);
```
vpodzime: Do we really want these features? Not sure what *superblocks* here are, but we are definitely not probing for partitions.

src/plugins/swap.c
```diff
@@ -198,56 +201,150 @@ gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
...
+        close (fd);
         return FALSE;
     }
 
-    g_io_channel_shutdown (dev_file, FALSE, &tmp_error);
-    g_clear_error (&tmp_error);
+    status = blkid_probe_lookup_value (probe, "TYPE", &value, NULL);
+    if (status != 0) {
+        g_set_error (error, BD_SWAP_ERROR, BD_SWAP_ERROR_UNKNOWN_STATE,
+                     "Failed to get filesystem type for the device '%s'", device);
```
vpodzime: *format type*

src/plugins/swap.c
```diff
@@ -198,56 +201,150 @@ gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
...
+        bd_utils_report_finished (progress_id, (*error)->message);
+        blkid_free_probe (probe);
+        close (fd);
+        return FALSE;
+    }
 
-    if (g_str_has_prefix (dev_status, "SWAP-SPACE")) {
+    if (g_strcmp0 (value, "swap") != 0) {
+        g_set_error (error, BD_SWAP_ERROR, BD_SWAP_ERROR_UNKNOWN_STATE,
+                     "Device '%s' is not a swap", device);
```
vpodzime: *is not formatted as swap*

src/plugins/swap.c
```diff
@@ -198,56 +201,150 @@ gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error) {
...
                      "Unknown swap space format, cannot activate.");
         bd_utils_report_finished (progress_id, (*error)->message);
+        blkid_free_probe (probe);
+        close (fd);
         return FALSE;
     }
 
+    status_len = (gint64) strlen (value);
+
+    status = blkid_probe_lookup_value (probe, "SBMAGIC_OFFSET", &value, NULL);
```
vpodzime: I think this is the last place the `probe` is needed so it should be free'd here.

vojtechtrefny: `value` is owned by the probe, so we can't free it here.

PR #410, opened 2018-11-25
Fixed a number of memory leaks in the lvm-dbus plugin

wtfuzz:


src/plugins/lvm-dbus.c
```diff
@@ -2197,7 +2213,7 @@ static gboolean filter_lvs_by_vg (const gchar **lvs, const gchar *vg_name, GSLis
             }
 
             if (g_strcmp0 (lv_vg_name, vg_name) == 0) {
-                *out = g_slist_prepend (*out, (gchar *) *lv_p);
+                *out = g_slist_prepend (*out, g_strdup((gchar *) *lv_p));
```
vpodzime: This function should avoid creating duplicates.

src/lib/plugin_apis/lvm.api
```diff
@@ -136,6 +136,7 @@ void bd_lvm_pvdata_free (BDLVMPVdata *data) {
     g_free (data->pv_name);
     g_free (data->pv_uuid);
     g_free (data->vg_name);
+    g_free (data->vg_uuid);
```
vpodzime: Please also update the copy function here.

vpodzime:
Jenkins, this is ok to test.

src/plugins/lvm-dbus.c
wtfuzz: Given how this function is used, it's difficult to not duplicate the string in a clean way. It would either have to deallocate strings that don't match the VG filter internally and then unref just the variant container afterwards, or g_strfreev() would have to be replaced with loops that free the strings in the input vector that are not in the matched_lvs vector.

```
    lvs = get_existing_objects (LV_OBJ_PREFIX, error);
    if (!lvs && (*error)) {
        /* error is already populated */
        return NULL;
    }
    success = filter_lvs_by_vg (lvs, vg_name, &matched_lvs, &n_lvs, error);
    g_strfreev ((gchar **) lvs);
    if (!success)
        return NULL;
```

src/lib/plugin_apis/lvm.api
wtfuzz: Done.

src/plugins/lvm-dbus.c
tbzatek: I'd suggest to make this duplicate-records concern separate from this pull request.

src/plugins/lvm-dbus.c
```diff
@@ -2239,7 +2255,7 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
         return NULL;
     }
     success = filter_lvs_by_vg (lvs, vg_name, &matched_lvs, &n_lvs, error);
-    g_free (lvs);
+    g_strfreev ((gchar **) lvs);
```
tbzatek: The typecast shouldn't be here, the `get_existing_objects()` function should not return `const gchar**` in the first place since it returns new fully allocated array, no sign of reusing foreign memory that would qualify it for `const`.

Could you fix it altogether please?

src/plugins/lvm-dbus.c
```diff
@@ -2290,6 +2306,7 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
     lv = matched_lvs;
     while (lv) {
         props = get_object_properties (lv->data, LV_CMN_INTF, error);
+        g_free(lv->data);
```
tbzatek: Hmm, there's a risk of leaking memory after first iteration since the rest of the items wouldn't be freed. Better to replace all `g_slist_free()` calls with `g_slist_free_full()`.

Plus missing space before opening parenthesis :-)

src/plugins/lvm-dbus.c
```diff
@@ -398,7 +398,7 @@ static const gchar** get_existing_objects (const gchar *obj_prefix, GError **err
         return NULL;
 
     g_variant_get (intro_v, "(s)", &intro_data);
-    g_variant_unref(intro_v);
+    g_variant_unref (intro_v);
```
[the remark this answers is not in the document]
wtfuzz: I've squashed all changes into a single commit.

src/plugins/lvm-dbus.c
wtfuzz: Nice catch. I've updated the PR to use ``g_slist_free_full(matched_lvs, g_free)``

src/plugins/lvm-dbus.c
vpodzime: > Given how this function is used, it's difficult to not duplicate the string in a clean way. It would either have to deallocate strings that don't match the VG filter internally

Sorry, but I don't follow. Isn't the original code in the function doing exactly that? The function takes a pointer to an array of strings, the matching ones are moved elsewhere, the non-matching ones are free'd. And then this array itself is free'd after the function call. Don't really see a problem here. What am I missing?

wtfuzz: After looking at that function again, I see what you're saying. I'll clean it up and test with valgrind tomorrow.

src/plugins/lvm-dbus.c
```diff
@@ -2249,7 +2265,7 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
         return NULL;
```
tbzatek: Btw. reading the code all over again, there's a number of other leaks in this function - e.g. in this case of return, `matched_lvs` would be leaked. This is endless... would you like me to add a fix for you or would you like to fix that yourself? :-)

src/plugins/lvm-dbus.c
```diff
@@ -2290,13 +2306,14 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
     lv = matched_lvs;
     while (lv) {
         props = get_object_properties (lv->data, LV_CMN_INTF, error);
+        g_free (lv->data);
```
tbzatek: Since you're doing proper `g_slist_free_full()` in error cases and the final success case now, this would result in double-free in the error cases since you're freeing the original list, not the iterated one (which is correct). 

Suggesting to remove this line.

src/plugins/lvm-dbus.c
wtfuzz: I certainly haven't looked at all return paths :) This should just be a matter of having ``g_slist_free_full (matched_lvs, g_free)`` before all returns.

src/plugins/lvm-dbus.c
wtfuzz: Good catch.

wtfuzz: Removed in updated commit.

src/plugins/lvm-dbus.c
```diff
@@ -2240,8 +2255,10 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
...
     success = filter_lvs_by_vg (lvs, vg_name, &matched_lvs, &n_lvs, error);
     g_free (lvs);
-    if (!success)
+    if (!success) {
+        g_slist_free_full (matched_lvs, g_free);
         return NULL;
+    }
 
     lvs = get_existing_objects (THIN_POOL_OBJ_PREFIX, error);
     if (!lvs && (*error)) {
```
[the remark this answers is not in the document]
wtfuzz: Thanks for noticing this. I've updated the commit.

vojtechtrefny:
Jenkins, test this please.

src/plugins/lvm-dbus.c
```diff
@@ -677,6 +684,8 @@ static void call_lvm_method_sync (const gchar *obj, const gchar *intf, const gch
     log_task_status (log_task_id, log_msg);
     g_free (log_msg);
 
+    g_variant_unref (ret);
```
vojtechtrefny: We have one extra unref somwhere here (`Warning: g_variant_unref: assertion 'value->ref_count > 0' failed` in the test results) and I think it's this one -- we already have unref calls in all branches above, so it shouldn't be needed here, right?

wtfuzz: @vojtechtrefny Yes, all branches have unref calls above, so this was an extra. I've updated the commit to remove it. Thanks for the catch!

src/plugins/lvm-dbus.c
```diff
@@ -909,6 +916,7 @@ static BDLVMPVdata* get_pv_data_from_props (GVariant *props, GError **error) {
     }
 
     vg_props = get_object_properties (value, VG_INTF, error);
+    g_free (value);
```
vpodzime: Shouldn't the above `if` block free `value` too?

vpodzime:
@wtfuzz last two nitpicks. Could you please take a look at them?

src/plugins/lvm-dbus.c
tbzatek: Correct.

So just a hint here - `GVariant` format strings allow either handing ownership or keeping ownership of the memory. In this case changing the format string to `&o` and the type of the variable to `const gchar *` would save you the need for explicit deallocation. Just be sure not to free the enclosing `GVariant` while this string is still used (it's stack allocated here though).

vpodzime:
ping @wtfuzz

src/plugins/lvm-dbus.c
vpodzime: Changing to `&o`, thanks for the tip!

src/plugins/lvm-dbus.c
```diff
@@ -1935,6 +1945,8 @@ gboolean bd_lvm_lvcreate (const gchar *vg_name, const gchar *lv_name, guint64 si
     }
 
     call_lvm_obj_method_sync (vg_name, VG_INTF, "LvCreate", params, extra_params, extra, error);
+
+    g_variant_unref (params);
```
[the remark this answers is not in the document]
vpodzime: It's actually doing it by passing both `params` and `extra_params` to `g_dbus_connection_call_sync()`.

vpodzime:
Jenkins, test this, please.

PR #413, opened 2018-12-17
Use major/minor macros from sys/sysmacros.h instead of linux/kdev_t.h

vojtechtrefny:
The macros from linux/kdev_t.h doesn't work for devices with
minor > 255.

Resolves: rhbz#1644825

### DIFF
--- a/src/plugins/mpath.c
+++ b/src/plugins/mpath.c
@@ -18,8 +18,8 @@
  */
 
 #include <glib.h>
-/* provides MAJOR, MINOR macros */
-#include <linux/kdev_t.h>
+/* provides major and minor macros */
+#include <sys/sysmacros.h>
 #include <libdevmapper.h>
 #include <unistd.h>
 #include <blockdev/utils.h>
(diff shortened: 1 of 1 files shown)


src/plugins/mpath.c
```diff
@@ -18,8 +18,8 @@
  */
 
 #include <glib.h>
-/* provides MAJOR, MINOR macros */
-#include <linux/kdev_t.h>
+/* provides major and minor functions */
```
vpodzime: *macros*, right?

src/plugins/mpath.c
```diff
@@ -319,8 +319,8 @@ static gchar** get_map_deps (const gchar *map_name, guint64 *n_deps, GError **er
     dep_devs = g_new0 (gchar*, deps->count + 1);
 
     for (i = 0; i < deps->count; i++) {
-        major = (guint64) MAJOR(deps->device[i]);
-        minor = (guint64) MINOR(deps->device[i]);
+        major = (guint64) major (deps->device[i]);
+        minor = (guint64) minor (deps->device[i]);
```
vpodzime: Please change the names of the variables to not collide with the macros.

PR #414, opened 2018-12-19
CentOS 7.6 LUKS 2 related fixes

vojtechtrefny:
CentoOS 7.6 has oder kernel that doesn't support storing keys in kernel keyring. Also older util-linux doesn't fully support LUKS 2.

### DIFF
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -1444,6 +1444,7 @@ gboolean bd_crypto_luks_change_key (const gchar *device, const gchar *pass, cons
 
 static gboolean luks_resize (const gchar *luks_device, guint64 size, const guint8 *pass_data, gsize data_len, const gchar *key_file, GError **error) {
     struct crypt_device *cd = NULL;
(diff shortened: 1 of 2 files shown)


src/plugins/crypto.c
```diff
@@ -1463,6 +1464,15 @@ static gboolean luks_resize (const gchar *luks_device, guint64 size, const guint
         return FALSE;
     }
 
+    ret = crypt_get_active_device (cd, luks_device, &cad);
+    if (ret != 0) {
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_DEVICE,
+                     "Failed to get information about '%s': %s",
+                     luks_device, strerror_l(-ret, c_locale));
+        bd_utils_report_finished (progress_id, (*error)->message);
+        return FALSE;
```
tbzatek: Leaking `cd`.

vojtechtrefny: Fixed.

PR #419, opened 2019-02-05
Vagrant template and spec update

vojtechtrefny:
autoconf-archive dependency was added in #418

### DIFF
--- a/misc/Vagrantfile.tmpl
+++ b/misc/Vagrantfile.tmpl
@@ -75,7 +75,7 @@ Vagrant.configure("2") do |config|
       apt-get -y install libblockdev2 libblockdev-* python3-blockdev
       apt-get -y build-dep libblockdev
       apt-get -y install libfdisk-dev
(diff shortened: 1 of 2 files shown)


misc/Vagrantfile.tmpl
```diff
@@ -31,10 +31,10 @@ Vagrant.configure("2") do |config|
     echo './autogen.sh && ./configure --without-vdo --without-nvdimm' >> /home/vagrant/.bash_history
   SHELL
 
-  config.vm.define "bd-f28", primary: true, autostart: true do |f28|
-    # F28-specific configuration
-    f28.vm.box = "fedora/28-cloud-base"
-    f28.vm.provision :shell, inline: <<-SHELL
+  config.vm.define "bd-f29", primary: true, autostart: true do |f29|
```
vpodzime: We might make this more automatic by adding some more Ruby logic to the file. A fun task for the future, I guess. :-)

PR #420, opened 2019-02-05
lvm: Add exported information to BDLVMVGdata

vojtechtrefny:
Blivet will use this new information to mark exported VGs as protected -- we had some anaconda bugs where installation crashed because user tried to remove an exported VG.
Unfortunately this doesn't work with the DBus plugin. This is already fixed upstream -- https://sourceware.org/git/?p=lvm2.git;a=commit;h=5bdcafff47ee6a2e57e0c57d18cbc36cf1a2973e -- but not yet released.

### DIFF
--- a/tests/lvm_dbus_tests.py
+++ b/tests/lvm_dbus_tests.py
@@ -1420,3 +1420,40 @@ class LvmTestPVremoveConfig(LvmPVonlyTestCase):
 
         succ = BlockDev.lvm_pvremove(self.loop_dev2, None)
         self.assertTrue(succ)
+
+@unittest.skipUnless(lvm_dbus_running, "LVM DBus not running")
+@unittest.skip("LVM DBus crashes if an exported VG is present in the system")
+class LvmVGExportedTestCase(LvmPVVGLVTestCase):
+
+    def _clean_up(self):
+        run_command("vgimport testVG")
+
+        LvmPVVGLVTestCase._clean_up(self)
+
+    @tag_test(TestTags.SLOW)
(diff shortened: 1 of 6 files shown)


tests/lvm_dbus_tests.py
```diff
@@ -1408,3 +1408,40 @@ def test_pvremove_with_config(self):
...
+@unittest.skip("LVM DBus crashes if an exported VG is present in the system")
+class LvmVGExportedTestCase(LvmPVVGLVTestCase):
+
+    def _clean_up(self):
+        run_command("vgimport testVG")
+
+        LvmPVVGLVTestCase._clean_up(self)
+
+    @unittest.skipIf("SKIP_SLOW" in os.environ, "skipping slow tests")
+    def test_exported_(self):
```
vpodzime: Trailing `_`?

vpodzime:
> This requires a rebuild of everything using _libblockdev_.

AFAICT we haven't done this since *libblockdev-2.0*.

tests/lvm_dbus_tests.py
vojtechtrefny: Oops, missing `vg`, fixed.

vojtechtrefny:
> This requires a rebuild of everything using _libblockdev_. What do you think about adding a new function to query this?

I don't like the idea of adding a new function just to do this. The blivet issue is not that important and it can wait for 3.0

vojtechtrefny:
Rebased to current master. Also master is now 3.0, so we can merge this now without worrying about he ABI change.

PR #428, opened 2019-02-12
Fix checking swap status on lvm/md

vojtechtrefny:
'bd_utils_resolve_device' returns already resolved part, there is
no '../' to remove.

### DIFF
--- a/tests/swap_test.py
+++ b/tests/swap_test.py
@@ -2,7 +2,7 @@ import unittest
 import os
 import overrides_hack
 
-from utils import create_sparse_tempfile, create_lio_device, delete_lio_device, fake_utils, fake_path, run_command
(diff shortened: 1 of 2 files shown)


tests/swap_test.py
```diff
@@ -109,6 +109,32 @@ def test_swapon_pagesize(self):
...
+
+        succ = BlockDev.swap_swapon("/dev/mapper/%s" % dm_name, -1)
+        self.assertTrue(succ)
+
+        on = BlockDev.swap_swapstatus("/dev/mapper/%s" % dm_name)
+        self.assertTrue(on)
+
+        succ = BlockDev.swap_swapoff("/dev/mapper/%s" % dm_name)
+        self.assertTrue(succ)
+
```
vpodzime: One more `swapstatus()` check here would be nice.

vojtechtrefny: Fixed.

PR #429, opened 2019-02-28
Sync spec file with python2 obsoletion added downstream

AdamWill:
Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/dist/libblockdev.spec.in
+++ b/dist/libblockdev.spec.in
@@ -125,7 +125,7 @@
 
 Name:        libblockdev
 Version:     2.21
-Release:     1%{?dist}
+Release:     2%{?dist}
 Summary:     A library for low-level manipulation with block devices
 License:     LGPLv2+
 URL:         https://github.com/storaged-project/libblockdev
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Why is this needed? I thought the plan is just to stop shipping python2 packages and no other steps are needed. I have more python packages that dropped the python2 subpackage in Fedora > 29 -- should I add the `Obsoletes: python2-xyz` everywhere?

AdamWill:
See the discussion on devel@ : the problem is that we run into dependency issues on upgrade. dnf does not want to remove packages unless they are explicitly obsoleted. It complains because it cannot keep the old python2-blockdev (because it depends on the matching libblockdev version, which is not available in the new release) but there is no new python2-blockdev for it to update to.

There is a `--allowerasing` arg for dnf which allows it to continue in this case, but it is a dangerous arg because it tells dnf it's fine to just go ahead and remove *any* package whose deps it can't satisfy. Sometimes this can wind up removing something really important due to some kind of dep issue. So we really want to try and make it so `--allowerasing` is not typically necessary for upgrades.

vojtechtrefny:
Thank you for the explanation, I completely missed the fedora-devel thread. I'll check our other projects, but this should be only problematic package.

PR #430, opened 2019-03-05
New function to get supported sector sizes for NVDIMM namespaces

vojtechtrefny:
Unfortunately 'ndctl_namespace_get_supported_sector_size' needs
an existing namespace to tell us what sector sizes it supports so
we need to hard code the values.

### DIFF
--- a/src/plugins/nvdimm.c
+++ b/src/plugins/nvdimm.c
@@ -636,3 +636,38 @@ gboolean bd_nvdimm_namespace_reconfigure (const gchar* namespace, BDNVDIMMNamesp
     g_free ((gchar *) args[5]);
     return ret;
 }
+
+
+static guint64 blk_sector_sizes[] = { 512, 520, 528, 4096, 4104, 4160, 4224, 0 };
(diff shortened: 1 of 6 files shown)


src/plugins/nvdimm.c
```diff
@@ -636,3 +636,53 @@ gboolean bd_nvdimm_namespace_reconfigure (const gchar* namespace, BDNVDIMMNamesp
...
+static guint64 blk_sector_sizes[] = { 512, 520, 528, 4096, 4104, 4160, 4224, 0 };
+static guint64 pmem_sector_sizes[] = { 512, 4096, 0 };
+static guint64 io_sector_sizes[] = { 0 };
+
+/**
+ * bd_nvdimm_namepace_get_supported_sector_sizes:
+ * @mode: namespace mode
+ * @error: (out): place to store error if any
+ *
+ * Returns: (transfer full) (array zero-terminated=1): list of supported sector sizes for @mode
```
vpodzime: I wonder if we could just return the static array as `const guint64 *` with `(transfer none)`.

vojtechtrefny: Fixed.

PR #431, opened 2019-03-12
Use existing cryptsetup API for changing keyslot passphrase

vojtechtrefny:
Instead of manually removing the keyslot and adding new a one.
Our old code also doesn't work in FIPS mode.

### DIFF
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -1368,8 +1368,6 @@ gboolean bd_crypto_luks_remove_key (const gchar *device, const gchar *pass, cons
 gboolean bd_crypto_luks_change_key_blob (const gchar *device, const guint8 *pass_data, gsize data_len, const guint8 *npass_data, gsize ndata_len, GError **error) {
     struct crypt_device *cd = NULL;
     gint ret = 0;
-    gchar *volume_key = NULL;
-    gsize vk_size = 0;
     guint64 progress_id = 0;
     gchar *msg = NULL;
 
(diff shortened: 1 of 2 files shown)


src/plugins/crypto.c
```diff
@@ -1394,41 +1392,17 @@ gboolean bd_crypto_luks_change_key_blob (const gchar *device, const guint8 *pass
...
-        crypt_free (cd);
-        g_free (volume_key);
-        bd_utils_report_finished (progress_id, (*error)->message);
-        return FALSE;
-    }
-
-    ret = crypt_keyslot_add_by_volume_key (cd, ret, volume_key, vk_size, (char*) npass_data, ndata_len);
+    ret = crypt_keyslot_change_by_passphrase (cd, CRYPT_ANY_SLOT, CRYPT_ANY_SLOT,
+                                              (char*) pass_data, data_len,
+                                              (char*) npass_data, ndata_len);
```
vpodzime: One disadvantage of this is that we lose the information about which phase failed. What is the error we get when the old passphrase is incorrect?

vojtechtrefny: It returns `EPERM` so just the standard "Operation not permitted". But more information should be available in the log (and we redirect cryptsetup log to our log, so it should be enough).

vpodzime: Aren't we processing `EPERM` in a special way somewhere else in this plugin?

vojtechtrefny: Good point. I've added custom error message.

PR #432, opened 2019-03-14
Test tags 

vojtechtrefny:
This is a work in progress pull request adding a new way to skip the tests and run only subset(s) of the testsuite.
The goal here is to be able to "categorize" the tests in a way that allows not only to skip slow and/or unstable tests, but also to run only slow and/or unstable tests. This way we can for example create a separate test run that runs only unstable tests to see if something changed or run only some small subset of the tests to quickly check core libblockdev functionality.

Example of the output with skipped slow tests -- https://paste.fedoraproject.org/paste/pCjYWZkNi1gIsGetYcLy7g

Known issues:
- skipped tests are not counted in the result -- this would require to run the tests manually and/or write a custom test runner to replace `unittest.TextTestRunner`
- the `run_tests.py` cmdline arguments are quite confusing right now -- some skip tests, some select "exclusive" tests to run, it might be better to use something `run_tests.py --skip slow unsafe --run-only core` 

Not implemented yet:
- part of this should be also adding a config file for tests that should be skipped "temporarily", e.g. only on Fedora 27
- ability to run the test suite against installed version of libblockdev

### DIFF
--- a/tests/crypto_test.py
+++ b/tests/crypto_test.py
@@ -9,7 +9,7 @@ import locale
 import re
 import tarfile
 
-from utils import create_sparse_tempfile, create_lio_device, delete_lio_device, skip_on, get_avail_locales, requires_locales, run_command, read_file, TestTags, tag_test
+from utils import create_sparse_tempfile, create_lio_device, delete_lio_device, get_avail_locales, requires_locales, run_command, read_file, TestTags, tag_test
 from gi.repository import BlockDev, GLib
 
 PASSWD = "myshinylittlepassword"
@@ -627,8 +627,6 @@ class CryptoTestEscrow(CryptoTestCase):
         self.addCleanup(os.unlink, self.public_cert)
 
     @tag_test(TestTags.SLOW)
-    @skip_on(("centos", "enterprise_linux"), "7", reason="volume_key asks for password in non-interactive mode on this release")
-    @skip_on("debian", skip_on_version="9", reason="volume_key asks for password in non-interactive mode on this release")
     def test_escrow_packet(self):
         """Verify that an escrow packet can be created for a device"""
 
--- a/tests/crypto_test.py
+++ b/tests/crypto_test.py
@@ -9,7 +9,7 @@ import locale
 import re
 import tarfile
 
-from utils import create_sparse_tempfile, create_lio_device, delete_lio_device, skip_on, get_avail_locales, requires_locales, run_command, read_file
+from utils import create_sparse_tempfile, create_lio_device, delete_lio_device, skip_on, get_avail_locales, requires_locales, run_command, read_file, TestTags, tag_test
 from gi.repository import BlockDev, GLib
 
 PASSWD = "myshinylittlepassword"
@@ -92,6 +92,7 @@ class CryptoTestGenerateBackupPassphrase(CryptoTestCase):
         # we don't need block devices for this test
         pass
 
+    @tag_test(TestTags.NOSTORAGE)
     def test_generate_backup_passhprase(self):
         """Verify that backup passphrase generation works as expected"""
 
@@ -101,7 +102,7 @@ class CryptoTestGenerateBackupPassphrase(CryptoTestCase):
             six.assertRegex(self, bp, exp)
 
 class CryptoTestFormat(CryptoTestCase):
-    @unittest.skipIf("SKIP_SLOW" in os.environ, "skipping slow tests")
+    @tag_test(TestTags.SLOW, TestTags.CORE)
     def test_luks_format(self):
         """Verify that formating device as LUKS works"""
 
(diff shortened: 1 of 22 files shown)


tests/run_tests.py
```diff
@@ -16,6 +16,82 @@
...
+    elif skip_tag == TestTags.UNSTABLE:
+        reason = "skipping unstable tests"
+    elif skip_tag == TestTags.UNSAFE:
+        reason = "skipping test that modifies system configuration"
+    elif skip_tag == TestTags.EXTRADEPS:
+        reason = "skipping test that requires special configuration"
+    elif skip_tag == TestTags.CORE:
+        reason = "skipping non-core test"
+    else:
+        reason = ""  # just to be sure there is some default value
```
vpodzime: Something like `"unknown reason"` would be better here.

tests/run_tests.py
```diff
@@ -57,21 +136,60 @@
...
+    if 'FEELINGLUCKY' in os.environ:
+        args.lucky = True
+
     sys.path.append(testdir)
     sys.path.append(projdir)
     sys.path.append(os.path.join(projdir, 'src/python'))
 
     start_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
 
+    loader = unittest.defaultTestLoader
```
vpodzime: So we don't want to create an instance of the loader here?

tests/run_tests.py
```diff
@@ -45,6 +121,9 @@
     argparser.add_argument('-j', '--jenkins', dest='jenkins',
                            help='run also tests that should run only in a CI environment',
                            action='store_true')
+    argparser.add_argument('-c', '--core', dest='core',
+                           help='run only tests that cover basic functionality of the library',
```
vpodzime: And regression tests, right?

tests/run_tests.py
vojtechtrefny: `unittest.defaultTestLoader` [provides an instance](https://docs.python.org/3/library/unittest.html#unittest.TestLoader) and we don't need a "custom" one.

tests/run_tests.py
vojtechtrefny: Fixed.

tests/run_tests.py
vojtechtrefny: Yes. I'll probably change the command line options to allow more detailed selection, but right now it runs regression tests too. Fixed.

tests/crypto_test.py
```diff
@@ -234,7 +235,7 @@ def _get_key_location(self, device):
 
         return m.group(1)
 
-    @unittest.skipIf("SKIP_SLOW" in os.environ, "skipping slow tests")
+    @tag_test(TestTags.SLOW)
```
vpodzime: What is the key to say if the test should be `CORE` or not? It should also be explained in the commit message.

tests/run_tests.py
```diff
@@ -129,6 +129,8 @@ def _print_skip_message(test, skip_tag):
         reason = "skipping test that requires special configuration"
     elif skip_tag == TestTags.CORE:
         reason = "skipping non-core test"
+    elif skip_tag == TestTags.SOURCEONLY:
+        reason = "skipping test that can run only against library compiled from source"
```
vpodzime: I think this commit can be squashed into the previous one.

vojtechtrefny: Fixed.

tests/crypto_test.py
vojtechtrefny: That's a good question. I think it's not possible to create some rules for this. I basically went through all the tests and if a test looked "important" I marked it as "core".

vpodzime: I'm okay with that, but I still think it should be mentioned in the commit message.

vojtechtrefny: OK, I've added detailed description of every tag to the commit message.

vojtechtrefny:
I have added new commits with support for skipping tests using a config file, instead of the `skip_on` decorator.
The idea here is that it will be easier to change the config file or even have a special for config file for some machines etc.

tests/run_tests.py
```diff
@@ -140,6 +149,37 @@ def _print_skip_message(test, skip_tag):
...
+    # all these can be lists or a single value, so covert everything to list
+    if distro is not None and type(distro) is not list:
+        distro = [distro]
+    if version is not None and type(version) is not list:
+        version = [version]
+    if arch is not None and type(arch) is not list:
+        arch = [arch]
+
+    if (distro is None or DISTRO in distro) and (version is None or VERSION in version) and \
+       (arch is None or ARCH in arch):
```
vpodzime: Please add a comment about where the global variables are defined and why (to not do it for every test, I guess).

vojtechtrefny: Comments added.

PR #436, opened 2019-04-11
WIP: Lock newly created LVs before calling lvconvert

vojtechtrefny:
UDev probes the new LVs and lvcovert fails because the device
(usually the metadata LV) is busy.


vojtechtrefny:
Ok, that didn't work, keeping the lvs locked is problem for LVM too.

PR #440, opened 2019-05-15
Leaks2

tbzatek:
Another bunch of small fixes. There are still a couple of left, to be added (given that they aren't udisks issues in fact...).

### DIFF
--- a/src/plugins/mdraid.c
+++ b/src/plugins/mdraid.c
@@ -178,7 +178,7 @@ gboolean bd_md_check_deps (void) {
     }
 
     if (!ret)
-        g_warning("Cannot load the MDRAID plugin");
+        g_warning ("Cannot load the MDRAID plugin");
 
     return ret;
 }
@@ -275,7 +275,7 @@ static GHashTable* parse_mdadm_vars (const gchar *str, const gchar *item_sep, co
     return table;
 }
 
-static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table, GError **error) {
+static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table, G_GNUC_UNUSED GError **error) {
     BDMDExamineData *data = g_new0 (BDMDExamineData, 1);
     gchar *value = NULL;
     gchar *first_space = NULL;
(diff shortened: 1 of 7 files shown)


tbzatek:
Okay, I think I'm done with memory leaks hunting. There are couple of minor ones left in udisks, nothing in libblockdev so far.

src/plugins/mdraid.c
```diff
@@ -357,8 +357,7 @@ static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean
         }
 
         if (bs_error) {
-            g_set_error (error, BD_MD_ERROR, BD_MD_ERROR_PARSE,
-                         "Failed to parse chunk size from mdexamine data: %s", bs_error->msg);
+            g_warning ("get_examine_data_from_table(): Failed to parse chunk size from mdexamine data: %s", bs_error->msg);
```
vpodzime: Why this change?

src/plugins/mdraid.c
```diff
@@ -275,7 +275,7 @@ static GHashTable* parse_mdadm_vars (const gchar *str, const gchar *item_sep, co
     return table;
 }
 
-static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table, GError **error) {
+static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table, G_GNUC_UNUSED GError **error) {
```
vpodzime: This should be squashed with the previous commit. And I believe the `error` should actually be used.

src/plugins/mdraid.c
tbzatek: Ah, that. We've agreed with @vojtechtrefny that this shouldn't fail the call - it's a single attribute that is not that important when missing. This change also mimics the previous behaviour where data have been returned successfully and the error got simply lost.

src/plugins/mdraid.c
tbzatek: Squashed. We could perhaps make use of `error` by defining couple of keys as mandatory. Just a thought.

src/plugins/mdraid.c
vpodzime: OK

src/plugins/mdraid.c
vpodzime: We don't have to do that now.

vojtechtrefny:
Jenkins, test this please.

PR #443, opened 2019-06-11
tests: Fix Debian testing "version" for skipping

vojtechtrefny:
Debian now reports version as "10" instead of "testing" for buster.


vojtechtrefny:
> Interesting, this will break sooner or later again once Debian moves to release 11. However, this may actually work on newer releases.

I plan to add a better configuration for skipping in the future. We have similar problem with skipping tests on rawhide.

PR #446, opened 2019-06-14
Use libfdisk to get partition spec

vojtechtrefny:


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

src/plugins/part.c
```diff
@@ -582,6 +590,129 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
+static BDPartFlag get_flag_from_guid (const gchar *guid) {
+    guint last_flag = 0;
+
+    last_flag = log2i (BD_PART_FLAG_BASIC_LAST);
+
+    for (guint i = 1; i < last_flag; i++) {
+      if (!part_flags[i - 1].guid)
+          continue;
+
+      if (g_strcmp0 (part_flags[i - 1].guid, guid) == 0 || g_strcmp0 (g_ascii_strdown (part_flags[i - 1].guid, -1), guid) == 0)
```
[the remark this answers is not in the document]
vojtechtrefny: Thanks, I didn't know this function exists.

src/plugins/part.c
```diff
@@ -582,6 +590,129 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
     return ret;
 }
 
+static BDPartFlag get_flag_from_guid (const gchar *guid) {
+    guint last_flag = 0;
+
+    last_flag = log2i (BD_PART_FLAG_BASIC_LAST);
+
+    for (guint i = 1; i < last_flag; i++) {
```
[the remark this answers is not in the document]
vojtechtrefny: The `part_flags` array is there to "translate" old parted-based flags to partition IDs and GUIDs (see the comment above the `PartFlag` definition). Libblockdev partition flags are based on the parted flags which are mix of flags, partition types (msdos) and guids (gpt). It's a big mess but this is the only way we can keep our code backwards compatible.

src/plugins/part.c
```diff
@@ -173,6 +173,14 @@ static const PartFlag part_flags[18] = {
 #define DEFAULT_PART_ID "0x83"
 #define DEFAULT_PART_GUID "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
 
+#define _PART_FAT12       0x01
```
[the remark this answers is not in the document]
vojtechtrefny: These are actually not partition codes, just "flags" to create partition ids from. I have added a comment with explanation.

src/plugins/part.c
```diff
@@ -593,48 +724,40 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
     gint part_num = 0;
     BDPartSpec *ret = NULL;
 
-    dev = ped_device_get (disk);
-    if (!dev) {
-        set_parted_error (error, BD_PART_ERROR_INVAL);
-        g_prefix_error (error, "Device '%s' invalid or not existing", disk);
+    part_num = get_part_num (part, error);
+    if (part_num == -1) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_INVAL,
```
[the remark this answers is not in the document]
vojtechtrefny: Fixed, thanks.

src/plugins/part.c
```diff
@@ -582,6 +590,129 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
+    if (!lb) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to read partition table.");
+        bd_part_spec_free (ret);
+        return NULL;
+    }
+
+    if (g_strcmp0 (fdisk_label_get_name (lb), "gpt") == 0 && fdisk_partition_has_partno (pa)) {
+        ret->type_guid = get_part_type_guid_and_gpt_flags (devname, fdisk_partition_get_partno (pa) + 1, &(ret->flags), error);
+        if (!ret->type_guid && *error) {
```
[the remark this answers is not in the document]
vojtechtrefny: We had this discussion in #447, we don't support calling functions with `error == NULL` right now.

src/plugins/part.c
```diff
@@ -582,6 +591,136 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
+            bd_part_spec_free (ret);
+            return NULL;
+        }
+
+        bd_flag = get_flag_from_guid (ret->type_guid);
+        if (bd_flag != BD_PART_FLAG_BASIC_LAST)
+            ret->flags = ret->flags | bd_flag;
```
vpodzime: I believe we usually write `ret->flags |= bd_flag` in this codebase.

src/plugins/part.c
```diff
@@ -582,6 +591,136 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
+
+    lb = fdisk_get_label (cxt, NULL);
+    if (!lb) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to read partition table.");
+        bd_part_spec_free (ret);
+        return NULL;
+    }
+
+    if (g_strcmp0 (fdisk_label_get_name (lb), "gpt") == 0 && fdisk_partition_has_partno (pa)) {
```
vpodzime: What's the reason for the `has_partno()` check?

vojtechtrefny: I'm using `get_partno` below. I could probably just use `ret->type == BD_PART_TYPE_NORMAL` because only "normal" partitions have GUIDs.

src/plugins/part.c

src/plugins/part.c
vpodzime: Both are equally cryptic without a comment. Maybe split this into two `if`s because otherwise this *either DOS or GPT-with-part-number* looks quite confusing, making me think something should be done for GPT partitions with no number (which I cannot really imagine).

vojtechtrefny: Ok, makes sense. I've added a separate `if` and a comment.

PR #447, opened 2019-06-17
Add function for (un)freezing filesystems

### LINKED ISSUE #445: fsfreeze
Hi, can the lib also provide an API call for freezing/thawing a filesystem?

Cheers

vojtechtrefny:
Fixes #445

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -465,6 +465,17 @@ gboolean bd_fs_mount (const gchar *device, const gchar *mountpoint, const gchar 
 gchar* bd_fs_get_mountpoint (const gchar *device, GError **error);
 
 /**
+ * bd_fs_is_mountpoint:
+ * @path: path (folder) to check
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether @path is a mountpoint or not
+ *
+ * Tech category: %BD_FS_TECH_MOUNT (no mode, ignored)
+ */
+gboolean bd_fs_is_mountpoint (const gchar *path, GError **error);
+
+/**
  * bd_fs_resize:
  * @device: the device the file system of which to resize
  * @new_size: new requested size for the file system (if 0, the file system is
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -410,6 +410,32 @@ gboolean bd_fs_clean (const gchar *device, GError **error);
 gchar* bd_fs_get_fstype (const gchar *device,  GError **error);
 
 /**
+ * bd_fs_freeze:
+ * @mountpoint: mountpoint of the device (filesystem) to freeze
+ * @error: (out): place to store error (if any)
+ *
+ * Freezes filesystem mounted on @mountpoint. The filesystem must
+ * support freezing.
+ *
+ * Returns: whether @mountpoint was successfully freezed or not
+ *
+ */
+gboolean bd_fs_freeze (const gchar *mountpoint, GError **error);
+
+/**
+ * bd_fs_unfreeze:
+ * @mountpoint: mountpoint of the device (filesystem) to un-freeze
+ * @error: (out): place to store error (if any)
+ *
+ * Un-freezes filesystem mounted on @mountpoint. The filesystem must
+ * support freezing.
+ *
+ * Returns: whether @mountpoint was successfully unfreezed or not
+ *
+ */
+gboolean bd_fs_unfreeze (const gchar *mountpoint, GError **error);
+
+/**
  * bd_fs_unmount:
  * @spec: mount point or device to unmount
  * @lazy: enable/disable lazy unmount
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -22,6 +22,8 @@
 #include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <linux/fs.h>
 #include <fcntl.h>
 
 #include <blockdev/utils.h>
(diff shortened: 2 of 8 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -409,6 +409,32 @@ gboolean bd_fs_clean (const gchar *device, GError **error);
...
+
+/**
+ * bd_fs_unfreeze:
+ * @device: the device to un-freeze (device must be mounted, trying to un-freeze
+ *          an unmounted device will result in an error)
+ * @error: (out): place to store error (if any)
+ *
+ * Un-freezes filesystem on @device. The filesystem must support un-freezing.
+ *
+ * Returns: whether @device was successfully freezed or not
```
vpodzime: unfreezed

src/lib/plugin_apis/fs.api
```diff
@@ -409,6 +409,32 @@ gboolean bd_fs_clean (const gchar *device, GError **error);
...
+ */
+gboolean bd_fs_freeze (const gchar *device, GError **error);
+
+/**
+ * bd_fs_unfreeze:
+ * @device: the device to un-freeze (device must be mounted, trying to un-freeze
+ *          an unmounted device will result in an error)
+ * @error: (out): place to store error (if any)
+ *
+ * Un-freezes filesystem on @device. The filesystem must support un-freezing.
```
vpodzime: I think it would make more sense to say that the FS needs to support freezing here. :smile:

src/lib/plugin_apis/fs.api
```diff
@@ -409,6 +409,32 @@ gboolean bd_fs_clean (const gchar *device, GError **error);
...
+ * @device: the device to freeze (device must be mounted, trying to freeze an
+ *          unmounted device will result in an error)
+ * @error: (out): place to store error (if any)
+ *
+ * Freezes filesystem on @device. The filesystem must support freezing.
+ *
+ * Returns: whether @device was successfully freezed or not
+ *
+ */
+gboolean bd_fs_freeze (const gchar *device, GError **error);
```
vpodzime: Any reason why this takes a device and not a mount point?

src/lib/plugin_apis/fs.api
vojtechtrefny: Fixed.

src/lib/plugin_apis/fs.api
vojtechtrefny: Fixed.

src/lib/plugin_apis/fs.api
vojtechtrefny: Because all our functions take devices and not mountpoints. Even `bd_fs_xfs_info` takes device. I can change this to mountpoint (it definitely makes more sense) but I thought it might be better to keep the API consistent.

vpodzime: IIRC, `bd_fs_xfs_info()` takes a device to be consistent with the other `fs_*_info()` functions. But that doesn't mean that we need to do it here too. Let's prefer simplicity here. :)

src/plugins/fs/generic.c
```diff
@@ -701,3 +703,75 @@ gboolean bd_fs_can_repair (const gchar *type, gchar **required_utility, GError *
...
+
+
+static gboolean fs_freeze (const char *device, gboolean freeze, GError **error) {
+    g_autofree gchar* mountpoint = NULL;
+    gint fd = -1;
+    gint status = 0;
+
+    mountpoint = bd_fs_get_mountpoint (device, error);
+    if (!mountpoint) {
+        if (*error != NULL) {
```
tbzatek: This leads to segfault when _error == NULL_.

src/plugins/fs/generic.c
```diff
@@ -701,3 +703,75 @@ gboolean bd_fs_can_repair (const gchar *type, gchar **required_utility, GError *
 gboolean bd_fs_can_set_label (const gchar *type, gchar **required_utility, GError **error) {
     return query_fs_operation (type, BD_FS_LABEL, required_utility, NULL, error);
 }
+
+
+static gboolean fs_freeze (const char *device, gboolean freeze, GError **error) {
+    g_autofree gchar* mountpoint = NULL;
```
tbzatek: Are you sure people would always use autoptr-capable compiler?

vpodzime: **YES**

src/lib/plugin_apis/fs.api
vojtechtrefny: Ok, I've changed the functions to take mountpoint. I've also added a new functions to check if path is a mountpoint.

src/plugins/fs/generic.c
tbzatek: We've had a little chat about that this morning. Anyway, looking at `glib/gmacros.h` defines, the `g_autofree` symbol is actually undefined when `__GNUC__` is undefined. So this way the compilation would fail, preventing any leaks. So no changes needed.

tbzatek:
> Looks good to me. I'm not sure how strict we should be about `error != NULL` in our functions. We don't say `allow-none` so people should just not pass `NULL`.

Yeah, we've talked about this this morning and it's true there are no annotations stating _NULL_ is allowed for `error`. However it's been a good practice in Glib-like projects for years. I guess it's optional whether the project adopts it. It's usually about adding a simple extra condition in the code.

However since the annotations are part of public API, the decision needs to be done at this very moment.

vpodzime:
> However since the annotations are part of public API, the decision needs to be done at this very moment.

A responsible caller should always be ready to check/pass further what went wrong.

tbzatek:
> A responsible caller should always be ready to check/pass further what went wrong.

Completely agree and it's the case most of the time. Passing NULL can be handy in case you don't really care about the error, such as: https://github.com/storaged-project/udisks/pull/577#issuecomment-480893867

Anyway, up to you to decide, don't want to block the review just because of this.

PR #449, opened 2019-06-21
Use libfdisk for functions for getting partitions

vojtechtrefny:
This covers following functions:

```
bd_part_get_part_by_pos
bd_part_get_disk_parts
bd_part_get_disk_free_regions
bd_part_get_best_free_region
```

There are two behaviour changes
- libfdisk ignores free spaces smaller than grain size (usually 1 MiB) and also rounds their start to nearest MiB -- I think this is ok as these are too small to actually create a partition
- parted has two extra partition types: *protected* (I have never seen this one) and *metadata* (2 sectors at the start of the disk and after the extended partition) -- the metadata partitions might be important for the `get_part_by_pos` because these are actually "unusable" parts of the disk and we might want to return a non-freespace "partition" for these. Do you think it would be a good idea to "manually" create part specs for the metadata "partitions"?

(First two commits are taken from #446)

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -902,51 +902,85 @@ BDPartDiskSpec* bd_part_get_disk_spec (const gchar *disk, GError **error) {
     return ret;
 }
 
-static BDPartSpec** get_disk_parts (const gchar *disk, guint64 incl, guint64 excl, gboolean incl_normal, GError **error) {
-    PedDevice *dev = NULL;
-    PedDisk *ped_disk = NULL;
-    PedPartition *ped_part = NULL;
-    guint num_parts = 0;
+static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean freespaces, GError **error) {
+    struct fdisk_context *cxt = NULL;
+    struct fdisk_table *table = NULL;
+    struct fdisk_partition *pa = NULL;
+    struct fdisk_iter *itr = NULL;
     BDPartSpec **ret = NULL;
+    BDPartSpec *spec = NULL;
     guint i = 0;
+    gint status = 0;
+    size_t num_parts = 0;
 
-    dev = ped_device_get (disk);
-    if (!dev) {
-        set_parted_error (error, BD_PART_ERROR_INVAL);
-        g_prefix_error (error, "Device '%s' invalid or not existing", disk);
+    cxt = get_device_context (disk, error);
+    if (!cxt) {
+        /* error is already populated */
         return NULL;
     }
 
-    ped_disk = ped_disk_new (dev);
-    if (!ped_disk) {
-        set_parted_error (error, BD_PART_ERROR_FAIL);
-        g_prefix_error (error, "Failed to read partition table on device '%s'", disk);
-        ped_device_destroy (dev);
+    table = fdisk_new_table ();
+    if (!table) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to create a new table");
+        close_context (cxt);
         return NULL;
     }
 
-    /* count the partitions we care about */
-    ped_part = ped_disk_next_partition (ped_disk, NULL);
-    while (ped_part) {
-        if (((ped_part->type & incl) && !(ped_part->type & excl)) ||
-            ((ped_part->type == 0) && incl_normal))
-            num_parts++;
-        ped_part = ped_disk_next_partition (ped_disk, ped_part);
+    itr = fdisk_new_iter (FDISK_ITER_FORWARD);
+    if (!itr) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to create a new iterator");
+        close_context (cxt);
+        return NULL;
+    }
+
+    if (parts) {
+        status = fdisk_get_partitions (cxt, &table);
+        if (status != 0) {
+            g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                         "Failed to get partitions: %s", strerror_l (-status, c_locale));
+            fdisk_free_iter (itr);
+            fdisk_unref_table (table);
+            close_context (cxt);
+            return NULL;
+        }
+    }
+
+    if (freespaces) {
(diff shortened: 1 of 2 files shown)


src/plugins/part.c
```diff
@@ -899,51 +899,93 @@ BDPartDiskSpec* bd_part_get_disk_spec (const gchar *disk, GError **error) {
...
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to create a new iterator");
+        close_context (cxt);
+        return NULL;
+    }
+
+    if (parts) {
+        status = fdisk_get_partitions (cxt, &table);
+        if (status != 0) {
+            g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
```
tbzatek: Just a hint: translate `status` to a real error message? The doc says _"a corresponding error"_, not very helpful.

src/plugins/part.c
```diff
@@ -899,51 +899,93 @@ BDPartDiskSpec* bd_part_get_disk_spec (const gchar *disk, GError **error) {
...
+            return NULL;
+        }
+    }
+
+    num_parts = fdisk_table_get_nents (table);
+    if (num_parts == 0) {
+        fdisk_free_iter (itr);
+        fdisk_unref_table (table);
+        close_context (cxt);
+        return NULL;
```
tbzatek: Hmm, zero partitions would still mean valid result, aren't you supposed to return empty list (i.e. just a single _NULL_ element)?

src/plugins/part.c
```diff
@@ -899,51 +899,93 @@ BDPartDiskSpec* bd_part_get_disk_spec (const gchar *disk, GError **error) {
...
     ret = g_new0 (BDPartSpec*, num_parts + 1);
     i = 0;
-    ped_part = ped_disk_next_partition (ped_disk, NULL);
-    while (ped_part) {
-        if (((ped_part->type & incl) && !(ped_part->type & excl)) ||
-            ((ped_part->type == 0) && incl_normal))
-            ret[i++] = get_part_spec (dev, ped_disk, ped_part, error);
-        ped_part = ped_disk_next_partition (ped_disk, ped_part);
+
+    while (fdisk_table_next_partition(table, itr, &pa) == 0) {
```
tbzatek: Missing space before opening parenthesis :-p

src/plugins/part.c
```diff
@@ -779,46 +870,28 @@ BDPartSpec* bd_part_get_part_spec (const gchar *disk, const gchar *part, GError
...
 
-    ped_disk = ped_disk_new (dev);
-    if (!ped_disk) {
-        set_parted_error (error, BD_PART_ERROR_FAIL);
-        g_prefix_error (error, "Failed to read partition table on device '%s'", disk);
-        ped_device_destroy (dev);
-        return NULL;
-    }
+    for (BDPartSpec **parts_p = parts; *parts_p; parts_p++) {
+        if ((*parts_p)->start <= position && ((*parts_p)->start + (*parts_p)->size) > position) {
```
tbzatek: I think the extra parenthesis may not be needed here... please check and look for compiler warnings.

src/plugins/part.c
```diff
@@ -899,51 +899,93 @@ BDPartDiskSpec* bd_part_get_disk_spec (const gchar *disk, GError **error) {
...
+                         "Failed to get partitions");
+            fdisk_free_iter (itr);
+            fdisk_unref_table (table);
+            close_context (cxt);
+            return NULL;
+        }
+    }
+
+    if (freespaces) {
+        status = fdisk_get_freespaces (cxt, &table);
```
vpodzime: Does this accumulate things in `&table` or override? I think we need to make sure `parts == TRUE && freespaces == TRUE` returns both types, right?

vpodzime: OK, maybe we don't. :)

tests/part_test.py
```diff
@@ -661,6 +662,13 @@ def test_get_disk_parts_empty(self):
         with self.assertRaises(GLib.GError):
             BlockDev.part_get_disk_parts (self.loop_dev)
 
+
+def _round_up_mib(size):
+    # convert size to nearest MiB (up)
+    rounded = Size(size).round_to_nearest(Size(1024**2), rounding=ROUND_UP)
+    return int(rounded.convert_to(B))
```
vpodzime: You can do `rounded.get_bytes()` or even just `int(rounded)`.

src/plugins/part.c
vpodzime: Same concern here.

src/plugins/part.c
```diff
@@ -779,46 +870,28 @@ BDPartSpec* bd_part_get_part_spec (const gchar *disk, const gchar *part, GError
...
-        if (set_parted_error (error, BD_PART_ERROR_FAIL))
-            g_prefix_error (error, "Failed to get partition at position %"G_GUINT64_FORMAT" (device '%s')",
-                            position, disk);
-        else
-            /* no such partition, but no error */
-            g_clear_error (error);
-        ped_disk_destroy (ped_disk);
-        ped_device_destroy (dev);
-        return NULL;
+            ret = *parts_p;
```
vpodzime: I think this should `break` or `!ret` should be added to the `for` loop condition above.

src/plugins/part.c
vojtechtrefny: I've added `strerror` here.

src/plugins/part.c

src/plugins/part.c
vojtechtrefny: I've added these as a reaction to gcc warning :-)

src/plugins/part.c
vojtechtrefny: The get functions just add partitions/freespaces to the table -- https://cdn.kernel.org/pub/linux/utils/util-linux/v2.26/libfdisk-docs/libfdisk-Table.html#fdisk-get-freespaces

src/plugins/part.c

src/plugins/part.c

tests/part_test.py
vojtechtrefny: Fixed.

PR #451, opened 2019-06-28
Hide some "private" functions

vojtechtrefny:
These shouldn't be publicly visible.

### DIFF
--- a/scripts/boilerplate_generator.py
+++ b/scripts/boilerplate_generator.py
@@ -191,16 +191,16 @@ def get_includes_str(includes):
     return ret
 
 def get_funcs_info(fn_infos, module_name):
-    ret = "static const gchar* const {0}_functions[] = {{\n".format(module_name)
+    ret = "const gchar* const {0}_functions[] = {{\n".format(module_name)
     for info in fn_infos:
         ret += '    "{0.name}",\n'.format(info)
     ret += "    NULL};\n\n"
 
-    ret += ("const gchar* const* get_{0}_functions (void) {{\n".format(module_name) +
+    ret += ("static const gchar* const* get_{0}_functions (void) {{\n".format(module_name) +
             "    return {0}_functions;\n".format(module_name) +
             "}\n\n")
 
-    ret += "static const guint8 {0}_num_functions = {1};\n\n".format(module_name, len(fn_infos))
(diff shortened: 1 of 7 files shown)


src/plugins/fs/ext.c
```diff
@@ -148,7 +148,8 @@ static gboolean extract_e2fsck_progress (const gchar *line, guint8 *completion)
  * Returns: whether the @tech-@mode combination is available -- supported by the
  *          plugin implementation and having all the runtime dependencies available
  */
-gboolean bd_fs_ext_is_tech_avail (BDFSTech tech UNUSED, guint64 mode, GError **error) {
+gboolean __attribute__ ((visibility ("hidden")))
```
tbzatek: You may also use `G_GNUC_INTERNAL` for that. According to docs, you'll get support to _Sun Studio compiler_ ;-)

Also the glib docs recommend to move the macro at the very beginning of the declaration.

scripts/boilerplate_generator.py
```diff
@@ -196,12 +196,12 @@ def get_funcs_info(fn_infos, module_name):
...
-    ret += ("const gchar* const* get_{0}_functions (void) {{\n".format(module_name) +
+    ret += ("static const gchar* const* get_{0}_functions (void) {{\n".format(module_name) +
             "    return {0}_functions;\n".format(module_name) +
             "}\n\n")
 
     ret += "static const guint8 {0}_num_functions = {1};\n\n".format(module_name, len(fn_infos))
-    ret += ("guint8 get_{0}_num_functions (void) {{\n".format(module_name) +
+    ret += ("static guint8 get_{0}_num_functions (void) {{\n".format(module_name) +
             "    return {0}_num_functions;\n".format(module_name) +
             "}\n\n")
```
vpodzime: These two should probably be removed, we are not using them anywhere, AFAICT.

vojtechtrefny: Ok, removed.

src/plugins/fs/ext.c
vojtechtrefny: > According to docs, you'll get support for *Sun Studio compiler* ;-)

Wow, we definitely need that :-)

> Also the glib docs recommend to move the macro at the very beginning of the declaration.

I'd like to keep this consistent with other "hidden" functions where we use the macro like this.

tbzatek: Fair enough :-)

PR #452, opened 2019-07-01
F2FS support

vojtechtrefny:


src/plugins/fs/f2fs.c
```diff
@@ -0,0 +1,349 @@
...
+    gint status = 0;
+    gboolean ret = FALSE;
+
+    if (!check_deps (&avail_deps, DEPS_FSCKF2FS_MASK, deps, DEPS_LAST, &deps_check_lock, error))
+        return FALSE;
+
+    ret = bd_utils_exec_and_report_status_error (args, extra, &status, error);
+    if (!ret && (status == 255)) {
+        /* no error should be reported for exit code 255 -- there are errors on the filesystem */
+        g_clear_error (error);
```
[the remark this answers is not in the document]
vojtechtrefny: Yes, that is what all fs check functions do -- `FALSE` without means the filesystem is corrupted, `FALSE` with error means we were unable to run the check command (or it failed with an unexpected return code).

vojtechtrefny:
The tests are still failing on Debian 10 because `dump.f2fs` output contains null byte characters and it breaks parsing.

vojtechtrefny:
Rebased to current master and added new commit that changes how `bd_utils_exec_and_capture_output` runs the command so we can parse its output even if it has null bytes in it.

src/plugins/fs/common.c
```diff
@@ -227,3 +227,81 @@ wipe_fs (const gchar *device, const gchar *fs_type, gboolean wipe_all, GError **
...
+            return FALSE;
+        }
+
+        *label = g_strdup (value);
+    }
+
+    status = blkid_probe_lookup_value (probe, "UUID", &value, NULL);
+    if (status != 0) {
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
+                     "Failed to get label for the device '%s'", device);
```
vpodzime: *Failed to get UUID...*

src/plugins/fs/common.c
```diff
@@ -227,3 +227,81 @@ wipe_fs (const gchar *device, const gchar *fs_type, gboolean wipe_all, GError **
...
+                         "Failed to get label for the device '%s'", device);
+            blkid_free_probe (probe);
+            synced_close (fd);
+            return FALSE;
+        }
+
+        *label = g_strdup (value);
+    }
+
+    status = blkid_probe_lookup_value (probe, "UUID", &value, NULL);
```
vpodzime: No `has_value()` check needed here?

src/plugins/fs/common.c
```diff
@@ -227,3 +227,81 @@ wipe_fs (const gchar *device, const gchar *fs_type, gboolean wipe_all, GError **
...
+    if (status != 0) {
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
+                     "Failed to get label for the device '%s'", device);
+        blkid_free_probe (probe);
+        synced_close (fd);
+        g_free (label);
+        return FALSE;
+    }
+
+    *uuid = g_strdup (value);
```
vpodzime: What if `value == NULL`? Should `*uuid` end up being `""`?

src/lib/plugin_apis/fs.api
```diff
@@ -334,6 +334,86 @@ GType bd_fs_ntfs_info_get_type () {
...
+#define BD_FS_TYPE_F2FS_INFO (bd_fs_f2fs_info_get_type ())
+GType bd_fs_f2fs_info_get_type();
+
+/**
+ * BDFSF2FSInfo:
+ * @label: label of the filesystem
+ * @uuid: uuid of the filesystem
+ * @cluster_size: cluster size used by the filesystem
+ * @cluster_count: number of clusters in the filesystem
+ * @free_cluster_count: number of free clusters in the filesystem
```
vpodzime: `free_cluster_count` described here, but `features` is part of the struct instead.

src/lib/plugin_apis/fs.api
```diff
@@ -1188,4 +1269,84 @@ gboolean bd_fs_ntfs_resize (const gchar *device, guint64 new_size, GError **erro
...
+ * @device: the device containing the file system to check
+ * @extra: (allow-none) (array zero-terminated=1): extra options for the repair (right now
+ *                                                 passed to the 'fsck.f2fs' utility)
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether an f2fs file system on the @device is clean or not
+ *
+ * Tech category: %BD_FS_TECH_F2FS-%BD_FS_TECH_MODE_CHECK
+ */
+gboolean bd_fs_f2fs_check (const gchar *device, const BDExtraArg **extra, GError **error);
```
vpodzime: Why do we want to support `**extra` for `f2fs_check()` and `f2fs_repair()`?

src/lib/plugin_apis/fs.api
```diff
@@ -1188,4 +1269,84 @@ gboolean bd_fs_ntfs_resize (const gchar *device, guint64 new_size, GError **erro
...
+ * Tech category: %BD_FS_TECH_F2FS-%BD_FS_TECH_MODE_QUERY
+ */
+BDFSF2FSInfo* bd_fs_f2fs_get_info (const gchar *device, GError **error);
+
+/**
+ * bd_fs_f2fs_resize:
+ * @device: the device containing the file system to resize
+ * @new_size: new requested size for the file system *in file system sectors* (see bd_fs_f2fs_get_info())
+ *            (if 0, the file system is adapted to the underlying block device)
+ * @safe: whether to perform safe resize or not (does not resize metadata)
```
vpodzime: Why would somebody want to do an unsafe resize?

src/plugins/fs/f2fs.c
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    for (i = 0; i <= BD_FS_MODE_LAST; i++)
+        if (mode & (1 << i))
+            required |= fs_mode_util[i];
+
+    return check_deps (&avail_deps, required, deps, DEPS_LAST, &deps_check_lock, error);
+}
+
+/**BD_FS_ERROR_INVAL
```
vpodzime: What's this `BD_FS_ERROR_INVAL` doing here? A copy pasta?

src/lib/plugin_apis/fs.api
vpodzime: So I see it is required for shrinking. I think it should just happen automatically. Or is there a use case where it would make sense to not use it when needed?

src/utils/exec.c
```diff
@@ -323,53 +334,103 @@ gboolean bd_utils_exec_and_capture_output (const gchar **argv, const BDExtraArg
...
-        if (g_error_matches (*error, G_SPAWN_ERROR, G_SPAWN_ERROR_FAILED)) {
-            /* process was terminated abnormally (e.g. using a signal) */
-            g_free (stdout_data);
-            g_free (stderr_data);
+    out_pipe = g_io_channel_unix_new (out_fd);
+    err_pipe = g_io_channel_unix_new (err_fd);
+
+    do {
+        wpid = waitpid (pid, &wstatus, 0);
+         if (wpid == -1) {
```
vpodzime: Weird alignment here.

src/utils/exec.c
```diff
@@ -287,16 +288,26 @@ gboolean bd_utils_exec_and_report_status_error (const gchar **argv, const BDExtr
  */
 gboolean bd_utils_exec_and_capture_output (const gchar **argv, const BDExtraArg **extra, gchar **output, GError **error) {
```
vpodzime: I wonder if we could just call `bd_utils_exec_and_report_progress()` with progress reporting arguments being `NULL`. Because re-implementing `g_spawn_sync()` properly would be quite a bit harder -- the child process may not terminate if the pipe buffers are not big enough for its output and the parent process is not reading from the pipes.

src/plugins/fs/common.c
vojtechtrefny: Fixed.

src/plugins/fs/common.c
vojtechtrefny: I guess it's not possible to have filesystem without UUID, but I've added the check just to be sure.

src/plugins/fs/common.c
vojtechtrefny: Again, I don't think it's possible to have empty UUID, but we can add a separate check here just in case.

src/lib/plugin_apis/fs.api
vojtechtrefny: Fixed., the `cluster` attributes were also wrong.

src/lib/plugin_apis/fs.api
vojtechtrefny: We already have extra arguments for these for other filesystems like vfat and ext and it is possible there will be more arguments for `fsck.f2fs` in the future (they are still adding new features).

src/plugins/fs/f2fs.c
vojtechtrefny: Fixed.

src/utils/exec.c
vojtechtrefny: Fixed.

src/lib/plugin_apis/fs.api
vojtechtrefny: I think better question is if there is a use case for "safe" resize when it's not needed (for growing). And the answer is "I don't know". I think the best solution is to give user full control here, the generic resize function is adding the safe option for shrinking which should be ok for most users IMHO.

vpodzime: Works for me.

src/utils/exec.c
vojtechtrefny: Fixed, `bd_utils_exec_and_capture_output` now shares code with `bd_utils_exec_and_report_progress`.

vpodzime:
Jenkins, test this, please.

PR #455, opened 2019-07-12
libfdisk: create part

vojtechtrefny:
Support for creating partitions using libfdisk. This is currently broken because of bug in udev (maybe) -- when creating partition table and a partition directly after that the add event for the partition isn't created.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,9 +226,8 @@ AS_IF([test "x$with_dm" != "xno" -a "x$with_dmraid" != "xno"],
       [LIBBLOCKDEV_CHECK_HEADER([dmraid/dmraid.h], [], [dmraid.h not available])],
       [])
 
-AS_IF([test "x$with_part" != "xno" -o "x$with_fs" != "xno"],
-      [LIBBLOCKDEV_PKG_CHECK_MODULES([PARTED], [libparted >= 3.1])
-      LIBBLOCKDEV_PKG_CHECK_MODULES([FDISK], [fdisk >= 2.31.0])]
+AS_IF([test "x$with_part" != "xno"],
+      [LIBBLOCKDEV_PKG_CHECK_MODULES([FDISK], [fdisk >= 2.31.0])],
       [])
 
 AS_IF([test "x$with_fs" != "xno"],
@@ -236,6 +235,8 @@ AS_IF([test "x$with_fs" != "xno"],
        # new versions of libmount has some new functions we can use
        AS_IF([$PKG_CONFIG --atleast-version=2.30.0 mount],
              [AC_DEFINE([LIBMOUNT_NEW_ERR_API])], [])
+
+       LIBBLOCKDEV_PKG_CHECK_MODULES([PARTED], [libparted >= 3.1])
 
        # older versions of parted don't provide the libparted-fs-resize.pc file
        AS_IF([$PKG_CONFIG libparted-fs-resize],
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -18,7 +18,6 @@
  */
 
 #include <string.h>
-#include <parted/parted.h>
 #include <ctype.h>
 #include <stdlib.h>
 #include <math.h>
@@ -29,7 +28,6 @@
 #include <sys/ioctl.h>
 #include <linux/fs.h>
 #include <blockdev/utils.h>
-#include <part_err.h>
 #include <libfdisk.h>
 #include <locale.h>
 #include <blkid.h>
@@ -328,31 +326,6 @@ static gboolean write_label (struct fdisk_context *cxt, struct fdisk_table *orig
 }
 
 /**
- * set_parted_error: (skip)
- *
- * Set error from the parted error stored in 'error_msg'. In case there is none,
- * the error is set up with an empty string. Otherwise it is set up with the
- * parted's error message and is a subject to later g_prefix_error() call.
- *
- * Returns: whether there was some message from parted or not
- */
-static gboolean set_parted_error (GError **error, BDPartError type) {
-    gchar *error_msg = NULL;
-    error_msg = bd_get_error_msg ();
-    if (error_msg) {
-        g_set_error (error, BD_PART_ERROR, type,
-                     " (%s)", error_msg);
-        g_free (error_msg);
-        error_msg = NULL;
-        return TRUE;
-    } else {
-        g_set_error_literal (error, BD_PART_ERROR, type, "");
-        return FALSE;
-    }
-}
-
-
-/**
  * bd_part_check_deps:
  *
  * Returns: whether the plugin's runtime dependencies are satisfied or not
@@ -373,7 +346,6 @@ gboolean bd_part_check_deps (void) {
  *
  */
 gboolean bd_part_init (void) {
-    ped_exception_set_handler ((PedExceptionHandler*) bd_exc_handler);
     c_locale = newlocale (LC_ALL_MASK, "C", c_locale);
     fdisk_init_debug (0);
     return TRUE;
@@ -387,7 +359,6 @@ gboolean bd_part_init (void) {
  *
  */
 void bd_part_close (void) {
-    ped_exception_set_handler (NULL);
     c_locale = (locale_t) 0;
 }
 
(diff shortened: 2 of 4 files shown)


pothos:
Can you say something about the differences between libparted and libfdisk? I'm a bit worried if libfdisk is as feature complete and will cause any regressions. The interaction of partitioning tool, the kernel, and udev is tricky. For example, does it use the blkpg ioctl to modify the partion table state in the kernel?

vojtechtrefny:
> Can you say something about the differences between libparted and libfdisk? I'm a bit worried if libfdisk is as feature complete and will cause any regressions

Libfdisk currently supports everything we need, we were even able to get rid of dependecy on sfdisk and sgdisk tools -- we are currently using these for partition ID and GUID management which is not fully supported by libparted. The GPT support in libparted always looked a little bit "hacky" to me (especially the parted flags) and this is much better in libfdisk.
I think our test suite has a quite good coverage and it should catch all possible regressions. We will also of course test this with UDisks test suite before merging the libfdisk branch (the rewrite is done on a separate branch).
Generally, (lib)fdisk is not a new piece of software and it's widely used, so I don't expect major problems.

> The interaction of partitioning tool, the kernel, and udev is tricky.

Yes, I know. We already have some crazy workarounds for udev-parted interactions (for example https://github.com/storaged-project/libblockdev/blob/master/src/plugins/part.c#L242). The good thing about libfdisk is that it has an active maintainer so I expect problems like these will be fixed in libfdisk.

vojtechtrefny:
> This is currently broken because of bug in udev (maybe) -- when creating partition table and a partition directly after that the add event for the partition isn't created.

So this was actually problem in my code, we need to run `fdisk_reread_changes` or `fdisk_reread_partition_table` after adding or removing  (or resizing) partitions.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

src/plugins/part.c
```diff
@@ -279,16 +279,52 @@ static void close_context (struct fdisk_context *cxt) {
...
-    ret = fdisk_write_disklabel (cxt);
+    gint dev_fd = 0;
+    guint num_tries = 1;
+
+    /* XXX: try to grab a lock for the device so that udev doesn't step in
+       between the two operations we need to perform (see below) with its
+       BLKRRPART ioctl() call which makes the device busy */
+    dev_fd = open (disk, O_RDONLY|O_CLOEXEC);
+    if (dev_fd >= 0) {
+        ret = flock (dev_fd, LOCK_SH|LOCK_NB);
```
tbzatek: So where do we release the lock then?

src/plugins/part.c
```diff
@@ -1726,6 +1726,78 @@ gboolean bd_part_delete_part (const gchar *disk, const gchar *part, GError **err
...
+                    *max_size  = fdisk_partition_get_size (cur) + fdisk_partition_get_size (pa);
+                else
+                    *max_size = fdisk_partition_get_size (pa) - (start - fdisk_partition_get_start (pa));
+                break;
+            }
+        }
+	  }
+
+    fdisk_free_iter (itr);
+	  return TRUE;
```
tbzatek: Weird code style alignment - spaces vs. tabs or just Github rendering issue?

src/plugins/part.c
vojtechtrefny: Closing the file descriptors also removes the lock.

vojtechtrefny:
> Haven't checked the logic in detail though.

Yes, the logic is quite complicated. I plan to go through this one more time before doing a "big" PR for merging this to master and check if the test coverage is good enough and also run UDisks test suite with this. Hopefully this will be enough to be sure the behaviour didn't change.

src/plugins/part.c
```diff
@@ -1317,107 +1254,354 @@ static PedPartition* add_part_to_disk (PedDevice *dev, PedDisk *disk, BDPartType
...
     msg = g_strdup_printf ("Started adding partition to '%s'", disk);
     progress_id = bd_utils_report_started (msg);
     g_free (msg);
 
-    dev = ped_device_get (disk);
-    if (!dev) {
-        set_parted_error (error, BD_PART_ERROR_INVAL);
-        g_prefix_error (error, "Device '%s' invalid or not existing", disk);
+    cxt = get_device_context (disk, error);
+        if (!cxt) {
```
vpodzime: Bad alignment here as well.

src/plugins/part.c
```diff
@@ -1317,107 +1254,354 @@ static PedPartition* add_part_to_disk (PedDevice *dev, PedDisk *disk, BDPartType
...
+        close_context (cxt);
+        return NULL;
+    }
 
-    /* the partition gets destroyed together with the disk*/
-    ped_disk_destroy (ped_disk);
-    ped_device_destroy (dev);
+    /* the in-memory model of the new partition is not updated, we need to
+       read the spec manually */
+    if (fdisk_partition_has_partno (npa)) {
```
vpodzime: So what is reading the spec here?

src/plugins/part.c
```diff
@@ -1726,6 +1726,78 @@ gboolean bd_part_delete_part (const gchar *disk, const gchar *part, GError **err
...
+    if (!cur) {
+        g_set_error (error, BD_PART_ERROR, BD_PART_ERROR_FAIL,
+                     "Failed to locate partition '%d' in table.", partno);
+        fdisk_free_iter (itr);
+        return FALSE;
+    }
+
+    start = fdisk_partition_get_start (cur);
+
+    while (fdisk_table_next_partition (tb, itr, &pa) == 0) {
```
vpodzime: I'd prefer `!found && fdisk_table_next_partition())` here and getting rid of the `break`s below.

src/plugins/part.c
```diff
@@ -1741,111 +1813,171 @@ gboolean bd_part_delete_part (const gchar *disk, const gchar *part, GError **err
...
+            fdisk_unref_partition (pa);
+            close_context (cxt);
+            bd_utils_report_finished (progress_id, (*error)->message);
+            return FALSE;
+        }
 
-            g_usleep (step_us);
-            wait_us -= step_us;
+        if (size == 0) {
+            g_info ("Not resizing, partition '%s' is already at its maximum size.", part);
```
vpodzime: Please use our logging.

tests/skip.yml
```diff
@@ -131,3 +131,9 @@
   skip_on:
     - distro: "debian"
       reason: "libparted provides weird values here"
+
+- test: part_test.PartCreateResizePartCase.test_create_resize_part_single
+  skip_on:
+    - distro: "fedora"
+      version: "29"
+      reason: "libfdisk calculation of max size for resizing is broken on <= 2.32"
```
vpodzime: Isn't this handled in the code?

configure.ac
```diff
@@ -235,6 +234,8 @@ AS_IF([test "x$with_fs" != "xno"],
        AS_IF([$PKG_CONFIG --atleast-version=2.30.0 mount],
              [AC_DEFINE([LIBMOUNT_NEW_ERR_API])], [])
 
+       LIBBLOCKDEV_PKG_CHECK_MODULES([PARTED], [libparted >= 3.1]),
```
vpodzime: Now I wonder how we get rid of this. ;)

src/plugins/part.c
vojtechtrefny: I've moved the comment to the line where we get the spec.

src/plugins/part.c
vojtechtrefny: Ok, you're the boss here :-)

configure.ac
vojtechtrefny: Slowly :-)

src/plugins/part.c
```diff
@@ -278,16 +278,52 @@ static void close_context (struct fdisk_context *cxt) {
...
-static gboolean write_label (struct fdisk_context *cxt, const gchar *disk, GError **error) {
+static gboolean write_label (struct fdisk_context *cxt, struct fdisk_table *orig, const gchar *disk, GError **error) {
     gint ret = 0;
-    ret = fdisk_write_disklabel (cxt);
+    gint dev_fd = 0;
+    guint num_tries = 1;
+
+    /* XXX: try to grab a lock for the device so that udev doesn't step in
+       between the two operations we need to perform (see below) with its
+       BLKRRPART ioctl() call which makes the device busy */
```
marcosfrm: Link this:
https://systemd.io/BLOCK_DEVICE_LOCKING/

marcosfrm: Completely missed the code was using LOCK_SH. Thanks for fixing it.

vojtechtrefny:
I've added two more commits:
- replacing the shared lock `LOCK_SH` with an exlusive one `LOCK_EX` based on @marcosfrm comment
- temporary workaround for adding logical partitions

These changes together with #504 are enough to run UDisks tests (with a small change, see https://github.com/storaged-project/udisks/pull/736) successfully. I have also tested the changes manually with GNOME Disks and everything seems to be working.

src/plugins/part.c
vojtechtrefny: The `LOCK_SH` is copy-pasted from our existing libparted based implementation. Not sure why we are not using the `LOCK_EX`, everything seems to be working correctly with it.

vpodzime:
>     * 
> 
> 
> These changes together with #504 are enough to run UDisks tests (with a small change, see [storaged-project/udisks#736](https://github.com/storaged-project/udisks/pull/736)) successfully. I have also tested the changes manually with GNOME Disks and everything seems to be working.

:tada: :tada: :tada:

vojtechtrefny:
One last change: The problem with `fdisk_reread_changes` and extended partitions is a bug in libfdisk and will be fixed in next release. I've added a workaround for older versions.
I consider this ready to be merged, if you want to do a new review, this is a good time. (I've also rebased both fdisk branches to latest master.)

vojtechtrefny:
The failed tests on rawhide and debian testing are most likely caused by a bug in libfdisk introduced in 2.36, see https://github.com/karelzak/util-linux/commit/4fe7f9b614e2b5bb97f6d89af02acb867cffccc1

tbzatek:
> I consider this ready to be merged, if you want to do a new review, this is a good time. (I've also rebased both fdisk branches to latest master.)

I think this is good to go, any new issues can be dealt with later.

pothos:
I've tested it and overall it looked good, but the behavior for GNOME Disks changed slightly. With the old implementation partition changes were more smooth. E.g., adding an extended partition on the loop device takes longer. Maybe this is related to locking or and whether/how the direct syscalls to update the kernel partition table are used? Due to it taking longer adding the extended partition sometimes fails now with the error message that it failed to notify the kernel (then the partition is there but not shown  and a kernel reread needs to be triggered, e.g., reattach the loop device), but it's hard to reproduce… In another case adding a logical partition failed similarly and ended up being created on disk but not known to the kernel (not sure if I why).

I also found out that resizing now fails when the requested size is too large which was previously just silently using the maximum size.

In summary, I guess there are some corner cases to address as expected but I see the benefits of the transition and don't want to hold it back, thanks for your work.

vojtechtrefny:
> I've tested it and overall it looked good, but the behavior for GNOME Disks changed slightly. With the old implementation partition changes were more smooth. E.g., adding an extended partition on the loop device takes longer. Maybe this is related to locking and whether/how the direct syscalls to update the kernel partition table are used? Due to it taking longer adding the extended partition sometimes fails now with the error message that it failed to notify the kernel (then the partition is there but not shown and a kernel reread needs to be triggered, e.g., reattach the loop device), but it's hard to reproduce… In another case adding a logical partition failed similarly and ended up being created on disk but not known to the kernel (not sure if I why).

This should be related to the last commit that has a workaround for a bug in libfdisk. There are to ways to tell kernel to read the partition table changes -- we can either use `fdisk_reread_changes` which tells kernel to add (or remove) only partitions that were changed, or use `fdisk_reread_partition_table` to tell kernel to reread whole partition table. Unfortunately `fdisk_reread_changes` is currently broken with extended partitions (will be fixed in next libfdisk release) and `fdisk_reread_partition_table` (or the `BLKRRPART` ioctl in general) will fail if the device is currently in use. I'll definitely do some stress testing and maybe change the timeout for the flock locking, but I assume the libfdisk will be generally available before we release this. (This change will be part of the 3.0 release which will change both API and ABI so I don't expect it to land on systems with older libfdisk. And the 3.0 release will take at least few more months.)

> 
> I also found out that resizing now fails when the requested size is too large which was previously just silently using the maximum size.

We should definitely keep this behaviour. The libparted implementation has some soft limits and adjusts the size in similar cases so we should do the same here. I've created issue for this #557

This is not the final merge, we still have time for fixing similar issues. I plan to run UDisks tests and do some manual testing before the final big merge to master.

Thank you for testing this.

PR #460, opened 2019-07-26
Move the NTFS read-only device test to a separate test case

vojtechtrefny:
And also skip this test case on Debian. On Debian mounting a read
only NTFS device succeeds even if 'rw' option is specified.

On the other hand on Fedora mounting without the 'rw' option fails
instead of mounting the device as read-only. NTFS mounting is just
weird but the test is now written in a way which expects the same
behaviour other filesystems have.

This is not a bug in libblockdev, running mount manually shows the
same inconsistent behaviour.

### DIFF
--- a/tests/fs_test.py
+++ b/tests/fs_test.py
@@ -1199,18 +1199,34 @@ class MountTest(FSTestCase):
         self.assertTrue(succ)
         self.assertFalse(os.path.ismount(tmp))
 
+    def test_mount_ntfs_ro(self):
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
> Perhaps a difference between ntfs-3g vs. the legacy kernel ntfs driver? Or different ntfs-3g versions?

Same version of ntfs-3g and I was testing this with `-t ntfs-3g`.

PR #463, opened 2019-07-31
Fix LVM_MAX_LV_SIZE in the GIR file

### LINKED ISSUE #462: In /usr/share/gir-1.0/BlockDev-2.0.gir LVM_MAX_LV_SIZE is negative
```
    <constant name="LVM_MAX_LV_SIZE"
              value="-9223372036854775808"
              c:type="BD_LVM_MAX_LV_SIZE">
      <type name="gint" c:type="gint"/>
    </constant>
```

This is because the value of 8 EiB is one larger than the largest positive value representable in a 64-bit integer, probably.

```
> dnf info libblockdev-devel
Installed Packages
Name         : libblockdev-devel
Version      : 2.22
Release      : 1.fc29
Architecture : x86_64
Size         : 2.0 M
Source       : libblockdev-2.22-1.fc29.src.rpm
Repository   : @System
From repo    : updates
Summary      : Development files for libblockdev
URL          : https://github.com/storaged-project/libblockdev
License      : LGPLv2+
Description  : This package contains header files and pkg-config files needed
             : for development with the libblockdev library.
```

vojtechtrefny:
GObject instrospection doesn't automatically detect types of the
constants and thinks that all are gint, which doesn't work for
LVM_MAX_LV_SIZE because it's bigger than gint64 and should be
guint64.
Unfortunately there is not way to make the constant in GIR arch
dependent, so we are deprecating it, users shoud use the
bd_lvm_get_max_lv_size function instead.

Fixes #462

### DIFF
--- a/src/lib/plugin_apis/lvm.api
+++ b/src/lib/plugin_apis/lvm.api
@@ -5,13 +5,15 @@
 #ifndef BD_LVM_API
 #define BD_LVM_API
 
-#ifdef __LP64__
-// 64bit system
-#define BD_LVM_MAX_LV_SIZE (8 EiB)
-#else
-// 32bit system
-#define BD_LVM_MAX_LV_SIZE (16 TiB)
-#endif
+/**
+ * BD_LVM_MAX_LV_SIZE:
+ *
+ * Deprecated: 2.23: Use bd_lvm_get_max_lv_size() function instead.
+ */
+/* 8 EiB (valid only for 64bit devices but we can't have architecture
+   specific constant here) */
+#define BD_LVM_MAX_LV_SIZE G_GUINT64_CONSTANT (9223372036854775808ULL)
+
 
 #define BD_LVM_DEFAULT_PE_START (1 MiB)
 #define BD_LVM_DEFAULT_PE_SIZE (4 MiB)


src/lib/plugin_apis/lvm.api
```diff
@@ -5,13 +5,13 @@
...
-// 64bit system
-#define BD_LVM_MAX_LV_SIZE (8 EiB)
-#else
-// 32bit system
-#define BD_LVM_MAX_LV_SIZE (16 TiB)
-#endif
+/**
+ * BD_LVM_MAX_LV_SIZE:
+ *
+ * Deprecated: 2.23: Use bd_lvm_get_max_lv_size function instead.
```
vpodzime: I think that if you put `()` after the name of the function, it will work as a link in the documentation.

src/lib/plugin_apis/lvm.api
```diff
@@ -5,13 +5,13 @@
...
+ */
+#define BD_LVM_MAX_LV_SIZE G_GUINT64_CONSTANT (9223372036854775808ULL)
```
vpodzime: Would still be nice to see the `8 EiB` value in a comment here.

vojtechtrefny: Comment added.

PR #466, opened 2019-09-05
Fix how we get process exit code from g_spawn_sync

vojtechtrefny:
exit_status set by g_spawn_sync is the wait status from waitpid,
not the process exit code. We need to use g_spawn_check_exit_status
to get that.

Taken from #452. The F2FS support needs some changes, but we should merge this fix as soon as possible because it affects all `bd_utils_exec_` functions.

### DIFF
--- a/src/utils/exec.c
+++ b/src/utils/exec.c
@@ -191,6 +191,7 @@ gboolean bd_utils_exec_and_report_status_error (const gchar **argv, const BDExtr
     guint args_len = 0;
     const gchar **arg_p = NULL;
     const BDExtraArg **extra_p = NULL;
+    gint exit_status = 0;
     guint i = 0;
 
     if (extra) {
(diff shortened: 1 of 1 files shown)


src/utils/exec.c
```diff
@@ -304,15 +325,33 @@ gboolean bd_utils_exec_and_capture_output (const gchar **argv, const BDExtraArg
...
     success = g_spawn_sync (NULL, (gchar **) argv, NULL, G_SPAWN_SEARCH_PATH,
                             (GSpawnChildSetupFunc) set_c_locale, NULL,
-                            &stdout_data, &stderr_data, &status, error);
-    log_out (task_id, stdout_data, stderr_data);
-    log_done (task_id, status);
+                            &stdout_data, &stderr_data, &exit_status, error);
 
     if (!success) {
         /* error is already populated */
         return FALSE;
```
vpodzime: Should we also add `g_free(stdout_data)` and the same for stderr here?

vojtechtrefny: Good catch, fixed.

PR #470, opened 2019-09-16
Logging changes

### LINKED ISSUE #401: Log messages from libdevmapper
In all the plugins that use the library we should setup the logging handler to pass the messages to `bd_utils_log()`.

### LINKED ISSUE #422: Use libblockdev log functions instead of g_warning/g_debug/...
We are currently "logging" everything to stdout/stderr using `g_warning` (and similar functions). We should use `bd_utils_log` instead.

vojtechtrefny:
Multiple changes to our logging functions including custom log levels and replacing `g_warning` calls with our logging functions.

Fixes #422 
Fixes #401


src/utils/logging.c
```diff
@@ -18,6 +18,7 @@
  */
 
 #include <glib.h>
+#include <syslog.h>
```
vpodzime: Why is this suddenly needed?

src/utils/logging.c
```diff
@@ -50,3 +52,27 @@ void bd_utils_log (gint level, const gchar *msg) {
...
+    if (log_func) {
+        va_start (args, format);
+        ret = g_vasprintf (&msg, format, args);
+        va_end (args);
+
+        if (ret < 0)
+            return;
+
+        log_func (level, msg);
+    }
```
vpodzime: Can we do `else if (level >= WARNING) g_warning(format, args)` here so that if the user code doesn't have the logging function set, they would still get the important messages somehow?

src/utils/logging.c
vojtechtrefny: Removed.

src/utils/logging.c
vojtechtrefny: It's library, in most cases user won't see stdout anyway.

I was thinking about adding a logging function that logs using g_warning that users can use if they don't want to write custom logging function. Maybe we could add this and make it default logging function instead of having none?

vpodzime: > It's library, in most cases user won't see stdout anyway.

Well, stdout from daemons started as systemd services by default goes to journal, doesn't it?

vpodzime: > I was thinking about adding a logging function that logs using g_warning that users can use if they don't want to write custom logging function. Maybe we could add this and make it default logging function instead of having none?

Sounds like a plan to me!

vojtechtrefny: I've added a new functions that logs using g_warning/info/etc. I'm still not sure if using this function as default is a good thing.

src/utils/logging.c
```diff
@@ -37,7 +37,10 @@ gboolean bd_utils_init_logging (BDUtilsLogFunc new_log_func, GError **error __at
     /* XXX: the error attribute will likely be used in the future when this
        function gets more complicated */
 
-    log_func = new_log_func;
+    if (new_log_func == NULL)
+        log_func = &bd_utils_log_stdout;
```
vpodzime: I think this should just assign `NULL` enabling a nice way to disable logging altogether.

src/utils/logging.c
```diff
@@ -75,3 +78,34 @@ void bd_utils_log_format (gint level, const gchar *format, ...) {
...
+    switch (level) {
+        case BD_UTILS_LOG_DEBUG:
+#ifdef DEBUG
+            g_debug ("%s", msg);
+#endif
+            break;
+        case BD_UTILS_LOG_INFO:
+        case BD_UTILS_LOG_NOTICE:
+            g_info ("%s", msg);
+            break;
```
vpodzime: This will produce a lot of noise from running external utilities and doing DBus method calls, right? It should also be guarded with the `#ifdef DEBUG` I think.

src/utils/logging.c
vojtechtrefny: Ok, fixed.

src/utils/logging.c
vojtechtrefny: I think the "stdout" logging should be the same as the default logging. If we want to avoid stdout spamming we should move the exec logs to debug (it actually is more debug than info IMHO) and maybe change the debug logging to runtime instead of build time configurable.
Or we could add new option to set default log level and discard everything below it (e.g. set default to warning and let user choose if they want to log info/debug etc.). Both cryptsetup and libdevmapper have similar functionality.

vpodzime: > If we want to avoid stdout spamming we should move the exec logs to debug (it actually is more debug than info IMHO)

I think this goes back to Blivet/Anaconda where stdout from spawned processes is `INFO`, IIRC. I agree `DEBUG` would make more sense.

vpodzime: > change the debug logging to runtime instead of build time configurable.
> Or we could add new option to set default log level and discard everything below it (e.g. set default to warning and let user choose if they want to log info/debug etc.).

Sounds like a good idea! :+1:

vojtechtrefny:
New function for setting log level added.

vpodzime:
Jenkins, test this, please.

vojtechtrefny:
Failing tests and docstrings fixed.

vpodzime:
> Failing tests and docstrings fixed.

:+1:

Issue #477, opened 2019-10-04
Add a generic function to get filesystem size


vojtechtrefny:
This will be helpful for UDisks -- it currently has a special function that needs to be edited every time we add support for a new filesystem -- https://github.com/storaged-project/udisks/blob/master/src/udiskslinuxfilesystem.c#L144

Related issue: https://github.com/storaged-project/udisks/pull/700

tbzatek:
UDisks could be ported to this new API within the 2.9.0 cycle now that we're open to more significant changes.

PR #480, opened 2019-10-15
Add a generic function to get size of a filesystem

vojtechtrefny:
Helper function that gets info for the filesystem and calculates
its size.

Fixes: #477

### DIFF
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -18,6 +18,7 @@
  */
 
 #include <glib.h>
+#include <glib/gstdio.h>
 #include <blkid.h>
 #include <string.h>
 #include <sys/types.h>
@@ -42,7 +43,8 @@ typedef enum {
     BD_FS_RESIZE,
     BD_FS_REPAIR,
     BD_FS_CHECK,
-    BD_FS_LABEL
+    BD_FS_LABEL,
+    BD_FS_GET_SIZE,
 } BDFsOpType;
 
 /**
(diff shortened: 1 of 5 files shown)


src/plugins/fs/generic.c
```diff
@@ -609,6 +650,104 @@ gboolean bd_fs_set_label (const gchar *device, const gchar *label, GError **erro
...
+/**
+ * bd_fs_get_size:
+ * @device: the device with file system to get size for
+ * @error: (out): place to store error (if any)
+ *
+ * Get size for filesystem on @device. This calls other fs info functions from this
+ * plugin based on detected filesystem (e.g. bd_fs_xfs_get_info for XFS). This
+ * function will return an error for unknown/unsupported filesystems.
+ *
+ * Returns: size of filesystem on @device, 0 in case of error.
```
vojtechtrefny: I'm not sure if it's safe to use `0` as a "error" value. Can we have a filesystem with zero size? Alternatively we could return boolean and return the size in an output variable.

src/plugins/fs/generic.c
```diff
@@ -474,6 +503,8 @@ static gboolean device_operation (const gchar *device, BDFsOpType op, guint64 ne
                 return bd_fs_ext4_check (device, NULL, error);
             case BD_FS_LABEL:
                 return bd_fs_ext4_set_label (device, label, error);
+            case BD_FS_SIZE:
+                g_assert_not_reached ();
```
vojtechtrefny: I'm usually strongly against asserts in libraries but this is not a user input validation. If something ends here, it's a programmer error (and we need to cover all cases in the switch to make gcc happy). Thoughts?

src/plugins/fs/generic.c
```diff
@@ -42,7 +43,8 @@ typedef enum {
     BD_FS_RESIZE,
     BD_FS_REPAIR,
     BD_FS_CHECK,
-    BD_FS_LABEL
+    BD_FS_LABEL,
+    BD_FS_SIZE,
```
vpodzime: `GET_SIZE` would make more sense to me.

src/plugins/fs/generic.c
vpodzime: I'd add a single assert to the beginning of the function to assert that `op != BD_FS_GET_SIZE`. And if GCC is not clever enough to take that as covered for the `switch` then just suppress the warning locally.

src/plugins/fs/generic.c
vpodzime: `0` is OK, I think. I don't think there can be an FS with zero size.

src/plugins/fs/generic.c
vojtechtrefny: Renamed.

src/plugins/fs/generic.c
vojtechtrefny: It's not only gcc. Also coverity and cppcheck and I don't want to add ten lines of comments and pragmas :)

tbzatek: Abort is just like a segfault - let the process fail, grab a backtrace and let developers fix it. That's in case of a programmer error, user input should be handled somewhere else.

Anyway, you may want to validate input at the very beginning of the function via `g_warn_if_fail()`/`g_return_val_if_fail()` and probably also use `g_warn_if_reached()` instead of a hard assert.

vpodzime: > Abort is just like a segfault - let the process fail, grab a backtrace and let developers fix it.

Isn't that exactly what we want here? This is not user input. It's an invalid call of a `static` function.

vojtechtrefny:
Jenkins, test this please.

PR #481, opened 2019-10-21
fix build on shells where 'test ==' fails

TimB87:
Hi!
This is related to #333 and #334. The build would fail at make shortly after entering docs directory.


vojtechtrefny:
Jenkins, ok to test.

Issue #483, opened 2019-11-06
bd_fs_wipe happily wipes mounted filesystems


mvollmer:
Calling bd_fs_wipe on a block device with a filesystem that is mounted will happily go ahead and remove the signatures.  It should fail with an error.

I think this can be achieved by adding O_EXCL here: https://github.com/storaged-project/libblockdev/blob/29332ce6ba52dc7fcfa193b6ed367ea112fa91c2/src/plugins/fs/generic.c#L124

vojtechtrefny:
Yes, `O_EXCL` will fix this, wipefs is using that -- https://github.com/karelzak/util-linux/blob/master/misc-utils/wipefs.c#L543

But do we want to fix this? There is a use case for wiping mounted devices, wipefs has a `--force` option that doesn't use `O_EXCL` and changing this might be seen as a change of behaviour.
The best solution would be to add a new `force` option to `bd_fs_wipe` but that's an API change.

Thoughts?

mvollmer:
Let's go one level up, to UDisks2.  It used to call wipefs without --force, and then got changed to call bd_fs_wipe, right?

So if the UDisks2 API has regressed in this regard, and we should fix that by having it call bd_fs_wipe with a new "exclusive" option.

Does that make sense?

tbzatek:
Makes sense for both of you. The docs of `bd_fs_wipe()` should be note this fact though. Perhaps introducing a new function such as `bd_fs_wipe_exclusive()` or `bd_fs_wipe_locked()` might do the trick and in UDisks we're fine with bumping `libblockdev` requirement for the 2.9 cycle.

vpodzime:
> The docs of bd_fs_wipe() should be note this fact though. Perhaps introducing a new function such as bd_fs_wipe_exclusive() or bd_fs_wipe_locked() might do the trick and in UDisks we're fine with bumping libblockdev requirement for the 2.9 cycle.

I think this is the right way to go.

vpodzime:
And for libblockdev-3.0, we should add the new argument if we break the API.

mvollmer:
Note that as far as Cockpit is concerned, things are mostly safe.  It will explicitly unmount filesystems before wiping them.

vojtechtrefny:
New function for 2.x: #484

We need to decide about the API change for master/3.0.

tbzatek:
> We need to decide about the API change for master/3.0.

Meaning the 3.0 would remove this new symbol again (and change the arguments)?

[...]

PR #484, opened 2019-11-07
Add new function 'bd_fs_wipe_force' to control force wipe

vojtechtrefny:
Our current implementation allows to wipe mounted devices by
default. We can't change this behaviour but we want to allow to
control this so we need a new function.

Related: https://github.com/storaged-project/libblockdev/issues/483

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -372,11 +372,28 @@ gboolean bd_fs_is_tech_avail (BDFSTech tech, guint64 mode, GError **error);
  * @all: whether to wipe all (%TRUE) signatures or just the first (%FALSE) one
  * @error: (out): place to store error (if any)
  *
+ * Note: This function will wipe signatures on a mounted @device without
+ *       asking. Use %bd_fs_wipe_force if you want to control this
+ *       behaviour manually.
+ *
  * Returns: whether signatures were successfully wiped on @device or not
  *
  * Tech category: %BD_FS_TECH_GENERIC-%BD_FS_TECH_MODE_WIPE
  */
 gboolean bd_fs_wipe (const gchar *device, gboolean all, GError **error);
+
+/**
+ * bd_fs_wipe_force:
+ * @device: the device to wipe signatures from
+ * @all: whether to wipe all (%TRUE) signatures or just the first (%FALSE) one
+ * @force: whether to force wipe even if the filesystem is mounted
+ * @error: (out): place to store error (if any)
+ *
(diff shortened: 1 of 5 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -378,6 +378,19 @@ gboolean bd_fs_is_tech_avail (BDFSTech tech, guint64 mode, GError **error);
...
+ * @device: the device to wipe signatures from
+ * @all: whether to wipe all (%TRUE) signatures or just the first (%FALSE) one
+ * @force: whether to force wipe even if the filesystem is mounted
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether signatures were successfully wiped on @device or not
+ *
+ * Tech category: %BD_FS_TECH_GENERIC-%BD_FS_TECH_MODE_WIPE
+ */
+gboolean bd_fs_wipe_force (const gchar *device, gboolean all, gboolean force, GError **error);
```
vpodzime: Please add a note about the potentially problematic behavior and this alternative function to the `bd_fs_wipe()` docs.

src/plugins/fs/generic.c
```diff
@@ -204,6 +199,35 @@ gboolean bd_fs_wipe (const gchar *device, gboolean all, GError **error) {
...
+ * @device: the device to wipe signatures from
+ * @all: whether to wipe all (%TRUE) signatures or just the first (%FALSE) one
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether signatures were successfully wiped on @device or not
+ *
+ * Tech category: %BD_FS_TECH_GENERIC-%BD_FS_TECH_MODE_WIPE
+ */
+gboolean bd_fs_wipe (const gchar *device, gboolean all, GError **error) {
+    return _fs_wipe (device, all, FALSE, error);
```
vpodzime: `force=FALSE` **is** a change in behavior here, right? AFAIK, #483 mentions we **don't want** to change behavior.

vojtechtrefny: Oops, fixed. I even added test to check the behaviour changed :man_facepalming:

src/lib/plugin_apis/fs.api
vojtechtrefny: I have added a note to `bd_fs_wipe`

src/plugins/fs/generic.c
vpodzime: Yeah, that was a bit confusing :smile:

PR #486, opened 2019-11-21
fs: Fix checking for UID/GID == 0

vojtechtrefny:
We need to check if the value was "0", not option itself.

### DIFF
--- a/src/plugins/fs/mount.c
+++ b/src/plugins/fs/mount.c
@@ -600,6 +600,7 @@ gboolean bd_fs_unmount (const gchar *spec, gboolean lazy, gboolean force, const 
     uid_t current_uid = -1;
     gid_t current_gid = -1;
     const BDExtraArg **extra_p = NULL;
+    gchar *endptr = NULL;
     MountArgs args;
 
     args.spec = spec;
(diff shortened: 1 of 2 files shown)


src/plugins/fs/mount.c
```diff
@@ -618,7 +618,7 @@ gboolean bd_fs_unmount (const gchar *spec, gboolean lazy, gboolean force, const
                 run_as_uid = g_ascii_strtoull ((*extra_p)->val, NULL, 0);
 
                 /* g_ascii_strtoull returns 0 in case of error */
-                if (run_as_uid == 0 && (g_strcmp0 ((*extra_p)->opt, "0") != 0)) {
+                if (run_as_uid == 0 && (g_strcmp0 ((*extra_p)->val, "0") != 0)) {
```
tbzatek: Good catch, however...

This is just a weird test. Conversion error detection should be done by checking the value returned via the second argument of `g_ascii_strtoull()`.

vojtechtrefny: Ok, time to do that correctly.

src/plugins/fs/mount.c
```diff
@@ -615,19 +616,19 @@ gboolean bd_fs_unmount (const gchar *spec, gboolean lazy, gboolean force, const
     if (extra) {
         for (extra_p=extra; *extra_p; extra_p++) {
             if ((*extra_p)->opt && (g_strcmp0 ((*extra_p)->opt, "run_as_uid") == 0)) {
-                run_as_uid = g_ascii_strtoull ((*extra_p)->val, NULL, 0);
+                run_as_uid = g_ascii_strtoull ((*extra_p)->val, endptr, 0);
 
                 /* g_ascii_strtoull returns 0 in case of error */
-                if (run_as_uid == 0 && (g_strcmp0 ((*extra_p)->opt, "0") != 0)) {
+                if (run_as_uid == 0 && endptr == &((*extra_p)->val)) {
```
vpodzime: Yeah, this is so much nicer :smile:

src/plugins/fs/mount.c
```diff
@@ -689,19 +691,19 @@ gboolean bd_fs_mount (const gchar *device, const gchar *mountpoint, const gchar
     if (extra) {
         for (extra_p=extra; *extra_p; extra_p++) {
             if ((*extra_p)->opt && (g_strcmp0 ((*extra_p)->opt, "run_as_uid") == 0)) {
-                run_as_uid = g_ascii_strtoull ((*extra_p)->val, NULL, 0);
+                run_as_uid = g_ascii_strtoull ((*extra_p)->val, endptr, 0);
```
vpodzime: Should be `&endptr` with `gchar *endptr`, AFAICT.

vojtechtrefny: Oops, fixed.

PR #488, opened 2019-12-23
Add some helper functions to get LVM VDO mode and state strings

vojtechtrefny:


src/plugins/lvm-dbus.c
```diff
@@ -3567,3 +3567,83 @@ gboolean bd_lvm_vdo_pool_resize (const gchar *vg_name, const gchar *pool_name, g
...
+const gchar* bd_lvm_get_vdo_operating_mode_str (BDLVMVDOOperatingMode mode, GError **error) {
+    switch (mode) {
+    case BD_LVM_VDO_MODE_RECOVERING:
+        return "recovering";
+    case BD_LVM_VDO_MODE_READ_ONLY:
+        return "read-only";
+    case BD_LVM_VDO_MODE_NORMAL:
+        return "normal";
+    case BD_LVM_VDO_MODE_UNKNOWN:
+        return "unknow";
```
tbzatek: Missing 'n'. One more occurrence below.d

vojtechtrefny: Fixed.

vojtechtrefny:
> Do we really care about returning an error? Wouldn't simple NULL be enough to indicate an invalid value?

It's not necessary, but it doesn't hurt. And setting the error will result in an exception in python which is better than returning `None`, IMHO.

tbzatek:
> And setting the error will result in an exception in python which is better than returning `None`, IMHO.

I see, good point.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #493, opened 2020-01-28
Add extra parameters for creating LVM VDO volumes

vojtechtrefny:
Our users expect to be able to manually control compression,
deduplication and VDO index memory size when creating LVM VDO
volumes. This also makes the API consistent with the VDO plugin
API.


vojtechtrefny:
> Please put the new lock_config argument changes into a separate commit. Looks good to me otherwise.

Fixed.

PR #498, opened 2020-02-07
Manually remove symlinks not removed by udev in tests

vojtechtrefny:
The metadata LV symlink is sometimes not removed after removing
a cache pool we created for tests. This is a huge problem for our
CI because other LVM tests will immediately fail when trying to
create the testVG volume group.
The bug was reported to LVM as rhbz#1800475

### DIFF
--- a/tests/lvm_dbus_tests.py
+++ b/tests/lvm_dbus_tests.py
@@ -5,6 +5,7 @@ import math
 import overrides_hack
 import six
 import re
+import shutil
 import subprocess
 from distutils.version import LooseVersion
 from itertools import chain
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

tests/lvm_dbus_tests.py
```diff
@@ -1201,6 +1202,10 @@ def _clean_up(self):
         except:
             pass
 
+        # lets help udev with removing stale symlinks
+        if not BlockDev.lvm_lvs("testVG") and os.path.exists("/dev/testVG/testCache_meta"):
+            shutil.rmtree("/dev/testVG")
```
tbzatek: Isn't this racy with udev? Does this throw an exception when the path is not found?

vojtechtrefny: Good point, fixed.

PR #500, opened 2020-02-14
Do not open devices as read-write for read-only fs operations

vojtechtrefny:
We don't need write support for operations like getting uuid or
label and closing the device opened as rw also emits udev change
event.


vpodzime:
> This could lead to an endless loop and constantly processing uevents.

That's exactly what was happening and what brought me to this code when I asked @vojtechtrefny to fix it.

PR #504, opened 2020-02-24
Manually add an extended metadata partition to list of partitions

vojtechtrefny:
When searching partition by position, libparted adds a special
metadata partition at the start of the extended partition.
libfdisk doesn't do this so we need to add such partition manually
when creating the list of all partitions for functions that search
a partition by its position.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -770,16 +770,15 @@ BDPartSpec* bd_part_get_part_spec (const gchar *disk, const gchar *part, GError 
     return ret;
 }
 
-static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean freespaces, GError **error) {
+static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean freespaces, gboolean metadata, GError **error) {
     struct fdisk_context *cxt = NULL;
     struct fdisk_table *table = NULL;
     struct fdisk_partition *pa = NULL;
     struct fdisk_iter *itr = NULL;
-    BDPartSpec **ret = NULL;
     BDPartSpec *spec = NULL;
-    guint i = 0;
+    BDPartSpec *prev_spec = NULL;
+    GPtrArray *array = NULL;
     gint status = 0;
-    size_t num_parts = 0;
 
     cxt = get_device_context (disk, error);
     if (!cxt) {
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
I've pushed a reworked version of the PR. We now "fill" all parts of the disk that are not freespace or partition by a "metadata partition". This is not exactly what parted does but it is as close as possible to it. Some of the "metadata" parts of the disk are actually freespaces in parted but too small to be usable, for example the with fdisk we have 1 MiB metadata between logical partitions, with parted this would less that 1 MiB of freespace and few KiBs of metadata.

src/plugins/part.c
```diff
@@ -827,38 +826,71 @@ static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean
...
-    i = 0;
+    array = g_ptr_array_new ();
 
     while (fdisk_table_next_partition(table, itr, &pa) == 0) {
         spec = get_part_spec_fdisk (cxt, pa, error);
         if (!spec) {
-            for (BDPartSpec **part_list_p = ret; *part_list_p; part_list_p++)
-                bd_part_spec_free (*part_list_p);
-            g_free (ret);
+            g_ptr_array_foreach (array, (GFunc) (void *) bd_part_spec_free, NULL);
```
tbzatek: You can use `g_ptr_array_new_with_free_func()` instead.

src/plugins/part.c
```diff
@@ -827,38 +826,71 @@ static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean
...
     }
-    ret[i] = NULL;
 
     fdisk_free_iter (itr);
     fdisk_unref_table (table);
     close_context (cxt);
 
-    return ret;
+    g_ptr_array_add (array, NULL);
+    return (BDPartSpec **) g_ptr_array_free (array, FALSE);;
```
tbzatek: Stray semicolon! ;-)

src/plugins/part.c
```diff
@@ -827,38 +826,71 @@ static BDPartSpec** get_disk_parts (const gchar *disk, gboolean parts, gboolean
...
+        /* libfdisk doesn't have a special partition for metadata so we need to add
+           a special metadata partition to the "empty" spaces between partitions
+           and free spaces to mimic behaviour of parted
+           metadata partitions should be present in the extended partition in front
+           of every logical partition */
+        if (prev_spec && metadata) {
+            if ((spec->start > prev_spec->start + prev_spec->size) ||
+                (prev_spec->type == BD_PART_TYPE_EXTENDED && spec->start > prev_spec->start) ) {
+                BDPartSpec *ext_meta = g_new0 (BDPartSpec, 1);
+                ext_meta->flags = 0;
```
tbzatek: You know `g_new0()` initializes everything to zero, right?

vojtechtrefny: Yes, I know. I wanted to set these explicitly here to emphasize that the partition without things like path and name is intention, not a mistake. I can remove it if you think it's not necessary.

src/plugins/part.c
vojtechtrefny: Two semicolons are better than one!!

src/plugins/part.c
tbzatek: OK, fair enough.

vojtechtrefny:
Jenkins, test this please.

PR #505, opened 2020-03-02
utils: Add functions to get and check current linux kernel version

tbzatek:


tbzatek:
So these are linux specific calls, I've put them in `module.h` along with other linux kernel module funcs. However the name "Linux" is ommitted at the current API, wondering whether libblockdev would ever want to support other platforms and if so, whether not to re-evaluate the API for the libblockdev-3.0 release.

tbzatek:
Hmm, looks like CentOS 7 gcc suffers from this initializer bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119

Hell, let's go with an uninitialzed struct, error cases should be covered anyway.

src/utils/module.c
```diff
@@ -250,3 +251,66 @@ gboolean bd_utils_unload_kernel_module (const gchar *module_name, GError **error
...
+/**
+ * bd_utils_get_linux_version:
+ * @error: (out): place to store error (if any)
+ *
+ * Retrieves version of currently running Linux kernel.
+ *
+ * Returns: detected Linux kernel version in the "MAJOR*1000000 + MINOR*1000 + MICRO" format.
+ */
+guint64 bd_utils_get_linux_version (GError **error)
+{
```
vojtechtrefny: Also please fix the `{}` brackets positions, this is not UDisks :-)

tbzatek: Fixed (hopefully).

tbzatek:
> Looks good to me in general, but I think the API should be more "user friendly". I think a new struct for the major, minor and micro (something like `BDUtilsLinuxVer`) would be better and also "standard" `cmp`-like function for version comparison returning -1, 0 or +1 could be useful.

Have a look at the changed API if it's okay with you. I'll try to come up with some tests in the meanwhile.

src/utils/module.c
```diff
@@ -250,3 +251,83 @@ gboolean bd_utils_unload_kernel_module (const gchar *module_name, GError **error
...
+      }
+
+    if (g_ascii_strncasecmp (buf.sysname, "Linux", sizeof buf.sysname) != 0) {
+        g_set_error (error, BD_UTILS_MODULE_ERROR, BD_UTILS_MODULE_ERROR_INVALID_PLATFORM,
+                     "Failed to get kernel version: spurious sysname '%s' detected", buf.sysname);
+        G_UNLOCK (detected_linux_ver);
+        return NULL;
+      }
+
+    if (sscanf (buf.release, "%d.%d.%d",
```
vpodzime: Please either initialize the struct with zeroes or compare the `sscanf()` return value to `3`.

tbzatek: I've added explicit memset and an `initialized` flag for your peaceful sleep in the night :-) (the static initializer over a struct still suffers from that gcc warning bug on RHEL7). I'm still convinced the struct is properly initialized by the compiler via the `static` keyword. And assuming simple data types as `gboolean` are atomic by nature.

As far as checking `sscanf()` result is concerned, it was on purpose to check if at least the first number got matched, in case we ever get a string like `6` instead of `6.0.0`.

PR #506, opened 2020-03-09
Add suport for setting filesystem UUID

vojtechtrefny:


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

src/plugins/fs/ext.c
```diff
@@ -520,6 +526,71 @@ gboolean bd_fs_ext4_set_label (const gchar *device, const gchar *label, GError *
     return ext_set_label (device, label, error);
 }
 
+static gboolean ext_set_uuid (const gchar *device, const gchar *uuid, GError **error) {
+    const gchar *args[5] = {"tune2fs", "-U", NULL, device, NULL};
+
+    if (!check_deps (&avail_deps, DEPS_TUNE2FS_MASK, deps, DEPS_LAST, &deps_check_lock, error))
+        return FALSE;
+
+    if (!uuid || g_strcmp0 (uuid, "") == 0)
```
vpodzime: Only checking `!uuid` for other file systems.

vojtechtrefny: Fixed. I originally thought that `NULL` and empty string should remove the UUID, but changed it lately to `NULL` generating a new one (which makes more sense and is supported by all filesystems) and for this case empty string doesn't makes sense. But I forgot to change the check for ext.

PR #507, opened 2020-03-10
ReiserFS support

vojtechtrefny:
First three commits are taken from #506. I have also found few f2fs issues when working this, these should be fixed now and f2fs tests should be actually running again.

### DIFF
--- a/src/plugins/fs/Makefile.am
+++ b/src/plugins/fs/Makefile.am
@@ -7,21 +7,23 @@ libbd_fs_la_LIBADD   = ${builddir}/../../utils/libbd_utils.la ${builddir}/../lib
 libbd_fs_la_LDFLAGS	 = -L${srcdir}/../../utils/ -version-info 2:0:0 -Wl,--no-undefined
 libbd_fs_la_CPPFLAGS = -I${builddir}/../../../include/ -I${srcdir}/../
 libbd_fs_la_SOURCES  = ../check_deps.c ../check_deps.h \
-						../fs.c   ../fs.h   \
-						common.c  common.h  \
-						ext.c     ext.h     \
-						generic.c generic.h \
-						mount.c   mount.h   \
-						ntfs.c    ntfs.h    \
-						vfat.c    vfat.h    \
-						xfs.c     xfs.h     \
-						f2fs.c    f2fs.h
+						../fs.c    ../fs.h    \
+						common.c   common.h   \
+						ext.c      ext.h      \
+						generic.c  generic.h  \
+						mount.c    mount.h    \
+						ntfs.c     ntfs.h     \
+						vfat.c     vfat.h     \
+						xfs.c      xfs.h      \
+						f2fs.c     f2fs.h     \
+						reiserfs.c reiserfs.h
 
 libincludefsdir = $(includedir)/blockdev/fs/
 libincludefs_HEADERS = ext.h     \
-					generic.h \
-					mount.h   \
-					ntfs.h    \
-					vfat.h    \
-					xfs.h     \
-					f2fs.h
+					generic.h  \
+					mount.h    \
+					ntfs.h     \
+					vfat.h     \
+					xfs.h      \
+					f2fs.h	   \
+					reiserfs.h
--- a/src/plugins/fs/reiserfs.c
+++ b/src/plugins/fs/reiserfs.c
@@ -1,0 +1,391 @@
+/*
+ * Copyright (C) 2020  Red Hat, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Vojtech Trefny <vtrefny@redhat.com>
+ */
+
+#include <blockdev/utils.h>
+#include <check_deps.h>
+
+#include "reiserfs.h"
+#include "fs.h"
+#include "common.h"
+
+static volatile guint avail_deps = 0;
+static GMutex deps_check_lock;
+
+#define DEPS_MKREISERFS 0
+#define DEPS_MKREISERFS_MASK (1 << DEPS_MKREISERFS)
+#define DEPS_REISERFSCK 1
+#define DEPS_REISERFSCK_MASK (1 << DEPS_REISERFSCK)
+#define DEPS_REISERFSTUNE 2
+#define DEPS_REISERFSTUNE_MASK (1 <<  DEPS_REISERFSTUNE)
+#define DEPS_DEBUGREISERFS 3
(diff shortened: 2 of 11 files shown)


src/plugins/fs/Makefile.am
```diff
@@ -7,21 +7,23 @@ libbd_fs_la_LIBADD   = ${builddir}/../../utils/libbd_utils.la ${builddir}/../lib
...
-						f2fs.c    f2fs.h
+						../fs.c    ../fs.h    \
+						common.c   common.h   \
+						ext.c      ext.h      \
+						generic.c  generic.h  \
+						mount.c    mount.h    \
+						ntfs.c     ntfs.h     \
+						vfat.c     vfat.h     \
+						xfs.c      xfs.h      \
+						f2fs.c     f2fs.h	  \
```
vpodzime: Might as well align the `\`

src/plugins/fs/reiserfs.c
```diff
@@ -0,0 +1,386 @@
...
+        return FALSE;
+
+    ret = g_new0 (BDFSReiserFSInfo, 1);
+
+    success = get_uuid_label (device, &(ret->uuid), &(ret->label), error);
+    if (!success) {
+        /* error is already populated */
+        bd_fs_reiserfs_info_free (ret);
+        return NULL;
+    }
```
vpodzime: This block getting the label should be done before we fork&exec the binary, it's cheaper. And the code would be better readable, I think.

src/plugins/fs/reiserfs.c
```diff
@@ -0,0 +1,386 @@
...
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_PARSE, "Failed to parse ReiserFS file system information");
+        g_strfreev (lines);
+        bd_fs_reiserfs_info_free (ret);
+        return FALSE;
+    }
+
+    /* extract data from something like this: "Free blocks (count of blocks - used [journal, bitmaps, data, reserved] blocks): 119529" */
+    val_start = strchr (*line_p, ':');
+    val_start++;
+    ret->free_blocks = g_ascii_strtoull (val_start, NULL, 0);
```
vpodzime: I wonder how much guaranteed the order of the items in the output is. Let's believe the tests will catch it when it changes and then we can parse the lines into a hash table like we do with LVM info.

src/plugins/fs/reiserfs.c
```diff
@@ -0,0 +1,386 @@
...
+ * @new_size: new requested size for the file system (if 0, the file system is
+ *            adapted to the underlying block device)
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the file system on @device was successfully resized or not
+ *
+ * Tech category: %BD_FS_TECH_REISERFS-%BD_FS_TECH_MODE_RESIZE
+ */
+gboolean bd_fs_reiserfs_resize (const gchar *device, guint64 new_size, GError **error) {
+    const gchar *args[6] = {"fsadm", "-y", "resize", device, NULL, NULL};
```
vpodzime: `fsadm` just runs `resize_reiserfs`, maybe we could do that directly?

src/plugins/fs/reiserfs.c
```diff
@@ -0,0 +1,386 @@
...
+    gboolean ret = FALSE;
+    BDFSReiserFSInfo *info = NULL;
+
+    if (!check_deps (&avail_deps, DEPS_RESIZEREISERFS_MASK, deps, DEPS_LAST, &deps_check_lock, error) ||
+        !check_deps (&avail_deps, DEPS_DEBUGREISERFS_MASK, deps, DEPS_LAST, &deps_check_lock, error))
```
vpodzime: `DEPS_DEBUGREISERFS_MASK` is checked by `bd_fs_reiserfs_get_info()`, right?

src/plugins/fs/Makefile.am
vojtechtrefny: Betryed by my editor, it look aligned (one hidden tab vs. spaces).

src/plugins/fs/reiserfs.c
vojtechtrefny: Ok. I thought starting with something more likely to fail is better, but I do it your way.

src/plugins/fs/reiserfs.c
vojtechtrefny: I think the output won't change here, reiserfs is pretty old and I assume all development happens in reiser5. We are parsing NTFS, vfat and f2fs outputs in similar way.

src/plugins/fs/reiserfs.c
vojtechtrefny: Good point, fixed.

src/plugins/fs/reiserfs.c
vojtechtrefny: For shrinking `resize_reiserfs` requires confirmation and it doesn't have `--yes` or similar option. I was too lazy and didn't want to write support for passing input to our exec functions and when I found that `fsadm` works I decided to use it instead. Anyway I have changed it and we now use `resize_reiserfs` directly. It's simpler and I also found that `fsadm` isn't packaged separately so our reiserfs implementation would depend on LVM :-)

vojtechtrefny:
> Additionally, `reiserfsprogs` provide a library, though GPLv2 with some _Anti-Plagiarism_ modification. After all, Namesys was a commercial company.

I saw that and I didn't want to add a compile time dependency to the FS plugin. People are already yelling at us for depending on things like libkmod and libcryptsetup so I guess they wouldn't be happy about this :-)

vojtechtrefny:
The csmock error is not new, I have just created a new path to the same error, which I believe is a false positive, see #236. Maybe someone smarter than me can take a look at the issue and figure out why cppcheck thinks that `g_set_error` doesn't set the error in this situation (and only in this situation, because it doesn't flag other similar lines).

src/utils/exec.c
```diff
@@ -344,6 +347,33 @@ static gboolean _utils_exec_and_report_progress (const gchar **argv, const BDExt
...
+                                              error);
+        if (io_status != G_IO_STATUS_NORMAL) {
+            if (error && (*error))
+                g_prefix_error (error, "Failed to write to stdin of the process: ");
+            else
+                g_set_error (error, BD_UTILS_EXEC_ERROR, BD_UTILS_EXEC_ERROR_FAILED,
+                             "Failed to write to stdin of the process.");
+
+            bd_utils_report_finished (progress_id, (*error)->message);
+            g_io_channel_shutdown (in_pipe, FALSE, NULL);
```
vpodzime: Missing `unref()` for the `in_pipe`?

src/utils/exec.c
```diff
@@ -344,6 +347,33 @@ static gboolean _utils_exec_and_report_progress (const gchar **argv, const BDExt
...
+            g_io_channel_shutdown (out_pipe, FALSE, NULL);
+            g_io_channel_unref (out_pipe);
+            g_io_channel_shutdown (err_pipe, FALSE, NULL);
+            g_io_channel_unref (err_pipe);
+            g_string_free (stdout_data, TRUE);
+            g_string_free (stderr_data, TRUE);
+            return FALSE;
+        } else
+            g_io_channel_shutdown (in_pipe, TRUE, NULL);
```
vpodzime: And here as well?

tbzatek:
> The csmock error is not new, I have just created a new path to the same error, which I believe is a false positive, see #236. Maybe someone smarter than me can take a look at the issue and figure out why cppcheck thinks that `g_set_error` doesn't set the error in this situation (and only in this situation, because it doesn't flag other similar lines).

I think the problem is that `error` can be NULL and `g_set_error()` simply returns in this case, while `error` is dereferenced a line later. `_utils_exec_and_report_progress()` is called from public library functions, I guess the static analyser stops here and assumes it's public and there's always a possibility of NULL input argument value.

PR #510, opened 2020-03-16
tests: Add option to run pdb for failed tests

vojtechtrefny:
Unfortunately the pdb is started at the point where the exception is raised so for failed assertions you'll end up in the unittest code and have to jump few levels up to the actual test but I didn't find any way how to workaround this.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -5,10 +5,12 @@ from __future__ import print_function
 import argparse
 import datetime
 import os
+import pdb
 import re
 import six
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
> Could you please prepare something similar for udisks as well?

Yes, that's the plan. I'll use libblockdev as a playground for some more changes and port these to UDisks later.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #511, opened 2020-03-17
Allow selecting tags when running tests

vojtechtrefny:
This shoudl allow things like running a separate job for unstable tests only -- we are currently ignoring all errors from tests marked as unstable, but there is a big difference between some of them failing and all of them failing which is easy to overlook. I'm planning to have a separate Jenkins job that will run only unstable tests and upload the results as a separate "category" so it will be easier to review.

Please take a look at the new command line options for selecting tags when running tests and let me know it they make sense and are understandable.

(The pdb commit is from #510.)

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -16,6 +16,8 @@ import yaml
 
 from distutils.spawn import find_executable
 
+from utils import TestTags, get_version
+
 LIBDIRS = 'src/utils/.libs:src/plugins/.libs:src/plugins/fs/.libs:src/lib/.libs'
 GIDIR = 'src/lib'
 
@@ -227,6 +229,10 @@ if __name__ == '__main__':
         sys.path.append(testdir)
         sys.path.append(projdir)
         sys.path.append(os.path.join(projdir, 'src/python'))
+
+    if not args.installed:
+        import gi.overrides
+        gi.overrides.__path__.insert(0, os.path.join(projdir, 'src/python/gi/overrides'))
 
     start_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
 
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -42,7 +42,7 @@ def _get_tests_from_suite(suite, tests):
 def _get_test_tags(test):
     """ Get test tags for single test case """
 
-    tags = []
+    tags = set()
 
     # test failed to load, usually some ImportError or something really broken
     # in the test file, just return empty list and let it fail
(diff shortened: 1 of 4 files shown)


vojtechtrefny:
> Would be nice to include unstable test results within the same log file as stable tests, for easier viewing from pull request test for example.

That's what we have now and we ignore the results of unstable tests. My long term goal is to have an HTML output for the tests, something similar to what cockpit has -- https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13793-20200326-063338-ea0c7bd5-debian-stable/log.html -- with a summary on top and nice overview of failed tests, but that's something for the future.
I'm also not sure if having the test results together would be better -- we either mark the failed unstable tests as not failed (we do this now) and this will lead to ignoring the failures, or we'll keep these failing and result will be all the tests failing all the time which is easy to get used to and simply ignore it as "this always fails, it's the unstable tests" and ignore "real" failures. I think having two jobs with two outputs is better for now.

tests/run_tests.py
```diff
@@ -249,41 +309,30 @@ def _parse_skip_config(config):
...
     ARCH = os.uname()[-1]
 
     # get list of tests to skip from the config file
     skipping = _parse_skip_config(os.path.join(testdir, SKIP_CONFIG))
 
-    # for some reason overrides_hack will fail if we import this at the start
-    # of the file
-    from utils import TestTags
+    # get sets of include/exclude tags as tags not strings from arguments
+    include_tags = set((TestTags.get_tag_by_value(t) for t in args.include_tags))
```
vpodzime: No need for double parantheses here.

tests/run_tests.py
```diff
@@ -249,41 +309,30 @@ def _parse_skip_config(config):
...
-        if TestTags.UNSAFE in tags or TestTags.EXTRADEPS in tags and not args.jenkins:
-            _print_skip_message(test, TestTags.UNSAFE)
-            continue
-        if TestTags.EXTRADEPS in tags and not args.jenkins:
-            _print_skip_message(test, TestTags.EXTRADEPS)
-            continue
-        if TestTags.SOURCEONLY in tags and args.installed:
-            _print_skip_message(test, TestTags.SOURCEONLY)
+        # if user specified include_tags, test must have at least one of these to run
+        if include_tags and not include_tags & tags:
```
vpodzime: Better put the `include_tags & tags` in parentheses, I think.

tests/run_tests.py
vojtechtrefny: Fixed.

tests/run_tests.py
vojtechtrefny: Fixed.

vojtechtrefny:
I'm moving this from draft to a "normal" PR, but before merging this, I'll also change the Jenkins to run the stable and unstable tests separately so we can see how this will look like in "real life".

tbzatek:
> That's what we have now and we ignore the results of unstable tests. My long term goal is to have an HTML output for the tests, something similar to what cockpit has -- https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13793-20200326-063338-ea0c7bd5-debian-stable/log.html -- with a summary on top and nice overview of failed tests, but that's something for the future.

That looks impressive! 

> I'm also not sure if having the test results together would be better -- we either mark the failed unstable tests as not failed (we do this now) and this will lead to ignoring the failures, or we'll keep these failing and result will be all the tests failing all the time which is easy to get used to and simply ignore it as "this always fails, it's the unstable tests" and ignore "real" failures. I think having two jobs with two outputs is better for now.

Right, I didn't realize there's the human problem where a human would only be tempted to click the failed tests in the PR, ignoring the passed ones even if they contain failed unstable tests.

vojtechtrefny:
Jenkins, test this please.

PR #512, opened 2020-03-31
exec: Disable encoding when reading data from stdout/stderr

vojtechtrefny:
Default encoding is UTF-8 and this is more strictly enforced in
newest GLib and things like nul bytes are not longer considered
valid.

See also https://github.com/storaged-project/udisks/pull/729 and
https://gitlab.gnome.org/GNOME/glib/merge_requests/967


vojtechtrefny:
This should fix failing tests on Debian where output of f2fs tools contain null bytes
```
gi.repository.GLib.GError: g_convert_error: Channel terminates in a partial character (3)
```

PR #514, opened 2020-04-09
fs: Add support for exFAT filesystem

vojtechtrefny:


vojtechtrefny:
Today Samsung announced a new exfat-utils implementation -- https://github.com/exfat-utils/exfat-utils -- it's just mkfs and fsck for now, but they already have some different command line options.

I hope they will consider keeping the new implementation backwards compatible where possible, but it's possible we'll have to check for the utils version and use different command line options based on which tools are being used.

src/plugins/fs/exfat.c
```diff
@@ -0,0 +1,302 @@
...
+ * @extra: (allow-none) (array zero-terminated=1): extra options for the creation (right now
+ *                                                 passed to the 'mkexfat' utility)
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether a new exfat fs was successfully created on @device or not
+ *
+ * Tech category: %BD_FS_TECH_EXFAT-%BD_FS_TECH_MODE_MKFS
+ */
+gboolean bd_fs_exfat_mkfs (const gchar *device, const BDExtraArg **extra, GError **error) {
+    const gchar *args[3] = {"mkexfatfs", device, NULL};
```
tbzatek, line 130: This looks simple enough so that even the other `exfat-utils` should work fine. This is going to be the most used tool and it's good that no fallback is needed at this time.

src/plugins/fs/exfat.c
```diff
@@ -0,0 +1,302 @@
...
+    if (!success) {
+        /* error is already populated */
+        bd_fs_exfat_info_free (ret);
+        return NULL;
+    }
+
+    success = bd_utils_exec_and_capture_output (args, NULL, &output, error);
+    if (!success)
+        /* error is already populated */
+        return FALSE;
```
tbzatek, line 267: Well, `FALSE` is basically `NULL` in C but I think it's a copy-paste error ;-)

vojtechtrefny: Fixed.

src/lib/plugin_apis/fs.api
```diff
@@ -502,6 +503,77 @@ GType bd_fs_reiserfs_info_get_type () {
...
+#define BD_FS_TYPE_EXFAT_INFO (bd_fs_exfat_info_get_type ())
+GType bd_fs_exfat_info_get_type();
+
+/**
+ * BDFSExfatInfo:
+ * @label: label of the filesystem
+ * @uuid: uuid of the filesystem
+ * @cluster_size: cluster size used by the filesystem
+ * @cluster_count: number of clusters in the filesystem
+ * @free_cluster_count: number of free clusters in the filesystem
```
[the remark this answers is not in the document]
vojtechtrefny: Oops, this is actually just copy-pasted doc string from vfat which I forgot to update, fixed.

src/plugins/fs/exfat.c
tbzatek: Thanks for fixing, what about freeing the `ret`?

vojtechtrefny: Oops, also fixed.

src/plugins/fs/exfat.c
vojtechtrefny: I've changed the mkfs and fsck commands to `mkfs.exfat` and `fsck.exfat`, these should work with both exfat-utils and the options are also compatible for now.

tbzatek: Oh, nice!

vojtechtrefny:
I've rebased this PR to the latest master and also made sure issues from #516 are fixed here too.

But I'm still not sure what to do with this PR. The exfatprogs from Samsung are being packaged for Fedora[1] but these still don't cover all the functionality the "old" FUSE implementation has. Especially some "dump" functionality is missing. And also the "API" is no longer compatible (`tune.exfat` vs `exfatlabel` for label manipulation). I'm moving this to draft for now and we'll see what happens in the future, but if exfatprogs release a new version with something like `dump.exfat` I'll probably change this to use exfatprogs and forget about exfat-utils.

vojtechtrefny:
So Fedora package review is still blocked and will be for some time, but exfatprogs is now available in Debian, so I'll rewrite this to use tools from exfatprogs and it will work only on Debian for now.

vojtechtrefny:
Rebased to master and fully rewritten to the new exfat-utils from Samsung (only big change is in `bd_fs_exfat_get_info`). I don't know when it will be available in Fedora, but it is on Debian so tests will run and 3.0 will take some more time to finish so maybe Fedora will catch up.

src/lib/plugin_apis/fs.api
```diff
@@ -598,6 +599,76 @@ GType bd_fs_nilfs2_info_get_type () {
...
+/**
+ * BDFSExfatInfo:
+ * @label: label of the filesystem
+ * @uuid: uuid of the filesystem
+ * @sector_size: sector size used by the filesystem
+ * @cluster_size: cluster size used by the filesystem
+ * @sector_count: number of sectors in the filesystem
+ * @free_sectors: number of free sectors in the filesystem
+ * @cluster_count: number of clusters in the filesystem
+ * @free_clusters: number of free clusters in the filesystem
```
tbzatek, line 614: Some mysterious disappearance of the struct members :zap:

src/plugins/fs/exfat.c
```diff
@@ -0,0 +1,317 @@
...
+        /* error is already populated */
+        bd_fs_exfat_info_free (ret);
+        return NULL;
+    }
+
+    lines = g_strsplit (output, "\n", 0);
+    g_free (output);
+
+    for (line_p=lines; *line_p; line_p++) {
+        if (!have_ss) {
```
tbzatek, line 280: Hmm, can't you just test for `ret->sector_size > 0` instead of a separate flag?

src/plugins/fs/exfat.c
```diff
@@ -0,0 +1,317 @@
...
+            val_start = g_strrstr (*line_p, "Block sector size : ");
+            if (val_start) {
+                ret->sector_size = g_ascii_strtoull (val_start + 20, NULL, 0);
```
tbzatek, line 283: Suggestion: replace those magic constants with `strlen()`, perhaps moving the strings to a `#define`. Just a matter of taste though.

src/plugins/fs/generic.c
```diff
@@ -82,6 +82,7 @@ static const BDFSInfo fs_info[] = {
     {"f2fs", "fsck.f2fs", "fsck.f2fs", "resize.f2fs", BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK, NULL, "dump.f2fs", NULL},
     {"reiserfs", "reiserfsck", "reiserfsck", "resize_reiserfs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK, "reiserfstune", "debugreiserfs", "reiserfstune"},
     {"nilfs2", NULL, NULL, "nilfs-resize", BD_FS_ONLINE_GROW | BD_FS_ONLINE_GROW, "tune-nilfs", "tune-nilfs", "tune-nilfs"},
+    {"exfat", "fsck.exfat", "fsck.exfat", NULL, 0, "exfatlabel", "dumpexfat", NULL},
```
tbzatek, line 85: Doesn't match with `tune.exfat` called above. Is that on purpose or just leftover from the older tools?

tbzatek (approved):
Looks good generally, minor suggestions found. :hatching_chick:

src/lib/plugin_apis/fs.api
vojtechtrefny: Yeah, I forgot to remove these from the docstring. Unfortunately the new exfatprogs don't provide these values.

src/plugins/fs/exfat.c
vojtechtrefny: I think the extra variables make it a little bit more readable, but ok.

src/plugins/fs/exfat.c
vojtechtrefny: Ok, I have moved the strings and numbers to constants.

src/plugins/fs/generic.c
vojtechtrefny: No, I forgot to rewrite this to the new tools, fixed.

tbzatek (commented):
Thanks! :shipit:

vojtechtrefny:
Jenkins, test this please.

Issue #515, opened 2020-04-09
Inconsistent behaviour of _is_tech_avail functions in the FS plugin


vojtechtrefny:
Not all filesystems support all functions, for example vfat doesn't support UUIDs, exFAT can't be resized etc. and this is not always reflected in the `_is_tech_avail` functions -- we sometimes return TRUE for queries for these "missing" tech/mode combinations (because there is no missing external tool) and sometimes return error (because this is not supported).
Also `bd_fs_check_deps` is currently broken -- it doesn't check the newly added filesystems and expects that all modes are available for all filesystems and should be checked.

PR #516, opened 2020-04-16
Fix checking for tech modes in the filesystem plugin

vojtechtrefny:
Fixes: #515 

There are no tests for this because the `*_check_deps` functions are not in the Python API, I don't know if this is intentional. Should I add these these to the API files so I can test this or should I try to find a different way? @vpodzime

### DIFF
--- a/src/plugins/fs.c
+++ b/src/plugins/fs.c
@@ -104,6 +104,25 @@ gboolean bd_fs_check_deps (void) {
         bd_utils_log_format (BD_UTILS_LOG_WARNING, "%s", error->message);
         g_clear_error (&error);
     }
+    ret = ret && bd_fs_f2fs_is_tech_avail (BD_FS_TECH_F2FS,
+                                           BD_FS_TECH_MODE_MKFS | BD_FS_TECH_MODE_WIPE |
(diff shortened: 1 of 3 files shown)


tbzatek:
Perhaps a subject to API break in libblockdev-3.0 ?

vpodzime:
>  Yeah, I think it would be nice to be able to query the supported operations for different file systems even from the GI API.

But that should be possible with [`bd_fs_is_tech_avail()`](http://storaged.org/libblockdev/libblockdev-FS.html#bd-fs-is-tech-avail), right?

vojtechtrefny:
> > Yeah, I think it would be nice to be able to query the supported operations for different file systems even from the GI API.
> 
> But that should be possible with [`bd_fs_is_tech_avail()`](http://storaged.org/libblockdev/libblockdev-FS.html#bd-fs-is-tech-avail), right?

Yes, the `_check_deps` functions are used only during initialization IMHO.

vojtechtrefny:
> Perhaps a subject to API break in libblockdev-3.0 ?

Adding new functions to the python API shouldn't be a problem.

PR #520, opened 2020-04-28
lvm: Add a function to get VDO pool name for a VDO LV

vojtechtrefny:
Small helper function for Blivet similar to bd_lvm_thlvpoolname.

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -2601,6 +2601,8 @@ gchar* bd_lvm_thlvpoolname (const gchar *vg_name, const gchar *lv_name, GError *
     if (!prop)
         return NULL;
     g_variant_get (prop, "o", &pool_obj_path);
+    g_variant_unref (prop);
+
     prop = get_object_property (pool_obj_path, LV_CMN_INTF, "Name", error);
     g_free (pool_obj_path);
     if (!prop)
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -3655,6 +3655,52 @@ gboolean bd_lvm_vdo_pool_convert (const gchar *vg_name UNUSED, const gchar *pool
 }
 
 /**
+ * bd_lvm_vdolvpoolname:
+ * @vg_name: name of the VG containing the queried VDO LV
+ * @lv_name: name of the queried VDO LV
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: (transfer full): the name of the pool volume for the @vg_name/@lv_name
+ * VDO LV or %NULL if failed to determine (@error) is set in those cases
+ *
+ * Tech category: %BD_LVM_TECH_VDO-%BD_LVM_TECH_MODE_QUERY
+ */
+gchar* bd_lvm_vdolvpoolname (const gchar *vg_name, const gchar *lv_name, GError **error) {
+    GVariant *prop = NULL;
+    const gchar *segtype = NULL;
+    gchar *pool_obj_path = NULL;
+    gchar *ret = NULL;
+
+    prop = get_lv_property (vg_name, lv_name, "SegType", error);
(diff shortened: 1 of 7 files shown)


src/plugins/lvm-dbus.c
```diff
@@ -3623,6 +3623,50 @@ gboolean bd_lvm_vdo_pool_convert (const gchar *vg_name UNUSED, const gchar *pool
...
 }
 
+/**
+ * bd_lvm_vdolvpoolname:
+ * @vg_name: name of the VG containing the queried VDO LV
+ * @lv_name: name of the queried VDO LV
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: (transfer full): the name of the pool volume for the @vg_name/@lv_name
+ * VDO LV or %NULL if failed to determine (@error) is set in those cases)
```
tbzatek, line 3633: Extra ')'

src/plugins/lvm-dbus.c
```diff
@@ -3623,6 +3623,50 @@ gboolean bd_lvm_vdo_pool_convert (const gchar *vg_name UNUSED, const gchar *pool
...
+    gchar *pool_obj_path = NULL;
+    gchar *ret = NULL;
+
+    prop = get_lv_property (vg_name, lv_name, "SegType", error);
+    if (!prop)
+        return NULL;
+
+    g_variant_get_child (prop, 0, "s", &segtype);
+    if (g_strcmp0 (segtype, "vdo") != 0) {
+        g_set_error (error, BD_LVM_ERROR, BD_LVM_ERROR_NOEXIST,
```
tbzatek, line 3649: Leaking `prop` and `segtype`. FYI, I think you can use `&s` as a format string and pass `const gchar *` instead. That would save you two lines of code and some CPU cycles!

src/plugins/lvm-dbus.c
```diff
@@ -3623,6 +3623,50 @@ gboolean bd_lvm_vdo_pool_convert (const gchar *vg_name UNUSED, const gchar *pool
...
+                     "The LV '%s' is not a VDO LV and thus have no VDO pool", lv_name);
+        return NULL;
+    }
+    g_free (segtype);
+    g_variant_unref (prop);
+
+    prop = get_lv_property (vg_name, lv_name, "PoolLv", error);
+    if (!prop)
+        return NULL;
+    g_variant_get (prop, "o", &pool_obj_path);
```
tbzatek, line 3659: Just like above, you may use `&o` to get a const string.

src/plugins/lvm-dbus.c
```diff
@@ -3623,6 +3623,50 @@ gboolean bd_lvm_vdo_pool_convert (const gchar *vg_name UNUSED, const gchar *pool
...
+    prop = get_object_property (pool_obj_path, LV_CMN_INTF, "Name", error);
```
tbzatek, line 3660: Leaking previous `prop` ?

src/plugins/lvm-dbus.c
vojtechtrefny: > FYI, I think you can use `&s` as a format string and pass `const gchar *` instead. That would save you two lines of code and some CPU cycles!

Cool, I didn't know this is possible.

src/plugins/lvm-dbus.c
vojtechtrefny: Fixed. And I've also fixed the same problem in the function I copied this from :)

src/plugins/lvm-dbus.c
vojtechtrefny: If I use `&o` here, the path would be freed with `g_variant_unref (prop)`, right? So I can't use it here because I need the path later.

tbzatek: Right, you would need to keep the `prop` around and free later. Anyway, looks good as it is.

PR #529, opened 2020-05-19
configure.ac: Avoid more bashisms

Polynomial-C:
My preivous PR didn't cover all bashisms. Matt was so kind to provide a patch which removes the remaining bashisms.

Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


vojtechtrefny:
Jenkins, ok to test.

Issue #458, opened 2019-07-24
Activate functions should not fail for activated devices


vojtechtrefny:
Functions that start/activate a device like `bd_md_activate` or `bd_dm_activate_raid_set` should not fail if the device is already active with given properties.

See https://github.com/storaged-project/blivet/pull/784

PR #536, opened 2020-06-02
Do not try to activate already active RAID devices

vojtechtrefny:
Fixes: #458

### DIFF
--- a/src/plugins/dm.c
+++ b/src/plugins/dm.c
@@ -516,6 +516,29 @@ static void find_raid_sets_for_dev (const gchar *name, const gchar *uuid, gint m
         }
     }
 }
+
+static gboolean get_raid_set_status (struct raid_set *rs, GError **error) {
+    struct dm_task *dmt = NULL;
+    struct dm_info info;
+    gboolean ret;
+
+    dmt = dm_task_create (DM_DEVICE_STATUS);
+    if (!dmt) {
+        g_set_error (error, BD_DM_ERROR, BD_DM_ERROR_TASK,
+                     "Failed to create DM task");
+        return FALSE;
+    }
+
+    if (!dm_task_set_name (dmt, rs->name) || !dm_task_run (dmt) || !dm_task_get_info (dmt, &info)) {
+        /* the DM device doesn't exist or some weird error appeared, just assume the RAID set is not active */
+        dm_task_destroy (dmt);
+        return FALSE;
+    }
+
+    ret = info.exists;
+    dm_task_destroy (dmt);
+    return ret;
+}
 #endif // WITH_BD_DMRAID
 
 /**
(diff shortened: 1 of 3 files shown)


src/plugins/dm.c
```diff
@@ -516,6 +516,29 @@ static void find_raid_sets_for_dev (const gchar *name, const gchar *uuid, gint m
...
+    gboolean ret;
+
+    dmt = dm_task_create (DM_DEVICE_STATUS);
+    if (!dmt) {
+        g_set_error (error, BD_DM_ERROR, BD_DM_ERROR_TASK,
+                     "Failed to create DM task");
+        return FALSE;
+    }
+
+    if (!dm_task_set_name (dmt, rs->name) || !dm_task_run (dmt) || !dm_task_get_info (dmt, &info)) {
```
vojtechtrefny, line 532: I'm not sure which of these can fail because the DM device doesn't exist and which means something is really broken (and libdevmapper "documentation" is not very helpful here) so I've decided to just ignore everything here. Worst case scenario is the RAID set change will fail so same behaviour we have now for some weird/unsupported cases.

src/plugins/mdraid.c
```diff
@@ -720,10 +720,24 @@ gboolean bd_md_activate (const gchar *raid_spec, const gchar **members, const gc
...
 
     if (!check_deps (&avail_deps, DEPS_MDADM_MASK, deps, DEPS_LAST, &deps_check_lock, error))
         return FALSE;
 
+    data = bd_md_detail (raid_spec, error);
+    if (!data)
+        g_clear_error (error);
+    else {
+        bd_utils_log_format (BD_UTILS_LOG_INFO,
+                             "RAID array '%s' is already activate with %"G_GUINT64_FORMAT" devices"
```
tbzatek, line 733: missing _d_ in _activate_

vojtechtrefny: Fixed.

src/plugins/dm.c
```diff
@@ -516,6 +516,29 @@ static void find_raid_sets_for_dev (const gchar *name, const gchar *uuid, gint m
         }
     }
 }
+
+static gboolean get_raid_set_status (struct raid_set *rs, GError **error) {
```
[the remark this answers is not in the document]
vojtechtrefny: We do this at many places in libblockdev and this is just an internal function. But i can add the `status` as an output parameter if you really want :-)

src/plugins/dm.c
vojtechtrefny: It's little bit too long for the comment, but I've adjusted the comment a little to contain more information.

src/plugins/dm.c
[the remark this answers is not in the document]
tbzatek: That's okay probably, didn't notice this was a static function.

Issue #471, opened 2019-09-17
Use crypt_keyfile_read to read key files


vojtechtrefny:
We should use the function available in cryptsetup to read key files and securely erase the passphrase later using `crypt_safe_free` (not yet part of the API, see https://gitlab.com/cryptsetup/cryptsetup/merge_requests/63).

Issue #527, opened 2020-05-14
Please review use of NULL GError**


tbzatek:
Just came across another segfault leading from a `NULL` error passed in `libblockdev` calls: `bd_vdo_info()` specifically leading to NULL pointer dereference in `_utils_exec_and_report_progress()`.

While I remember the general guideline for `libblockdev` being a _always pass non-NULL error_, I think this should be also guarded by `g_warn_if_fail (error != NULL)` in functions that dereference it. Or, make `libblockdev` safe with `NULL` error everywhere. 

In any case this is a difference from general GLib guidelines and it should be noted on a visible place.

vojtechtrefny:
I think we should allow passing `NULL` instead of error for our functions. It sometimes wouldn't make a lot of sense but it would definitely be easier and safer for our users.

vojtechtrefny:
So we now have two options:

1. Force our users to always provide us non-`NULL` GError. That's basically what we do right now, but is a little problematic because some functions can work just fine without the error set and some will crash when trying to dereference it. If we decided to keep doing this, we probably should force this during compilation by using the [nonnull attribute](https://gcc.gnu.org/onlinedocs/gcc-3.4.3/gcc/Function-Attributes.html).

2. Allow NULL instead of errors. This is what GLib does and it is part of the [Rules for use of GError](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html). We could simply log error instead of returning it and make sure we don't dereference it in our code. The problem here are some of our functions that return gboolean and `FALSE` could mean both error and valid result, for example `bd_fs_is_mountpoint` where `FALSE` could mean both "not a mountpoint" and "error when checking". For these function we'd need to:
  
      - Add another output argument or
      - change the return type to int to be able distinguish between true, false and error (-1) or
      - still force users to create GErrror if they want to be able to tell difference between "false" and "false with error".

Thoughts @vpodzime @tbzatek

vpodzime:
I vote for the option 2. And regarding the functions where it would be ambiguous, I don't care. It's the callee's responsibility to pass non-`NULL` `GError **` if they want to know if there was an error in those functions. IIRC, the documentation of these functions mentions that `FALSE` is returned in multiple cases and `GError **error` should be use to tell the difference. If they code that calls these functions doesn't care, that's their problem, not ours.

tbzatek:
[...]

PR #540, opened 2020-06-04
Use crypt_keyfile_read to read key files

vojtechtrefny:
Fixes: #471

This also unfortunately means bumping required version of cryptsetup to 2.3.0 which means libblockdev won't compile on older distributions like RHEL 7 (which also doesn't have libfdisk so we can't support it anyway), Debian 10 (which I assume wouldn't update to 3.0 anyway) and RHEL 8 (cryptsetup 2.3.0 was added in 8.2 which means our CentOS 8 machines won't work for now, but CentOS Stream is ok).

`crypt_keyfile_read` also needs `crypt_device` which made the code a little bit more complicated and added some duplication between the blob and non-blob functions but this will be fixed in the planned crypto plugin API cleanup (which I'm not doing now because it would break UDisks daily builds).


src/plugins/crypto.c
```diff
@@ -1204,45 +1215,106 @@ gboolean bd_crypto_luks_add_key (const gchar *device, const gchar *pass, const g
...
+    gchar *msg = NULL;
+
+    msg = g_strdup_printf ("Started adding key to the LUKS device '%s'", device);
+    progress_id = bd_utils_report_started (msg);
+    g_free (msg);
 
     if (!pass && !key_file) {
         g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_NO_KEY,
                      "No passphrase nor key file given, cannot add key.");
+        bd_utils_report_finished (progress_id, (*error)->message);
```
tbzatek, line 1230: Just thinking about adding `g_warn_if_fail (error != NULL)` since you now dereference `error`.

vojtechtrefny: Lets deal with this in #527

vojtechtrefny:
We discussed this with @vpodzime a little and even if don't 100 % agree with his point of view, it's probably not the best decision to require cryptsetup 2.3.0 right now when it's not available in major distributions. Lets wait for now, libblockdev 3.0 will take some more time and we can revisit this before the release.

vojtechtrefny:
Replaced by #752

Issue #427, opened 2019-02-07
Add python API documentation


vojtechtrefny:
It is possible to automatically generate documentation for the python API, so we should consider doing that and add it to http://storaged.org/libblockdev/

PR #541, opened 2020-06-05
Python bindings documentation

vojtechtrefny:
Fixes: #427

This is my first real experience with buildah and containers so I hope it makes at least a little sense :-)

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -30,7 +30,42 @@ MAKING RELEASES
 
    - [ ] upload the tarball created above (``make release``) to the release.
 
+ - [ ] Generate documentation for the Python bindings as described below and copy
+       it to *docs/html*
+
  - [ ] Update the documentation by rsyncing the contents of the *docs/html*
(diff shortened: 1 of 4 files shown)


misc/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN ./autogen.sh && ./configure --prefix=/usr && make -j6 && make install
+
+WORKDIR /root
+
+# get latest pgi-docgen and generate documentation for libblockdev
+RUN git clone https://github.com/pygobject/pgi-docgen
+
+WORKDIR /root/pgi-docgen
+
+RUN ./tools/build.sh BlockDev-2.0
```
vpodzime, line 39: AFAICT, each `RUN` line in this file commits a new layer.

vojtechtrefny: podman says the image has only 2 layers, so I assume beuildah doesn't create a new layer for every RUN.

vojtechtrefny:
> I'd, however, at least rename the file so that it refers to building docs 

Renamed.

Issue #550, opened 2020-07-23
mkfs.vfat exit code not handled correctly


hsattler:
Hi,

I use libblockdev via udisks2. In udisks2, I call the Repair method on a vfat USB stick.
If the stick was removed without unmounting before, this will trigger a real repair.

However, fsck.vfat returns with exit code 1. This is expected according to its man page:
0 -> No recoverable errors have been detected.
1 -> Recoverable errors have been detected or dosfsck has discovered an internal inconsistency.

Thus udisks2 always returns an error message.

Please fix the exit code handling....

vojtechtrefny:
Sorry, I didn't realize `fsck.vfat` returns non-zero even when all the errors are fixed. Thank you for the report, #551 should fix this.

hsattler:
For fixing this from outside udisks2, I took a different approach that you also might want to consider:
After repair return an error, I call check again to make sure that the repair worked. The check return with exit code 0 on success.

tbzatek:
That sounds like a good idea, thanks @hsattler!

vojtechtrefny:
Good idea, I've updated the PR. Thanks.

PR #551, opened 2020-07-23
fs: Ignore return code 1 during VFAT repair

vojtechtrefny:
fsck.vfat returns 1 ("Recoverable errors have been detected") even
when all the errors were successfully fixed so we should not set
error in this case.

Fixes: #550


tbzatek:
Wait, what about "Recoverable errors have been detected but not fixed" scenario? Would a repair failure result in `0` return code? Manpage only lists three return codes:

```
EXIT STATUS
       0   No recoverable errors have been detected.
       1   Recoverable errors have been detected or fsck.fat has discovered an internal inconsistency.
       2   Usage error.  fsck.fat did not access the filesystem.
```

vojtechtrefny:
> Wait, what about "Recoverable errors have been detected but not fixed" scenario? Would a repair failure result in `0` return code?

Good question. And the answer is: "I have no idea" :-)

hsattler:
Please see my comment in the bug report: https://github.com/storaged-project/libblockdev/issues/550#issuecomment-663395300

Issue #509, opened 2020-03-13
Missing functionality in FS plugin


vojtechtrefny:
Both Blivet and UDisks are not using the FS plugin for filesystem operations. UDisks uses libblockdev for some operations (resize mostly), Blivet doesn't use libblockdev at all. The main reason is missing functionality, especially missing filesystems like reiserfs or exfat and missing parameters like label for mkfs functions.

List of missing/covered functionality: https://gist.github.com/vojtechtrefny/deede90304788563215b93b2d2648d85

vojtechtrefny:
All functionality needed for UDisks and blivet should now be covered by #621 #781 and other recent changes to the plugin.

PR #553, opened 2020-08-03
Add functions to check filesystem label validity

vojtechtrefny:
One of the missing filesystem functionality needed by Blivet. Related: #509

### DIFF
--- a/src/plugins/fs/xfs.c
+++ b/src/plugins/fs/xfs.c
@@ -228,6 +228,35 @@ gboolean bd_fs_xfs_set_label (const gchar *device, const gchar *label, GError **
 }
 
 /**
+ * bd_fs_xfs_check_label:
+ * @label: label to check
+ * @error: (out) (allow-none): place to store error
+ *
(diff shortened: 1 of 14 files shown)


src/plugins/fs/xfs.c
```diff
@@ -227,6 +227,37 @@ gboolean bd_fs_xfs_set_label (const gchar *device, const gchar *label, GError **
...
+    size_t len = 0;
+
+    len = strlen (label);
+    if (len > 12) {
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_LABEL_INVALID,
+                     "Label for XFS filesystem must be at most 12 characters long.");
+        return FALSE;
+    }
+
+    for (size_t i = 0; i < len; i++) {
```
tbzatek, line 250: You may also use `strchr()` instead of the cycle...

vojtechtrefny: Thanks, I completely forgot that `strchr` exists... Fixed.

PR #554, opened 2020-08-04
fs: Add functions to check filesystem UUID validity

vojtechtrefny:
First commit is from #553


vojtechtrefny:
Rebased to latest master.

PR #556, opened 2020-08-10
spec: Use "python3_pkgversion" instead of hardcoding python3

vojtechtrefny:
This should fix failing RPM build tests on EPEL 8.

### DIFF
--- a/dist/libblockdev.spec.in
+++ b/dist/libblockdev.spec.in
@@ -142,7 +142,7 @@ BuildRequires: python2-devel
 Obsoletes:     python2-blockdev < %{version}-%{release}
 %endif
 %if %{with_python3}
-BuildRequires: python3-devel
+BuildRequires: python%{python3_pkgversion}-devel
 %endif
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
> This should fix failing RPM build tests on EPEL 8.

Or not...

This works locally for me, but for some reason not in our CI. When running in our CI something enables python38 module stream and installs python38-rpm-macros from it, but not python38 (but we still have the "default" python36). And because of this the `%{python3_sitearch}` macro we use in `%files` doesn't work:

```
<mock-chroot> sh-4.4# rpm --eval "%{python3_sitearch}"
sh: /usr/bin/python3.8: No such file or directory
```

vojtechtrefny:
Jenkins, test this please.

tbzatek:
No idea from my side on this one.

vojtechtrefny:
Looks like this fixed itself and we don't need to change spec for EPEL 8 after all.

PR #558, opened 2020-08-13
configure: Fix libuuid requirement for FS plugin

vojtechtrefny:
It's needed only for FS plugin, not for crypto or swap.
Added to a wrong condition in 4e96ceebc7dc51a728c80c29a606a6a37269efe7


vojtechtrefny:
Jenkins, test this please.

Issue #245, opened 2017-06-27
Migrate libblockdev from libparted to libfdisk


squimrel:
Basically libparted is way to smart for libblockdev.

For example libblockdev [only knows two table types](https://github.com/storaged-project/libblockdev/blob/dbe13d5ca58d698b340a268f5ce3556e56e07616/src/plugins/part.c#L176).

Where as [libparted knows a lot more](https://git.savannah.gnu.org/cgit/parted.git/tree/libparted/labels) even things like the mac label.

Which is not necessarily a good thing. For example I ran into a specific issue where the [isohybrid](http://www.syslinux.org/wiki/index.php?title=Isohybrid) layout is treated as a mac label and therefore I couldn't add a fourth partition to it because a mac label may only have a maximum of three partitions.
Basically in my use-case I'd only like to modify the MBR partition table (The isohybrid layout consists of multiple partition tables)

[libfdisk on the other hand is much simpler](https://github.com/karelzak/util-linux/tree/master/libfdisk/src). It only knows the bsd and sun label apart from the dos and mac label.

It also seems like this project could benefit from libfdisk since in libfdisk you can specify the partition number and use default values determined by libfdisk for start and end if you want to.

Apparently I poorly don't seem to have time to work on this myself at the moment but I would really appreciate it if you would work on this.

MartinBriza:
Please also include a concrete example of a use case when this is a problem. Like what kind of media you use and what are you trying to achieve with it using UDisks

squimrel:
We're trying to add a fourth partition on a flash drive that already has a Fedora image on it. We'd like to use this fourth partition to store an overlay file that's used for persistent storage when booted into the live system.

squimrel:
By the way it would be reasonable to argue that trying manipulate partitions on an isohybrid layout is nonsense since the tools I know of only modify one partition table at the time and isohybrid has three partition tables on it.

vpodzime:
> ... isohybrid has three partition tables on it.

Yes, but aren't two of them basically just "masks" ensuring that the device is bootable on various platforms? Is there any specification of which of the partition tables to actually respect as a partitioning schema of the device?

squimrel:
> aren't two of them basically just "masks"

I'm not sure if all the partition tables specify the same partitions.

> Is there any specification of which of the partition tables to actually respect as a partitioning schema of the device?

[...]

Issue #557, opened 2020-08-12
libfdisk part resize fails when new size is few MiBs bigger than max size


vojtechtrefny:
with libparted:

```
In [3]: BlockDev.part_resize_part("/dev/sdh", "/dev/sdh8", 423*1024**2, BlockDev.PartAlign.OPTIMAL)                                                                                                               
Out[3]: True
```

New size is "rounded down" to actual max size (419 MiB):
```
sdh8                        8:120  0  419M  0 part
```

Our current implementation allows up to 4 MiB "difference" in size with alignments other than NONE, see https://github.com/storaged-project/libblockdev/blob/master/src/plugins/part.c#L828

with libfdisk:
```
In [3]: BlockDev.part_resize_part("/dev/sdh", "/dev/sdh8", 423*1024**2, BlockDev.PartAlign.OPTIMAL)                                                                                                               

** (process:78464): WARNING **: 12:58:28.732: [fdisk] Failed to resize partition #8.
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-3-cc09488348f6> in <module>
----> 1 BlockDev.part_resize_part("/dev/sdh", "/dev/sdh8", 423*1024**2, BlockDev.PartAlign.OPTIMAL)

Error: g-bd-part-error-quark: Failed to resize partition '/dev/sdh8': Numerical result out of range (2)
```

PR #559, opened 2020-08-13
libfdisk based part plugin

vojtechtrefny:
Fixes: #245

So, this is finally it. After nearly 2 years we have the final version of the libfdisk rewrite of the part plugin.

I'll keep this as a draft for now. I plan to go through the code once more and make sure nothing got broken during the countless rebases etc. I also plan to go through the tests and make sure all corner cases are covered and also run the UDisks test suite with this version.

All parts of this PR have been reviewed before (except the last two commits I added today) but feel free to review this again. But this should really be the final version and I plan to fix smaller issues like #557 separately.

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -191,14 +191,11 @@ static gboolean set_parted_error (GError **error, BDPartError type) {
 static volatile guint avail_deps = 0;
 static GMutex deps_check_lock;
 
-#define DEPS_SGDISK 0
-#define DEPS_SGDISK_MASK (1 << DEPS_SGDISK)
-#define DEPS_SFDISK 1
+#define DEPS_SFDISK 0
 #define DEPS_SFDISK_MASK (1 << DEPS_SFDISK)
-#define DEPS_LAST 2
+#define DEPS_LAST 1
 
 static const UtilDep deps[DEPS_LAST] = {
-    {"sgdisk", "0.8.6", NULL, "GPT fdisk \\(sgdisk\\) version ([\\d\\.]+)"},
     {"sfdisk", NULL, NULL, NULL},
 };
 
@@ -260,6 +257,8 @@ void bd_part_close (void) {
     c_locale = (locale_t) 0;
 }
 
+#define UNUSED __attribute__((unused))
+
 /**
  * bd_part_is_tech_avail:
  * @tech: the queried tech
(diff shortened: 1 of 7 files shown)


src/plugins/part.c
```diff
@@ -467,65 +690,161 @@ static BDPartSpec* get_part_spec (PedDevice *dev, PedDisk *disk, PedPartition *p
...
         return NULL;
     }
 
-    ret = get_part_spec (dev, ped_disk, ped_part, error);
+    array = g_ptr_array_new_with_free_func ((GDestroyNotify) (void *) bd_part_spec_free);
 
-    /* the partition gets destroyed together with the disk */
-    ped_disk_destroy (ped_disk);
-    ped_device_destroy (dev);
+    while (fdisk_table_next_partition(table, itr, &pa) == 0) {
```
tbzatek, line 797: missing space :-p

src/plugins/part.c
```diff
@@ -1000,107 +1085,370 @@ static PedPartition* add_part_to_disk (PedDevice *dev, PedDisk *disk, BDPartType
...
+        close_context (cxt);
         return NULL;
     }
 
-    succ = disk_commit (ped_disk, disk, error);
-    if (succ)
-        ret = get_part_spec (dev, ped_disk, ped_part, error);
+    if (fdisk_partition_has_partno (npa)) {
+        if (isdigit (disk[strlen(disk) - 1]))
+            ppath = g_strdup_printf ("%sp%zu", disk, fdisk_partition_get_partno (npa) + 1);
```
tbzatek, line 1437: Okay, I think this deserves more testing on _exotic_ drives: (e)MMC, NVMe, NVMeOF as the device node naming is rather non-traditional there.

src/plugins/part.c
```diff
@@ -2007,6 +2902,9 @@ const gchar* bd_part_get_flag_str (BDPartFlag flag, GError **error) {
     return NULL;
 }
 
+/* string for BD_PART_TYPE_PROTECTED is "primary", that's what parted returns... */
+static const gchar*const part_types[6] = { "primary", "logical", "extended", "free", "metadata", "primary" };
```
tbzatek, line 2906: missing space.

src/plugins/part.c
vojtechtrefny: I think this should be OK for all devices. Yes, there are some weird names for NVMes and similar devices, but this should not affect partition names. The device name + number (or + "p" + number for device names ending with a digit) system should work everywhere -- it works this way even for "non-disk" devices like loop or md.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
I did one last fix and now all tests are passing except on CentOS 8 where we hit #557. I'm merging this now and I'll deal with the resize issue later.

PR #560, opened 2020-08-18
dm: Fix comparing DM RAID member devices UUID

vojtechtrefny:
There is no "UUID" property in UDev we must use the "ID_FS_UUID"
one.
This comparison works only because most DM RAID members don't have
UUID so the check is skipped, but it fails for DDF RAID members
which have a special GUID/UUID in UDev database.

### DIFF
--- a/src/plugins/dm.c
+++ b/src/plugins/dm.c
@@ -482,7 +482,7 @@ static gboolean raid_dev_matches_spec (struct raid_dev *raid_dev, const gchar *n
 
     context = udev_new ();
     device = udev_device_new_from_subsystem_sysname (context, "block", dev_name);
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Sounds a little bit weird since you're now taking UUID of the filesystem inside but this could work theoretically...

vojtechtrefny:
> Sounds a little bit weird since you're now taking UUID of the filesystem inside but this could work theoretically...

This is UUID of the DDF member format. The function is trying to get a DM RAID set for given (potential) member device e.g. looking for RAID name for "sda". It is iterating over all members for every DM RAID set and comparing major/minor and UUID of the member with given device (sda).

vpodzime (approved):
ACK

PR #567, opened 2020-09-21
tests: Fix getting list of NVDIMM devices for tests

vojtechtrefny:
"ndctl list" now outputs the result as a list for single device
and not as dict, see https://github.com/pmem/ndctl/commit/72c46ab194d9987e7b67bba3ab16d020a5a4b10b


tbzatek:
Are you planning to keep backward compatibility with older `ndctl` releases somehow?

vojtechtrefny:
> Are you planning to keep backward compatibility with older `ndctl` releases somehow?

No, this change was introduced in version 62 and it is also available in CentOS 7 so this shouldn't be a problem. And it's only issue in tests, the plugin itself uses the ndctl library.

tbzatek (approved):
Okay then :-)

PR #575, opened 2020-10-02
Add separate tool for VFAT filesystem resize and use it in the FS plugin

vojtechtrefny:
This way we can remove compile time dependency on libparted from
libblockdev and still support VFAT resize.


tools/vfat-resize.c
```diff
@@ -0,0 +1,141 @@
...
+    PedFileSystem *fs = NULL;
+    PedSector start = 0;
+    PedSector length = 0;
+    gint status = 0;
+    guint64 new_size = 0;
+    BSSize bs_size;
+    BSError *bs_error = NULL;
+    gchar *device = NULL;
+
+    if ((g_strcmp0 (argv[1], "-h") == 0) || g_strcmp0 (argv[1], "--help") == 0) {
```
tbzatek, line 63: Huh, in case `argc == 0` accessing `argv[1]` may lead to segfault unless there's a NULL. Still, better to check `argc` first. Also there's an extra parenthesis (cosmetic issue).

tbzatek (commented):
Apart from the minor detail it looks good, certainly a good step forward in terms of licensing compliance.

tools/vfat-resize.c
vojtechtrefny: I think `argc` will always be at least 1 and `argv` is guaranteed to be NULL terminated so this couldn't happen, but I fixed it anyway to be safe.

tbzatek (approved):
I see, good, thanks.

src/plugins/fs/vfat.c
```diff
@@ -18,11 +18,11 @@
  */
 
 #include <blockdev/utils.h>
-#include <blockdev/part_err.h>
 #include <check_deps.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <stdio.h>
```
vpodzime, line 25: I don't think this is needed (probably a remnant from some debug printing)

tools/vfat-resize.c
```diff
@@ -0,0 +1,136 @@
...
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <glib.h>
+#include <parted/parted.h>
+#include <bytesize/bs_size.h>
+
+
+void print_usage (const char *cmd) {
+    g_print ("Usage: %s device [size]\n"
```
vpodzime, line 35: This is from `<glib/gprintf.h>` or something like that, isn't it?

vpodzime (commented):
Looks good to me otherwise.

src/plugins/fs/vfat.c
vojtechtrefny: No, this is intentional, we need `stdio` for `sscanf`. I guess we forgot to include it because it was brought in be parted.

tools/vfat-resize.c
vojtechtrefny: No, it's from `glib/gmessages.h` which is included in `glib.h`. You probably think of `g_vasprintf` which needs `glib/gprintf.h` included explicitly.

PR #576, opened 2020-10-05
fs: Fix bd_fs_can_get_free_space name in declaration

vojtechtrefny:
Fortunately #518 is not released yet so it's not an API break. I can't believe I missed this. And also can't believe gcc is ok with this.


tbzatek (approved):
Eh, good one :-)

PR #578, opened 2020-10-08
WIP: Add function for running mkfs with extra options

vojtechtrefny:
This is not even remotely finished, but I'd like to hear your opinions about the API. The goal here is to provide mkfs API for UDisks and Blivet -- both need to be able to run mkfs with extra options (to set label, UUID etc.) but it doesn't really make sense for them to keep list of command line options for all filesystems to run mkfs with. So this API should provide a simple way to run mkfs with these extra options.

Please take a look at the proposed solution and feel free to come up with other/better ideas :-) I'll add some inline comments for parts where I was not sure how to handle something.

Don't bother reviewing the code too closely, I just made sure it compiles, everything else is most likely broken, it's just about the API design.

### DIFF
--- a/src/plugins/fs/ext.c
+++ b/src/plugins/fs/ext.c
@@ -237,6 +237,26 @@ void bd_fs_ext4_info_free (BDFSExt4Info *data) {
     bd_fs_ext2_info_free ((BDFSExt2Info*) data);
 }
 
+BDExtraArg __attribute__ ((visibility ("hidden")))
+**bd_fs_ext2_mkfs_options (BDFsMkfsOptions *options) {
+    BDExtraArg **ret = g_new0 (BDExtraArg*, 5);
+    guint next_arg = 0;
+
+    if (options->label)
+        ret[next_arg++] = bd_extra_arg_new ("-L", options->label);
+
+    if (options->uuid)
+        ret[next_arg++] = bd_extra_arg_new ("-U", options->uuid);
+
+    if (options->dry_run)
+        ret[next_arg++] = bd_extra_arg_new ("-n", "");
+
+    if (options->no_discard)
+        ret[next_arg++] = bd_extra_arg_new ("-E", "nodiscard");
+
+    return ret;
+}
+
 static gboolean ext_mkfs (const gchar *device, const BDExtraArg **extra, const gchar *ext_version, GError **error) {
     const gchar *args[6] = {"mke2fs", "-t", ext_version, "-F", device, NULL};
 
--- a/src/plugins/fs/generic.h
+++ b/src/plugins/fs/generic.h
@@ -10,6 +10,22 @@ gchar* bd_fs_get_fstype (const gchar *device,  GError **error);
 gboolean bd_fs_freeze (const gchar *mountpoint, GError **error);
 gboolean bd_fs_unfreeze (const gchar *mountpoint, GError **error);
 
+typedef enum {
+    BD_FS_MKFS_LABEL     = 1 << 0,
+    BD_FS_MKFS_UUID      = 1 << 1,
+    BD_FS_MKFS_DRY_RUN   = 1 << 2,
+    BD_FS_MKFS_NODISCARD = 1 << 3,
+} BDFsMkfsOptionsFlags;
+
+typedef struct BDFsMkfsOptions {
+    gchar *label;
+    gchar *uuid;
+    gboolean dry_run;
+    gboolean no_discard;
+} BDFsMkfsOptions;
+
+gboolean bd_fs_mkfs (const gchar *device, const gchar *fsname, BDFsMkfsOptions *options, GError **error);
+
 gboolean bd_fs_resize (const gchar *device, guint64 new_size, GError **error);
 gboolean bd_fs_repair (const gchar *device, GError **error);
 gboolean bd_fs_check (const gchar *device, GError **error);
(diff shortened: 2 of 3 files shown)


src/plugins/fs/ext.c
```diff
@@ -237,6 +237,26 @@ void bd_fs_ext4_info_free (BDFSExt4Info *data) {
     bd_fs_ext2_info_free ((BDFSExt2Info*) data);
 }
 
+BDExtraArg __attribute__ ((visibility ("hidden")))
+**bd_fs_ext2_mkfs_options (BDFsMkfsOptions *options) {
```
vojtechtrefny, line 241: I made this function "private" but it can also be public. I'm not sure if it would be useful or not.

src/plugins/fs/generic.h
```diff
@@ -25,6 +41,7 @@ typedef enum {
     BD_FS_ONLINE_GROW = 1 << 4
 } BDFsResizeFlags;
 
+gboolean bd_fs_can_mkfs (const gchar *type, BDFsMkfsOptionsFlags *options, gchar **required_utility, GError **error);
```
vojtechtrefny, line 44: I don't like this function. I made it this way because this is how `bd_fs_can_resize` works with `BDFsResizeFlags` but I think it might be too complicated to call a function just to check whether a filesystem supports running mkfs with label. Both UDisks and Blivet need to check this and having the "flags" available directly somewhere might be more useful.

vojtechtrefny:
Some relevant parts of UDisks and Blivet code dealing with mkfs and parameters: [UDisks FSInfo](https://github.com/storaged-project/udisks/blob/5b026c27dec4cded3624921b755dc6496efadfb0/src/udiskslinuxfsinfo.c#L65), [Blivet FSMkfs](https://github.com/storaged-project/blivet/blob/3.3-devel/blivet/tasks/fsmkfs.py#L193)

src/plugins/fs/generic.h
vpodzime: Why don't we provide a function that would return the options/flags for the given FS? Instead of calling `bd_fs_can_mkfs()` with all the options/flags.

src/plugins/fs/generic.h
```diff
@@ -10,6 +10,22 @@ gchar* bd_fs_get_fstype (const gchar *device,  GError **error);
...
+    BD_FS_MKFS_DRY_RUN   = 1 << 2,
+    BD_FS_MKFS_NODISCARD = 1 << 3,
+} BDFsMkfsOptionsFlags;
+
+typedef struct BDFsMkfsOptions {
+    gchar *label;
+    gchar *uuid;
+    gboolean dry_run;
+    gboolean no_discard;
+} BDFsMkfsOptions;
```
tbzatek, line 25: Just thinking in terms of API stability for the future... would it make sense to add other commonly used filesystem _features_, no matter if they're going to be actually used or not? Extending the `struct` in the future would be difficult.

Thinking about `mkfs`-time options like `xattrs`, `compression`, `encryption`, `zoned` (SMR HDD, NVMe ZNS), `case-insensitive`, etc.

src/plugins/fs/generic.h
tbzatek: Returning capabilities would be one case, checking if the fs tools are actually installed would be another case. OTOH isn't it better just to fail the actual `bd_fs_mkfs()` call with an _unsupported_ error or similar rather than checking beforehand?

src/plugins/fs/generic.h
vojtechtrefny: I'm not sure about this. We could add some unused "reserve" to the struct that would allow us to add new features without changing the struct size.

src/plugins/fs/generic.h
vojtechtrefny: We can do both -- fail, but also allow users to check for supported features. And both UDisks and Blivet needs the "check feature" feature so we need it.

vojtechtrefny:
> So `bd_fs_mkfs()` is supposed to be a generic entry point for all `mkfs`. How do you plan to provide extra arguments from callers?

I wat thinking about adding the "extra" parameter for `bd_fs_mkfs`, but I see `bd_fs_mkfs` more as a helper function for people that don't want to specify the parameters manually so it doesn't really makes sense to me to allow adding parameters manually -- if you want to do this, you should use the "type specific" mkfs function.

tbzatek:
> I wat thinking about adding the "extra" parameter for `bd_fs_mkfs`, but I see `bd_fs_mkfs` more as a helper function for people that don't want to specify the parameters manually so it doesn't really makes sense to me to allow adding parameters manually -- if you want to do this, you should use the "type specific" mkfs function.

Hmm, that kinda beats usefulness of the new API. A common basic `bd_fs_mkfs` would be useful for sure, however I would like to get rid of UDisks `FSInfo` in an ideal case. Reusing the newly defined struct/db might be the key.

One more use case would be allowing extra options to be passed in `org.freedesktop.UDisks2.Block.Format()` method, further propagated down to whatever `bd_fs_` call needed. You still need some logic e.g. for where in the string to put the extra options, escaping, etc.

tbzatek:
TODO (tbzatek): Add `reiser4` and `reiser5`

src/plugins/fs/generic.h
vpodzime: I think it makes most sense to return the capabilities for the given FS. It can potentially depend on other things than just the tools available (implementation, version of libblockdev, *configure* options,...).

src/plugins/fs/generic.h
vpodzime: Adding some reserve would be an option. Listing all the possible options in advance is impossible, OTOH, I think.

vojtechtrefny:
Closing, replaced by #583

PR #581, opened 2020-11-03
misc: Update Vagrantfile

vojtechtrefny:


tbzatek (approved):
Good I guess... you may also want to replace `yum` with `dnf` for CentOS8, although it's an alias anyway.

PR #582, opened 2020-11-04
tests: Remove MDTestNominateDenominateActive test case

vojtechtrefny:
This test never worked correctly. It creates array with a spare,
stops the array and then tries to activate it by adding only the
spare to it. The result is a "half active" array that can't be
stopped using its name which means the array is not properly
removed during cleanup and causes other tests to fail.
Basic (de)nominate functionallity is already covered by the
MDTestNominateDenominate test case so this can be safely removed.


tbzatek (approved):
Fine with me.

PR #585, opened 2020-11-13
UDF support

vojtechtrefny:
This also contains #583, I'll rebase it after #583 is merged. The UDF support is quite simple (UDF doesn't support resizing and doesn't have fsck), only complicated thing is mkfs, see the comment below.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -517,6 +517,7 @@ typedef enum {
     BD_FS_TECH_NILFS2,
     BD_FS_TECH_EXFAT,
     BD_FS_TECH_BTRFS,
+    BD_FS_TECH_UDF,
 } BDFSTech;
 
 typedef enum {
@@ -787,6 +788,82 @@ GType bd_fs_btrfs_info_get_type () {
         type = g_boxed_type_register_static("BDFSBtrfsInfo",
                                             (GBoxedCopyFunc) bd_fs_btrfs_info_copy,
                                             (GBoxedFreeFunc) bd_fs_btrfs_info_free);
+    }
+
+    return type;
+}
+
+/**
+ * BDFSUdfInfo:
+ * @label: label of the filesystem
+ * @uuid: uuid of the filesystem
+ * @revision: UDF revision
+ * @lvid: Logical Volume Identifier (most UDF implementations use this identifier as a disk label)
+ * @vid: Volume Identifier
+ * @block_size: block size used by the filesystem
+ * @block_count: number of blocks in the filesystem
+ * @free_blocks: number of free blocks in the filesystem
+ */
+typedef struct BDFSUdfInfo {
+    gchar *label;
+    gchar *uuid;
+    gchar *revision;
+    gchar *lvid;
+    gchar *vid;
+    guint64 block_size;
+    guint64 block_count;
+    guint64 free_blocks;
+} BDFSUdfInfo;
+
+/**
+ * bd_fs_udf_info_free: (skip)
+ * @data: (allow-none): %BDFSUdfInfo to free
+ *
+ * Frees @data.
+ */
+void bd_fs_udf_info_free (BDFSUdfInfo *data) {
+    if (data == NULL)
+        return;
+
+    g_free (data->label);
+    g_free (data->uuid);
+    g_free (data->revision);
+    g_free (data->lvid);
+    g_free (data->vid);
+    g_free (data);
+}
+
+/**
+ * bd_fs_udf_info_copy: (skip)
+ * @data: (allow-none): %BDFSUdfInfo to copy
+ *
+ * Creates a new copy of @data.
+ */
+BDFSUdfInfo* bd_fs_udf_info_copy (BDFSUdfInfo *data) {
+    if (data == NULL)
+        return NULL;
+
+    BDFSUdfInfo *ret = g_new0 (BDFSUdfInfo, 1);
+
+    ret->label = g_strdup (data->label);
+    ret->uuid = g_strdup (data->uuid);
+    ret->revision = g_strdup (data->revision);
+    ret->lvid = g_strdup (data->lvid);
+    ret->vid = g_strdup (data->vid);
+    ret->block_size = data->block_size;
+    ret->block_count = data->block_count;
+    ret->free_blocks = data->free_blocks;
+
+    return ret;
+}
+
+GType bd_fs_udf_info_get_type () {
+    static GType type = 0;
+
+    if (G_UNLIKELY(type == 0)) {
+        type = g_boxed_type_register_static("BDFSUdfInfo",
+                                            (GBoxedCopyFunc) bd_fs_udf_info_copy,
+                                            (GBoxedFreeFunc) bd_fs_udf_info_free);
     }
 
     return type;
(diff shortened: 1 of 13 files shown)


src/plugins/fs/udf.c
```diff
@@ -0,0 +1,494 @@
...
+ */
+gboolean bd_fs_udf_mkfs (const gchar *device, const BDExtraArg **extra, GError **error) {
+    const gchar *args[7] = {"mkudffs", "--utf8", NULL, NULL, NULL, NULL, NULL};
+    gint blocksize = -1;
+    g_autofree gchar *bs_str = NULL;
+
+    if (!check_deps (&avail_deps, DEPS_MKUDFFS_MASK, deps, DEPS_LAST, &deps_check_lock, error))
+        return FALSE;
+
+    if (!extra) {
```
vojtechtrefny, line 215: UDF is first filesystem where we need to specify some extra arguments to make it work. These "defaults" are from UDisks and I'm not sure if we want to allow users to specify their own values and how. Options:

- Add extra arguments to `bd_fs_udf_mkfs` -- I don't like this, I would like to keep all `bd_fs_xxx_mkfs` function the same.
- Add second mkfs function, someting like `bd_fs_udf_mkfs_full`
- Do what I did here -- we have some defaults and user can override them with the `extra` argument, but needs to specify everything manually.

tbzatek: For the first point is there a strong reason to keep argument consistency with other filesystems now that we have the generic `bd_fs_mkfs()`? Adding an argument at least for the `--media-type` might still benefit from the rest of the _"automatic"_ options. I.e. to change just the media type the caller wouldn't need to have knowledge of the argument structure.

The reason why I personally consider media type (and anonymous vs. full ownerships) important is a wide usage of UDF for DVD and BD media where ISO9660 + RockRidge don't support files over 2GB. However I have no idea how exactly the burning softwarez do create the UDF filesystem and what options do they provide. I could still imagine this use case for both `libblockdev` and `udisks`.

However, I'm also fine with the suggestion \#2.

At the end combination of \#2 and \#3 seems like a good way forward, if you insist on \#1.

src/lib/plugin_apis/fs.api
```diff
@@ -2313,4 +2390,97 @@ gboolean bd_fs_exfat_check_label (const gchar *label, GError **error);
...
+/**
+ * bd_fs_udf_mkfs:
+ * @device: the device to create a new UDF fs on
+ * @extra: (allow-none) (array zero-terminated=1): extra options for the creation (right now
+ *                                                 passed to the 'mkudffs' utility)
+ * @error: (out): place to store error (if any)
+ *
+ * Note: This functions by default calls 'mkudffs' with '--media-type=hd --udfrev=0x201
+ *       --blocksize=<detect>' to make sure older versions of mkudffs work correctly.
+ *       If you specify additional functions additional options using the @extra
```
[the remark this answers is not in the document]
vojtechtrefny: Fixed.

src/plugins/fs/udf.c
vojtechtrefny: I changed `bd_fs_udf_mkfs`, it now takes `media_type`, `revision` and `block_size` as arguments. `bd_fs_mkfs` made it more complicated because it adds everything as extra arguments so it overwrites our defaults and changing `bd_fs_udf_mkfs` now seems to be lesser evil than changing `bd_fs_mkfs` behaviour or adding some more notes to the documentation. I still don't like it, but it's probably better this way.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,503 @@
...
+gboolean bd_fs_udf_mkfs (const gchar *device, const gchar *media_type, gchar *revision, guint64 block_size, const BDExtraArg **extra, GError **error) {
+    const gchar *args[7] = {"mkudffs", "--utf8", NULL, NULL, NULL, device, NULL};
+    gint detected_bs = -1;
+    gboolean ret = FALSE;
+
+    if (!check_deps (&avail_deps, DEPS_MKUDFFS_MASK, deps, DEPS_LAST, &deps_check_lock, error))
+        return FALSE;
+
+    if (block_size != 0)
+        args[2] = g_strdup_printf ("--blocksize=%"G_GUINT64_FORMAT"", block_size);
```
tbzatek, line 213: I appreciate your futureproof attitude for huge block sizes :snowman:

tbzatek (approved):
Looks good now, thanks for the changes.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,503 @@
...
+    const BDExtraArg **extra_p = NULL;
+    g_autofree gchar *vid = NULL;
+
+    if (options->label) {
+        /* there are different limits for lvid and vid and we need to set both so lets just
+           truncate vid (GParted does the same thing) */
+        if (g_str_is_ascii (options->label) && strlen (options->label) > 30)
+            vid = g_strndup (options->label, 30);
+        else if (g_utf8_validate (options->label, -1, NULL) && g_utf8_strlen (options->label, -1) > 15)
+            vid = g_utf8_substring (options->label, 0, 15);
```
pali, line 146: Hello! You have there incorrect checks in validation. See restrictions in: https://github.com/storaged-project/udisks/issues/381

src/plugins/fs/udf.c
pali: Teoretical maximal block size for UDF is 32768. But kernel supports block size up to the page size, so on x86 just 4096 and maybe on itanium with 8k pages it could be 8192.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,503 @@
...
+
+    if (!check_deps (&avail_deps, DEPS_UDFLABEL_MASK, deps, DEPS_LAST, &deps_check_lock, error))
+        return FALSE;
+
+    /* there are different limits for lvid and vid and we need to set both so lets just
+       truncate vid (GParted does the same thing) */
+    if (g_str_is_ascii (label) && strlen (label) > 30)
+        vid = g_strndup (label, 30);
+    else if (g_utf8_validate (label, -1, NULL) && g_utf8_strlen (label, -1) > 15)
+        vid = g_utf8_substring (label, 0, 15);
```
pali, line 279: Same here, incorrect checks for validation.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,503 @@
...
+ * Returns: whether @label is a valid label for the UDF file system or not
+ *          (reason is provided in @error)
+ *
+ * Tech category: always available
+ */
+gboolean bd_fs_udf_check_label (const gchar *label, GError **error) {
+    if (g_str_is_ascii (label)) {
+        if (strlen (label) > 126) {
+            g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_LABEL_INVALID,
+                        "Label for UDF filesystem must be at most 126 ASCII characters long.");
```
pali, line 308: Why _at most 126 ASCII characters_? In UDF label can contain at most 126 UNICODE code points if all code points are up to the U+FF. Or 63 UNICODE code points if there is at least one code point above U+FF.

vojtechtrefny: Oh, I can't read :facepalm: I read it as 126 ascii and 63 unicode. Ok, back to the drawing board.

vojtechtrefny:
I pushed new version and I hope the label limits are now finally correct. Sorry it took me so long.

pali:
Great! I quickly looked at the code and seems to be fine now, thanks!

vojtechtrefny:
Jenkins, test this please.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,567 @@
...
+static gchar* get_vid (const gchar *label) {
+    gchar *vid = NULL;
+    const gchar *next_p = NULL;
+    gunichar unichar;
+    guint pos = 0;
+
+    if (!g_utf8_validate (label, -1, NULL))
+        return NULL;
+
+    if (g_utf8_strlen (label, -1) < 15)
```
tbzatek, line 144: This should be `<= 15` I suppose.

tbzatek (approved):
Required great deal of caffeine to go through the index counting.

Please check my two minor notes, otherwise it looks good.

src/plugins/fs/udf.c
```diff
@@ -0,0 +1,567 @@
...
+        vid = g_strdup (label);
+    else {
+        /* vid can be at most 30 characters (or 15 > 0xFF) so we'll truncate the label if needed */
+        next_p = label;
+        while (next_p && *next_p && *next_p != '\0') {
```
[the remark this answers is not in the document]
vojtechtrefny: Fixed.

src/plugins/fs/udf.c
vojtechtrefny: Fixed.

Issue #587, opened 2020-11-15
e2fsck exit code 1 is not failure


dtsund-umich:
I was using gnome-disk-utility (gnome-disk-utility 3.36.1
UDisks 2.8.4 (built against 2.8.4)), which uses this library, to repair an ext4 partition.  The repair was successful, but it was reported as a failure because e2fsck ended with exit code 1.  But exit code 1 from e2fsck means "file system errors corrected", not failure; exit code 4 indicates that the file system had errors that were not fixed. Telling gnome-disk-utility to repair the filesystem a second time resulted in a report of success, presumably because e2fsck returned an exit code of 0 the second time.

I reported this as a bug for gnome-disk-utility, and was told it was a bug in this library, so now I'm reporting it here.

vojtechtrefny:
Thank you for the report, fixed in #588 I'll try to do a new minor release soon.

PR #588, opened 2020-11-18
fs: Do not report error when errors were fixed by e2fsck

vojtechtrefny:
Fixes: #587

### DIFF
--- a/src/plugins/fs/ext.c
+++ b/src/plugins/fs/ext.c
@@ -410,15 +410,32 @@ static gboolean ext_repair (const gchar *device, gboolean unsafe, const BDExtraA
     const gchar *args_progress[7] = {"e2fsck", "-f", unsafe ? "-y" : "-p", "-C", "1", device, NULL};
     const gchar *args[5] = {"e2fsck", "-f", unsafe ? "-y" : "-p", device, NULL};
     gint status = 0;
+    gboolean ret = FALSE;
 
     if (!check_deps (&avail_deps, DEPS_E2FSCK_MASK, deps, DEPS_LAST, &deps_check_lock, error))
         return FALSE;
 
     if (bd_utils_prog_reporting_initialized ()) {
-        return bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
+        ret = bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
     } else {
-        return bd_utils_exec_and_report_status_error (args, extra, &status, error);
+        ret = bd_utils_exec_and_report_status_error (args, extra, &status, error);
     }
+
+    if (!ret) {
+        if (status == 1) {
+            /* no error should be reported for exit code 1 - File system errors corrected */
+            g_clear_error (error);
+            ret = TRUE;
+        } else if (status == 2) {
+            /* no error should be reported for exit code 2 - File system errors corrected, system should be rebooted */
+            bd_utils_log_format (BD_UTILS_LOG_WARNING,
+                                 "File system errors on %s were successfully corrected, but system reboot is advised.",
+                                 device);
+            g_clear_error (error);
+            ret = TRUE;
+        }
+    }
+    return ret;
 }
 
 /**


vojtechtrefny:
Question: What should we do about "File system errors corrected, system should be rebooted"? I don't think we should completely ignore this but showing error also doesn't sound like the best thing to do. Maybe a special error state that UDisks (and GNOME Disks) can catch and show just an "info" message to the user?

tbzatek:
> Question: What should we do about "File system errors corrected, system should be rebooted"? I don't think we should completely ignore this but showing error also doesn't sound like the best thing to do. Maybe a special error state that UDisks (and GNOME Disks) can catch and show just an "info" message to the user? @tbzatek @vpodzime

Hmm, I think this is typically designed for init systems. Right now we don't have a way to display a hint, we can log it however (through `g_syslog()` or so) and let some potential (currently nonexisting) storage notification daemon to pick it up. I would not fail the operation as it breaks the semantic of the `gboolean` return value and not every consumer is watching for error codes.

Still, in case `e2fsck` returns such code, would kernel allow mounting that fs afterwards? This should be probably handled later at the mount process, `e2fsck` is only changing data on a block device, unless it can operate on a mounted filesystem.

src/plugins/fs/ext.c
```diff
@@ -410,15 +410,22 @@ static gboolean ext_repair (const gchar *device, gboolean unsafe, const BDExtraA
...
-        return bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
+        ret = bd_utils_exec_and_report_progress (args_progress, extra, extract_e2fsck_progress, &status, error);
     } else {
-        return bd_utils_exec_and_report_status_error (args, extra, &status, error);
+        ret = bd_utils_exec_and_report_status_error (args, extra, &status, error);
     }
+
+    if (!ret && (status == 1)) {
+        /* no error should be reported for exit code 1 - File system errors corrected */
+        g_clear_error (error);
```
tbzatek, line 426: Then `ret` should be set accordingly.

vojtechtrefny: Good catch, fixed.

tbzatek (approved):

vojtechtrefny:
> Right now we don't have a way to display a hint, we can log it however (through `g_syslog()` or so) and let some potential (currently nonexisting) storage notification daemon to pick it up.

I was thinking about returning special error from libblockdev so UDisks can return a special error for GNOME Disks to catch.

> Still, in case `e2fsck` returns such code, would kernel allow mounting that fs afterwards? This should be probably handled later at the mount process, `e2fsck` is only changing data on a block device, unless it can operate on a mounted filesystem.

`e2fsck` can run on mounted devices but it's not recommend. Not sure how `mount` behaves when this happens.

vpodzime (approved):
Looks good to me. I think the reboot hint is outside of the scope of the code here. Like @tbzatek suggested, it could be logged so that some future storage monitoring system grabs it and acts on it, but that's all. And I wouldn't bother with that now. If such storage monitoring system ever exists, it should probably be *e2fsck* itself communicating with it somehow (through the log messages, for example).

tbzatek:
> I was thinking about returning special error from libblockdev so UDisks can return a special error for GNOME Disks to catch.

Take it from a pragmatic point of view: is the filesystem fixed? It is.

Returning an error (and `False` return code) would create unnecessary confusion to legacy consumers. If we set `error` and return `True`, it would create new memleaks. Besides that would qualify as an ABI break, though it's not that important here I guess.

vojtechtrefny:
> > I was thinking about returning special error from libblockdev so UDisks can return a special error for GNOME Disks to catch.
> 
> Take it from a pragmatic point of view: is the filesystem fixed? It is.
> 
> Returning an error (and `False` return code) would create unnecessary confusion to legacy consumers. If we set `error` and return `True`, it would create new memleaks. Besides that would qualify as an ABI break, though it's not that important here I guess.

So I should change the code to return `True` when reboot is needed? Because right now it returns `False` and sets the error. (The change fixed only `1 - File system errors corrected` and not `2    - File system errors corrected, system should be rebooted`.)

tbzatek:
> So I should change the code to return `True` when reboot is needed? Because right now it returns `False` and sets the error. (The change fixed only `1 - File system errors corrected` and not `2 - File system errors corrected, system should be rebooted`.)

Yes, I think so. Both the `1` and `2` return codes should be treated equally as _"filesystem has been fixed"_ and return `True` IMO.

vpodzime:
> Yes, I think so. Both the `1` and `2` return codes should be treated equally as _"filesystem has been fixed"_ and return `True` IMO.

Yup, agreed.

vojtechtrefny:
> > So I should change the code to return `True` when reboot is needed? Because right now it returns `False` and sets the error. (The change fixed only `1 - File system errors corrected` and not `2 - File system errors corrected, system should be rebooted`.)
> 
> Yes, I think so. Both the `1` and `2` return codes should be treated equally as _"filesystem has been fixed"_ and return `True` IMO.

Ok, I've updated the PR.

vpodzime (approved):
ACK

PR #592, opened 2020-11-23
lvm: Fix bd_lvm_vdopooldata_* symbols

tbzatek:
There was a discrepancy around struct BDLVMVDOPooldata related functions
between the API file, C header and C source code, likely a typo with
bd_lvm_vdodata_ name prefix. The public API lvm.h that is shipped publicly
defines the correct bd_lvm_vdopooldata_ symbol prefix however any call
to this function results in an unresolved symbol in runtime.

### DIFF
--- a/docs/libblockdev-sections.txt
+++ b/docs/libblockdev-sections.txt
@@ -368,8 +368,6 @@ bd_lvm_vdo_pool_convert
 bd_lvm_vdo_pool_create
 bd_lvm_vdo_pool_resize
 bd_lvm_vdo_resize
-bd_lvm_vdodata_copy
-bd_lvm_vdodata_free
 bd_lvm_vdopooldata_copy
 bd_lvm_vdopooldata_free
 BDLVMTech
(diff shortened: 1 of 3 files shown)


tbzatek:
We should really do something with the code duplication between the API files and the real code. It has been error prone for quite some time and we're still discovering new copy-paste errors.

This should be backported to the `libblockdev-2.x` branch and new release should be made.

vojtechtrefny (approved):

vojtechtrefny:
> We should really do something with the code duplication between the API files and the real code. It has been error prone for quite some time and we're still discovering new copy-paste errors.

I'm open to suggestions but I'm not sure we can do something that would be better.

We could eliminate the potential for more copy-paste errors by adding C API tests. The problem really is only with the functions that are not tested from Python.

tbzatek:
> I'm open to suggestions but I'm not sure we can do something that would be better.

I was thinking about something like adding preprocessor directives to the API file that would then generate the clean public header file. Avoiding code duplication and only filter out some blocks. Or perhaps move the specific part to a separate included file. Needs more elaboration...

Issue #552, opened 2020-07-23
Consider using libtrfsutil in the btrfs plugin


vojtechtrefny:
[libbtrfsutil](https://github.com/kdave/btrfs-progs/tree/master/libbtrfsutil) was added in 2018 (btrfs-progs 4.16), we should consider using it in the btrfs plugin instead of calling btrfs tools and parsing their output. It looks like it supports everything we need (but we should take a closer look first to be 100 % sure).

jelly:
One downside I noticed here is that libbtrsutil is not actually used by the `btrfs` binary, but it is used by `snapper` it probably doesn't break a lot but worth to check if it has unit tests.

If libblockdev wants to extend btrfs information, then libbrtfsutil needs to be extended.

tbzatek:
> libbtrsutil is not actually used by the `btrfs` binary

Isn't it linked statically? Sometimes distros tend to link everything statically for initramfs use.

> If libblockdev wants to extend btrfs information, then libbrtfsutil needs to be extended.

Theoretically we may opt for a mixed use - library calls for most of the stuff and external tools as a fallback for missing functionality. Assuming both the library and the tools are part of a single package.

PR #595, opened 2020-11-24
Btrfs support in filesystem plugin

vojtechtrefny:
First two commits are from #583 and #585.

The idea here is to provide functions for users that want to use btrfs as a filesystem and not as a volume manager. This implementation also uses only the command line tools so there are no compile time dependencies (as opposed to the btrfs plugin which we plan to rewrite to use libbtrfsutil library, see #552).

Few highlights:
- `resize` and `get_info` functions will fail when called on filesystem that is part of a multidevice btrfs volume. This is indented and we want users to use btrfs plugin for these. Other functions do not check number of btrfs devices and should generally work, but we'll encourage people to not use these on multidevice btrfs volumes.
- Some functions (`resize`, `set_label`...) take moutpoint as an argument and not the device. We already do this for some xfs functions and it probably makes more sense than taking device that must be mounted. The generic functions for resize and other actions that require the device to be mounted will temporarily mount the device if necessary.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -516,6 +516,7 @@ typedef enum {
     BD_FS_TECH_REISERFS,
     BD_FS_TECH_NILFS2,
     BD_FS_TECH_EXFAT,
+    BD_FS_TECH_BTRFS,
 } BDFSTech;
 
 typedef enum {
@@ -725,6 +726,67 @@ GType bd_fs_exfat_info_get_type () {
         type = g_boxed_type_register_static("BDFSExfatInfo",
                                             (GBoxedCopyFunc) bd_fs_exfat_info_copy,
                                             (GBoxedFreeFunc) bd_fs_exfat_info_free);
+    }
+
+    return type;
+}
+
+/**
+ * BDFSBtrfsInfo:
+ * @label: label of the filesystem
+ * @uuid: uuid of the filesystem
+ * @size: size of the filesystem in bytes
+ * @free_space: free space on the filesystem in bytes
+ */
+typedef struct BDFSBtrfsInfo {
+    gchar *label;
+    gchar *uuid;
+    guint64 size;
+    guint64 free_space;
+} BDFSBtrfsInfo;
+
+/**
+ * bd_fs_btrfs_info_copy: (skip)
+ * @data: (allow-none): %BDFSBtrfsInfo to copy
+ *
+ * Creates a new copy of @data.
+ */
+BDFSBtrfsInfo* bd_fs_btrfs_info_copy (BDFSBtrfsInfo *data) {
+    if (data == NULL)
+        return NULL;
+
+    BDFSBtrfsInfo *ret = g_new0 (BDFSBtrfsInfo, 1);
+
+    ret->label = g_strdup (data->label);
+    ret->uuid = g_strdup (data->uuid);
+    ret->size = data->size;
+    ret->free_space = data->free_space;
+
+    return ret;
+}
+
+/**
+ * bd_fs_btrfs_info_free: (skip)
+ * @data: (allow-none): %BDFSBtrfsInfo to free
+ *
+ * Frees @data.
+ */
+void bd_fs_btrfs_info_free (BDFSBtrfsInfo *data) {
+    if (data == NULL)
+        return;
+
+    g_free (data->label);
+    g_free (data->uuid);
+    g_free (data);
+}
+
+GType bd_fs_btrfs_info_get_type () {
+    static GType type = 0;
+
+    if (G_UNLIKELY(type == 0)) {
+        type = g_boxed_type_register_static("BDFSBtrfsInfo",
+                                            (GBoxedCopyFunc) bd_fs_btrfs_info_copy,
+                                            (GBoxedFreeFunc) bd_fs_btrfs_info_free);
     }
 
     return type;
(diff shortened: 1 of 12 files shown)


src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ * Tech category: %BD_FS_TECH_BTRFS-%BD_FS_TECH_MODE_SET_LABEL
+ */
+gboolean bd_fs_btrfs_set_label (const gchar *mpoint, const gchar *label, GError **error);
+
+/**
+ * bd_fs_btrfs_check_label:
+ * @label: label to check
+ * @error: (out) (allow-none): place to store error
+ *
+ * Returns: whether @label is a valid label for the UDF file system or not
```
tbzatek, line 2621: UDF?

src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ *
+ */
+gboolean bd_fs_btrfs_mkfs (const gchar *device, const BDExtraArg **extra, GError **error);
+
+/**
+ * bd_fs_btrfs_wipe:
+ * @device: the device to wipe a UDF signature from
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the UDF signature was successfully wiped from the @device or
```
tbzatek, line 2567: UDF?

src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ */
+gboolean bd_fs_btrfs_repair (const gchar *device, const BDExtraArg **extra, GError **error);
+
+/**
+ * bd_fs_btrfs_set_label:
+ * @mpoint: the mount point of the file system to resize
+ * @label: label to set
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the label of UDF file system on the @mpoint was
```
tbzatek, line 2606: :expressionless:

src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ */
+gboolean bd_fs_btrfs_check_label (const gchar *label, GError **error);
+
+/**
+ * bd_fs_btrfs_set_uuid:
+ * @device: the device containing the file system to set the UUID (serial number) for
+ * @uuid: (allow-none): UUID to set or %NULL to generate a new one
+ * @error: (out): place to store error (if any)
+ *
+ * Returns: whether the UUID of the UDF file system on the @device was
```
tbzatek, line 2637: :cry:

src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ * Returns: whether a new btrfs fs was successfully created on @device or not
+ *
+ * Tech category: %BD_FS_TECH_BTRFS-%BD_FS_TECH_MODE_MKFS
+ *
+ */
+gboolean bd_fs_btrfs_mkfs (const gchar *device, const BDExtraArg **extra, GError **error);
+
+/**
+ * bd_fs_btrfs_wipe:
+ * @device: the device to wipe a UDF signature from
```
tbzatek, line 2564: :flushed:

src/lib/plugin_apis/fs.api
```diff
@@ -2483,4 +2545,151 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+ * Tech category: %BD_FS_TECH_BTRFS-%BD_FS_TECH_MODE_SET_UUID
+ */
+gboolean bd_fs_btrfs_set_uuid (const gchar *device, const gchar *uuid, GError **error);
+
+/**
+ * bd_fs_btrfs_check_uuid:
+ * @uuid: UUID to check
+ * @error: (out) (allow-none): place to store error
+ *
+ * Returns: whether @uuid is a valid UUID for the UDF file system or not
```
tbzatek, line 2652: :tired_face:

src/plugins/fs/generic.c
```diff
@@ -523,6 +524,108 @@ static gboolean nilfs2_resize_device (const gchar *device, guint64 new_size, GEr
...
 }
 
+static BDFSBtrfsInfo* btrfs_get_info (const gchar *device, GError **error) {
+    g_autofree gchar* mountpoint = NULL;
+    gboolean unmount = FALSE;
+    gboolean ret = FALSE;
+    GError *local_error = NULL;
+    BDFSBtrfsInfo* btrfs_info = NULL;
+
+    mountpoint = fs_mount (device, "btrfs", &unmount, error);
```
tbzatek, line 534: Hint: we should be using separate mount namespace for this, someday.

src/plugins/fs/generic.c
```diff
@@ -523,6 +524,108 @@ static gboolean nilfs2_resize_device (const gchar *device, guint64 new_size, GEr
...
+    if (unmount) {
+        ret = bd_fs_unmount (mountpoint, FALSE, FALSE, NULL, &local_error);
+        if (!ret) {
+            if (btrfs_info) {
+                /* info was successful but unmount failed */
+                g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_UNMOUNT_FAIL,
+                             "Failed to unmount '%s' after getting info: %s",
+                             device, local_error->message);
+                g_clear_error (&local_error);
+                return FALSE;
```
tbzatek, line 549: Leaking `btrfs_info`?

tbzatek (commented):
_indented_ -> _intended_

Needs some work...

src/plugins/fs/generic.c
vojtechtrefny: Fixed.

src/lib/plugin_apis/fs.api
vojtechtrefny: Yes, I decided to rename btrfs to UDF, to make it even more confusing :-)

tbzatek (approved):
Good to go, don't forget to reorder the mkfs extra options...

vojtechtrefny:
Rebased to latest master and removed the UDF commit. UDF PR needs to be fixed so we can merge this first.

PR #596, opened 2020-11-25
lvm: Fix bd_lvm_vdopooldata_* symbols

vojtechtrefny:
There was a discrepancy around struct BDLVMVDOPooldata related functions
between the API file, C header and C source code, likely a typo with
bd_lvm_vdodata_ name prefix. The public API lvm.h that is shipped publicly
defines the correct bd_lvm_vdopooldata_ symbol prefix however any call
to this function results in an unresolved symbol in runtime.

------

#592 cherry picked to 2.x-branch


tbzatek (approved):
Good job :satisfied:

PR #598, opened 2020-11-27
fs: Wipe all signatures when running wipe with wipe_all set to TRUE

vojtechtrefny:
When wiping filesystem signatures from a device, we are not trying
to wipe all signatures, but only signatures matching the filesystem
name -- so for example all "ntfs" signatures.
This makes sense for FAT with multiple "vfat" signatures, but
NTFS has one "ntfs" signature and one nested "dos" partition table
which we also need to remove.
This change makes sure all signatures (no only filesystem ones and
not only these matching the filesystem name) are wiped from the
device.

-----

I'm working on some filesystem plugin test changes and I noticed there are some missing tests for NTFS so I added those and found this issue. I'm posting this for a separate review because it's quite big change and it can potentially break something. Alternative approach would be to change this behaviour only for NTFS (the `wipe_fs` function is internal so we can add an extra argument for this specific use case).

### DIFF
--- a/src/plugins/fs/common.c
+++ b/src/plugins/fs/common.c
@@ -71,6 +71,18 @@ has_fs (blkid_probe probe, const gchar *device, const gchar *fs_type, GError **e
     return TRUE;
 }
 
+/**
+ * wipe_fs: (skip)
+ * @device: the device to wipe signature(s) from
+ * @fs_type: (allow-none): filesystem type to wipe from the device or %NULL for any
+ * @wipe_all: whether to wipe all signatures from the device or not: some filesystems might have
+ *            multiple nested signatures, this will make sure to wipe all of them including
+ *            those that are different from @fs_type -- e.g. for NTFS we will wipe both the
+ *            NTFS signature and the nested DOS partition table created by mkfs.ntfs
+ * @error: (out): place to store error
+ *
+ * Returns: whether the wipe was successful or not
+ */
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Good catch.

What about issuing BLKDISCARD as well? Most filesystems do that automatically, a little help might not harm perhaps.

src/plugins/fs/common.c
```diff
@@ -192,7 +192,7 @@ wipe_fs (const gchar *device, const gchar *fs_type, gboolean wipe_all, GError **
     blkid_reset_probe (probe);
 
     if (wipe_all) {
-        has_fs_type = has_fs (probe, device, fs_type, error);
+        has_fs_type = has_fs (probe, device, NULL, error);
```
vpodzime, line 195: The name of the variable no longer makes sense. And maybe the whole code can be simplified?

vpodzime (changes requested):

src/plugins/fs/common.c
vojtechtrefny: Good point, fixed.

vpodzime (approved):
Looks good to me now. Just please check that the documentation is still valid for the new behavior.

vojtechtrefny:
> Looks good to me now. Just please check that the documentation is still valid for the new behavior.

This is an internal documentation so there wasn't a docstring but I added one with the `wipe_all` explanation for future reference.

PR #599, opened 2020-11-27
Filesystems tests update

vojtechtrefny:
A little refactoring of `fs_test.py`. Mostly just moving the code around, but I've also added some new tests for NTFS and `bd_fs_is_tech_available`.

Note: The new NTFS wipe test is expected to fail, see #598


tbzatek (approved):
Just a quick look, sanity check.

PR #600, opened 2020-12-03
LVM thin metadata calculation fix

vojtechtrefny:
We are currently using `thin_metadata_size` to calculate metadata size for thin pools but this tools uses different method to calculate the recommended metadata size than lvcrete which creates bigger metadata by default. This changes the calculation to match the lvcreate default metadata size.


tbzatek (approved):
Nice to see an external dependency removal!

PR #604, opened 2020-12-23
Kernel keyring support for crypto plugin

vojtechtrefny:
Just two simple functions to add key to the keyring and open LUKS device using it.


tbzatek (approved):
Nice!

tbzatek (approved):

PR #605, opened 2021-01-07
tests: Include btrfs_test in tests/fs_tests/__init__.py

vojtechtrefny:
Without this the btrfs test suite won't be run when running all
tests from fs_tests.


tbzatek (approved):
Wasn't that by purpose? :sweat_smile:

PR #607, opened 2021-01-08
misc: Add ansible playbook to install test dependencies

vojtechtrefny:
and use it in our Vagrantfile.


tbzatek (approved):
Just a sanity check, didn't go through the package lists.

Issue #609, opened 2021-01-20
LUKS and integrity "info" functions should work with both the backing and cleartext device


vojtechtrefny:
Currently both `bd_crypto_luks_info` and `bd_crypto_integrity_info` needs the cleartext/mapped device to work because they use the `crypt_init_by_name` function. Ideally these would work also when called with the backing device. There is no reason to require the LUKS/integrity device to be unlocked, all the information we need are available from the LUKS/integrity metadata.

PR #610, opened 2021-02-02
WIP: Crypto plugin API cleanup

vojtechtrefny:
This contain two changes:
 - We now require cryptsetup >= 2.0.3, LUKS 2 specific functions have been removed (format and resize).
 - `_blob` functions were removed.

The former shouldn't cause any problems, but the latter one make things messy and I'd like to hear your opinions.

We currently have two sets of functions with different ways to take passphrase "input". The "normal" functions have both `passphrase` (`const gchar *`) and `key_file` (also `const gchar *`) parameters. The newer `_blob` functions accept only passphrase as `pass_data` (`const guint8 *`). Having two functions for everything makes the API too complicated and we should remove the `_blob` functions IMHO. The question is what parameters should we except now. Our options as I see them:

 1. Both passphrase and key file: This what I did, but I'm not really happy about it. Especially the functions that take two sets of passphrases are now crazy complicated with 6 parameters just for passphrases.
 2. Forget key files. This means we'll be forcing our users to read the passphrase from a keyfile themselves adding one more place where it can leak.
 3. Create more functions. We could add new special functions for key files (and potentially other options), similar to `crypt_activate_by_XYZ` which cryptsetup uses.
 4. Add a custom struct or union as a "passphrase parameter" and use it for different "types" of passphrases.

Few things to keep in mind:
- We currently have two options: passphrase (string or binary) and key file, but this might not be enough and we should at least consider adding [kernel keyring](https://github.com/storaged-project/libblockdev/pull/604) as a third option.
- We have two functions that need two passphrases: `change_key` and `add_key`.
- What should we do if user specifies two passphrases (for example one using passphrase option and second using key file). Should we ignore the second one or use both where it makes sense (e.g. adding two keyslots in format)?

I'm currently leaning towards the option 4, but I'd like to hear your options @tbzatek @vpodzime 

Few unrelated notes:
- I made separate commits for every changed/removed functions, I'll squash it in the final version, this is just to make it a little easier to read/review.
- Few more functions will be changed (`bd_crypto_tc_open_full` for example, but these are not "controversial").
- I also plan to do something about the "info" functions (see #609) and probably remove `bd_crypto_luks_uuid`, `bd_crypto_luks_get_metadata_size` and `bd_crypto_luks_status` and add the missing values to `BDCryptoLUKSInfo`.


tbzatek:
I guess the question is whether to support multiple/various combinations of credentials of various types. Think of a analogy to multifactor authorization. Not sure if `cryptsetup` supports it now already or if there's a chance for `cryptsetup` 3.0+ to ever support that. The key is to design the `libblockdev` API flexible rather now than later.

On the other hand I'm in favour of simple convenient functions where it's clear from the first sight what are their purpose. Introducing a C `union` or other kind of an universal list of data is perhaps too complicated for consumers. So my favourite solution is \#3, followed by \#1.

Basically the changes you've made look good enough to me by scrolling through. It's rather clear what the transition should look like and the new API is not complicated.

vpodzime:
I think having a `struct BDCryptoLUKSCredentials` or something like that makes most sense. Adding separate parameters for all the possible ways to pass the secret to the function and twice the number for the changing functions seems quite crazy to me. And if we want to, the `struct` can actually have something like `guint32 methods_set` as the first field serving as a bitmask flag for which fields are set, allowing us to add more things to the struct in the future. The key is to have a proper initializer for the struct.

PR #611, opened 2021-02-02
Python 2 support removed

vojtechtrefny:
https://pythonclock.org/ :-)


tbzatek (approved):
Great cleanup. Would you please take care of a similar stuff in udisks, after it's branched for 2.10?

PR #617, opened 2021-02-22
crypto: Remove bd_crypto_luks_uuid function

vojtechtrefny:
Users should use bd_crypto_luks_info instead.

Separated from #614 so we can merge it without API change.

### DIFF
--- a/docs/3.0-api-changes.xml
+++ b/docs/3.0-api-changes.xml
@@ -28,6 +28,14 @@
 
   </simplesect>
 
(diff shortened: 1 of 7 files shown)


tbzatek (approved):
OK I guess.

vojtechtrefny:
Rebase to the latest master and I did the same trick in the Python overrides to keep the Python API unchanged as in #819

tbzatek (approved):
Always good to see more stuff removed than added!

PR #618, opened 2021-02-22
Adapt to dosfstools 4.2 changes

vojtechtrefny:


tests/fs_tests/vfat_test.py
```diff
@@ -15,6 +18,19 @@ def setUp(self):
 
         self.mount_dir = tempfile.mkdtemp(prefix="libblockdev.", suffix="vfat_test")
 
+        if self._get_dosfstools_version() <= LooseVersion("4.1"):
+            self._mkfs_options = None
+        else:
+            self._mkfs_options = [BlockDev.ExtraArg.new("--mbr=n", "")]
```
vojtechtrefny, line 24: I'm not 100 % convinced this is the best way to deal with the `mkfs.vfat` behaviour change, but I think we shouldn't try to be smarter than dosfstools here and if they think `mkfs.vfat` should create partition table when formatting a disk, we shouldn't change it. So I've decided to adjust tests instead.

tbzatek: Well, I personally consider this an ABI break, even though I understand the reason behind it. (An interactive question would be even worse...).

tbzatek (approved):
Not nice. No nice words for `dosfstools` upstream either.

What can we do... okay for now, having a library for FAT/exFAT would be nice to have some day.

PR #621, opened 2021-03-04
fs: Add API for querying filesystem "features"

vojtechtrefny:
I posting this here to get some feedback, this is third iteration of the API and I'm still not sure if it makes sense.

Users of this API will be both UDisks where some of these values are currently hardcoded in `udiskslinuxfsinfo` and Blivet where we need this in the filesystem related tasks.

Feel free to recommend better names for the "features" and flags, I was always bad at naming things. I'm also not sure if the single "generic" function is enough, maybe we should consider separate functions for the filesystems.

Also I've added test case only for XFS, I'll add test cases for all filesystems in the final version.

### DIFF
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -2508,4 +2508,92 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
  */
 BDFSUdfInfo* bd_fs_udf_get_info (const gchar *device, GError **error);
 
+typedef enum {
+    BD_FS_SUPPORT_SET_LABEL = 1 << 1,
+    BD_FS_SUPPORT_SET_UUID = 1 << 2
+} BDFSConfigureFlags;
+
+typedef enum {
+    BD_FS_FSCK_CHECK = 1 << 1,
+    BD_FS_FSCK_REPAIR = 1 << 2
+} BDFSFsckFlags;
+
+typedef enum {
+    BD_FS_FEATURE_OWNERS  = 1 << 1
+} BDFSFeatureFlags;
+
+/**
+ * BDFSFeatures:
+ * @resize: supported resizes modes
+ * @mkfs: supported options for mkfs
+ * @fsck: support for fsck operations (check and repair)
+ * @configure: support for changing properties of an existing filesystem
+ * @features: other supported features
+ */
+typedef struct BDFSFeatures {
+    BDFsResizeFlags resize;
+    BDFSMkfsOptionsFlags mkfs;
+    BDFSFsckFlags fsck;
+    BDFSConfigureFlags configure;
+    BDFSFeatureFlags features;
+} BDFSFeatures;
+
+/**
+ * bd_fs_features_copy: (skip)
+ * @data: (allow-none): %BDFSFeatures to copy
+ *
+ * Creates a new copy of @data.
+ */
+BDFSFeatures* bd_fs_features_copy (BDFSFeatures *data) {
+    if (data == NULL)
+        return NULL;
+
+    BDFSFeatures *ret = g_new0 (BDFSFeatures, 1);
+
+    ret->resize = data->resize;
+    ret->mkfs = data->mkfs;
+    ret->fsck = data->fsck;
+    ret->configure = data->configure;
+    ret->features = data->features;
+
+    return ret;
+}
+
+/**
+ * bd_fs_features_free: (skip)
+ * @data: (allow-none): %BDFSFeatures to free
+ *
+ * Frees @data.
+ */
+void bd_fs_features_free (BDFSFeatures *data) {
+    if (data == NULL)
+        return;
+
+    g_free (data);
+}
+
+GType bd_fs_features_get_type () {
+    static GType type = 0;
+
+    if (G_UNLIKELY(type == 0)) {
+        type = g_boxed_type_register_static("BDFSFeatures",
+                                            (GBoxedCopyFunc) bd_fs_features_copy,
+                                            (GBoxedFreeFunc) bd_fs_features_free);
+    }
+
+    return type;
+}
+
+/**
+ * bd_fs_features:
+ * @fstype: name of the filesystem to get features for (e.g. "ext4")
+ * @error (allow-none): (out): place to store error (if any)
+ *
+ * Returns (transfer-none): features supported by @fstype, see %BDFSFeatures for more information.
+ *
+ * Tech category: always available
+ *
+ */
+const BDFSFeatures* bd_fs_features (const gchar *fstype, GError **error);
+
 #endif  /* BD_FS_API */
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -50,6 +50,77 @@ typedef enum {
     BD_FS_GET_FREE_SPACE,
 } BDFsOpType;
 
+static const BDFSFeatures fs_features[BD_FS_LAST_FS] = {
+    /* padding for BD_FS_TECH_GENERIC and MOUNT to make accessing the FS techs simpler */
+    { 0 }, { 0 },
+    /* EXT2 */
+    { .resize = BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK,
+      .mkfs = BD_FS_MKFS_LABEL | BD_FS_MKFS_UUID | BD_FS_MKFS_DRY_RUN | BD_FS_MKFS_NODISCARD,
+      .fsck = BD_FS_FSCK_CHECK | BD_FS_FSCK_REPAIR,
+      .configure = BD_FS_SUPPORT_SET_LABEL | BD_FS_SUPPORT_SET_UUID,
+      .features =  BD_FS_FEATURE_OWNERS },
+    /* EXT3 */
+    { .resize = BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK,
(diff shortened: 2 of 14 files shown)


src/plugins/fs/generic.c
```diff
@@ -51,6 +51,84 @@ typedef enum {
     BD_FS_GET_FREE_SPACE,
 } BDFsOpType;
 
+    BDFSFsckFlags fsck;
+    BDFSConfigureFlags configure;
+    BDFSFeatureFlags features;
```
vpodzime, line 56: What is this? A copy pasta?

src/plugins/fs/generic.c
```diff
@@ -1520,3 +1598,46 @@ gboolean bd_fs_mkfs (const gchar *device, const gchar *fstype, BDFSMkfsOptions *
...
+ * @error (allow-none): (out): place to store error (if any)
+ *
+ * Returns (transfer-none): features supported by @fstype, see %BDFSFeatures for more information.
+ *
+ * Tech category: always available
+ *
+ */
+const BDFSFeatures* bd_fs_features (const gchar *fstype, GError **error) {
+    if (g_strcmp0 (fstype, "exfat") == 0) {
+        return &fs_features[BD_FS_TECH_EXFAT - BD_FS_OFFSET];
```
vpodzime, line 1615: I think it would be nicer to have a function that would return `BD_FS_TECH_EXFAT` for `"exfat"` and use it here and possibly in other places.

vpodzime (commented):
I don't quite understand the difference between `BDFSConfigureFlags configure;` `BDFSFeatureFlags features;`. Or, I do understand the difference, but aren't the possible modification operations also features? I'd personally merge these two.

Looks good to me otherwise.

src/lib/plugin_apis/fs.api
```diff
@@ -2698,4 +2698,92 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+} BDFSConfigureFlags;
+
+typedef enum {
+    BD_FS_FSCK_CHECK = 1 << 1,
+    BD_FS_FSCK_REPAIR = 1 << 2
+} BDFSFsckFlags;
+
+typedef enum {
+    BD_FS_FEATURE_OWNERS  = 1 << 1
+} BDFSFeatureFlags;
```
tbzatek, line 2713: I vaguely remember we were talking about flags for xattrs, quota, compression, dedup, multidisk, ... ? Does it fit here?

src/plugins/fs/generic.c
```diff
@@ -76,18 +154,18 @@ typedef struct BDFSInfo {
...
-    {"reiserfs", "mkfs.reiserfs", BD_FS_MKFS_LABEL | BD_FS_MKFS_UUID, "reiserfsck", "reiserfsck", "resize_reiserfs", BD_FS_ONLINE_GROW | BD_FS_OFFLINE_GROW | BD_FS_OFFLINE_SHRINK, "reiserfstune", "debugreiserfs", "reiserfstune"},
-    {"nilfs2", "mkfs.nilfs", BD_FS_MKFS_LABEL | BD_FS_MKFS_DRY_RUN | BD_FS_MKFS_NODISCARD, NULL, NULL, "nilfs-resize", BD_FS_ONLINE_GROW | BD_FS_ONLINE_SHRINK, "tune-nilfs", "tune-nilfs", "tune-nilfs"},
-    {"exfat", "mkfs.exfat", BD_FS_MKFS_LABEL, "fsck.exfat", "fsck.exfat", NULL, 0, "tune.exfat", "tune.exfat", NULL},
-    {"btrfs", "mkfs.btrfs", BD_FS_MKFS_LABEL | BD_FS_MKFS_UUID | BD_FS_MKFS_NODISCARD, "btrfsck", "btrfsck", "btrfs", BD_FS_ONLINE_GROW | BD_FS_ONLINE_SHRINK, "btrfs", "btrfs", "btrfstune"},
-    {"udf", "mkudffs", BD_FS_MKFS_LABEL | BD_FS_MKFS_UUID, NULL, NULL, NULL, 0, "udflabel", "udfinfo", "udflabel"},
+    {"xfs", "mkfs.xfs", fs_features[BD_FS_TECH_XFS - BD_FS_OFFSET].mkfs, "xfs_db", "xfs_repair", "xfs_growfs", fs_features[BD_FS_TECH_XFS - BD_FS_OFFSET].resize, "xfs_admin", "xfs_admin", "xfs_admin"},
```
tbzatek, line 157: Umm, can't you just define `fs_features[]` with first two (`BD_FS_OFFSET`) elements dummy with appropriate comments provided? Just a visual cleanup.

src/plugins/fs/generic.c
tbzatek: macros?

tbzatek (commented):
I totally forgot what was discussed in previous iterations so I may repeat myself at places :-)

tbzatek:
Another thing worth considering in the light of recent events: provide a flag to indicate that mkfs might create partition table and unspecified number of partitions. It would give a hint to either re-read the partition table or wait for partitions to be probed.

However this is not universal, tools may implement complex logic for such decision and we don't certainly want to duplicate that logic in `libblockdev`. Even though often it's just a single partition or none at all.

src/plugins/fs/generic.c
vojtechtrefny: I'm surprised it compiled. It was probably just some random accidental Ctrl+C/Ctrl+V. Fixed.

src/lib/plugin_apis/fs.api
```diff
@@ -2698,4 +2698,92 @@ gboolean bd_fs_udf_check_uuid (const gchar *uuid, GError **error);
...
+
+/**
+ * BDFSFeatures:
+ * @resize: supported resizes modes
+ * @mkfs: supported options for mkfs
+ * @fsck: support for fsck operations (check and repair)
+ * @configure: support for changing properties of an existing filesystem
```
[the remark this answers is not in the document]
vojtechtrefny: The `configure` flags are meant for existing filesystems. Some filesystem can set label during `mkfs` but not change it later.

src/lib/plugin_apis/fs.api
vojtechtrefny: Yes. That's why I added a separate flag here. My idea was `configure` to be for things that can be changed for an existing filesystem and `features` for "features" (don't know how to explain it better). But I agree with @vpodzime there is a thin line between these two. *compression* is a feature but it's also something that can be configured. So `features` should be for things that are not "normal".

src/plugins/fs/generic.c
vojtechtrefny: I can, but is it really necessary?

src/plugins/fs/generic.c
vpodzime: I still think that function would make this nicer.

vpodzime (commented):
Looks good to me otherwise. I think this is good to go as an API definition.

src/plugins/fs/generic.c
vojtechtrefny: Yes, I ignored this one for now, I wanted to fix the "obvious" errors first. Are you thinking about a public function/macro or just something for internal use?

src/plugins/fs/generic.c
```diff
@@ -50,6 +50,76 @@ typedef enum {
     BD_FS_GET_FREE_SPACE,
 } BDFsOpType;
 
+static const BDFSFeatures fs_features[BD_FS_LAST_FS] = {
+    { 0 }, { 0 },
```
vpodzime, line 54: It would be nice to have the explanation of the two `{ 0 }`s here.

tbzatek:
There's one more thing I am not fully sure how to put in this API: ability to control the `mkfs` tool to create/don't create a protective MBR table. See this crazy quirk: https://github.com/storaged-project/udisks/commit/ef1979d8cbfb8ad976de0af21a70f8c0f98fbbe9

Some tools do that by default, some don't. Also depends on certain version. For UDisks we need to control whether to create the protective table or not and also (for UDF filesystem?) we need to find out whether it's on by default or not... i.e. a tool version check.

I expect more quirks for other filesystems may appear in the future. Do we want some API for that?

src/plugins/fs/generic.c
vojtechtrefny: Ok, I have added a comment for that.

vojtechtrefny:
> There's one more thing I am not fully sure how to put in this API: ability to control the `mkfs` tool to create/don't create a protective MBR table.

That will probably be a new `BDFSMkfsOptions` mkfs option. Lets deal with this in a separate PR.

> I expect more quirks for other filesystems may appear in the future. Do we want some API for that?

We can definitely add more flags where it makes sense. I'd like to avoid adding a new "features" if it means only one filesystem supports it, this should be more generic API. And if plan to add some more advanced features for for example f2fs I would probably consider adding a new plugin and do what we did with btrfs -- multidevice and volume management is in a separate plugin and fs plugin supports btrfs only as a "simple" filesystem.

tbzatek:
> That will probably be a new `BDFSMkfsOptions` mkfs option. Lets deal with this in a separate PR.

OK, so TODO for you:
- [ ] add `force` option
- [ ] add something like`no-protective-parttable` option

tbzatek:
Another aspect relevant to the filesystem definitions: need MBR/GPT partition type ID/UUID suggested for each of the filesystem type.

https://github.com/storaged-project/udisks/blob/6eba6d2454bd87ae1e3b4945a3d84b9a215d9014/src/udiskslinuxblock.c#L2510-L2539

vojtechtrefny:
I've reported issues for the new features/APIs discussed above, let's get this merged so we can start working on using this in UDisks (and possibly blivet) to make sure we have (or track if missing) all the functionality they need.

PR #622, opened 2021-03-08
NVMe plugin - phase 1

tbzatek:
Introducing a new `nvme` plugin based on `libnvme`: https://github.com/linux-nvme/libnvme

For the moment only simple controller and namespace queries are provided. Future plans include namespace operations (format, secure erase) and controller operations (health self-tests, namespace add/del/modify/attach/detach), possibly also NVMe over Fabrics methods.


tbzatek:
This won't compile due to missing `libnvme`. For the moment I'm linking statically using the following hack to override `pkg-config` definitions:
`PKG_CONFIG_PATH="path/libnvme" NVME_CFLAGS="-Ipath/libnvme/src" NVME_LIBS="path/libnvme/src/libnvme.a" ./configure ...`

As a first step please check the API, need to sort out the library availability with their upstream.

src/plugins/nvme/nvme-error.c
```diff
@@ -0,0 +1,315 @@
...
+
+
+static inline const gchar * _nvme_generic_status_to_string (guint status)
+{
+    /*
+     * Generic Command Status Codes:
+     */
+    switch (status) {
+    case NVME_SC_INVALID_OPCODE:       /* 0x1 */
+        return "INVALID_OPCODE: The associated command opcode field is not valid";
```
vojtechtrefny, line 55: Do you think it would possible to get these error strings directly from libnvme? It looks like a lot of internal stuff to have in libblockdev. We did something similar for mount errors in the fs plugin and kzak agreed to add a "get error string" function for us.

src/plugins/nvme/nvme.h
```diff
@@ -0,0 +1,327 @@
...
+    gint selftest_ext_time;
+    guint64 hmb_pref_size;
+    guint64 hmb_min_size;
+    guint64 size_total;
+    guint64 size_unalloc;
+    gboolean feat_sanitize_crypto;
+    gboolean feat_sanitize_block;
+    gboolean feat_sanitize_overw;
+    gboolean feat_secure_erase_all;
+    gboolean feat_secure_erase_crypto;
```
vojtechtrefny, line 93: It might be better to make all these "features" flags instead of booleans. Thoughts?

tests/nvme_test.py
```diff
@@ -0,0 +1,156 @@
...
+            BlockDev.init(cls.requested_plugins, None)
+        else:
+            BlockDev.reinit(cls.requested_plugins, True, None)
+
+
+class NVMeTestCase(NVMeTest):
+    def setUp(self):
+        if not find_executable("nvme"):
+            raise unittest.SkipTest("nvme executable (nvme-cli package) not found in $PATH, skipping.")
+        if not find_executable("nvmetcli"):
```
vojtechtrefny, line 25: Please add both `nvme-cli` and `nvmetcli` to the [install-test-dependencies](https://github.com/storaged-project/libblockdev/blob/master/misc/install-test-dependencies.yml) playbook.

vojtechtrefny (commented):
I didn't go through all the code, but the API looks good in general.

src/plugins/nvme/nvme-error.c
tbzatek: Posted upstream: https://github.com/linux-nvme/libnvme/pull/19

src/plugins/nvme/nvme.h
tbzatek: Changed to enums.

tests/nvme_test.py
tbzatek: Done, good idea. Debian doesn't ship `nvmetcli`, tests won't run.

tbzatek:
Superseded by #653.

PR #623, opened 2021-03-16
vdo: Do not use g_memdup in bd_vdo_stats_copy

vojtechtrefny:
g_memdup is deprecated and the replacement g_memdup2 is not yet
available so lets just do the copy manually.

-----

We don't use `g_memdup` on master but 3.0 is still far away so we need to fix so 2.x compiles without the deprecation errors.


tbzatek:
I'm using `memcpy()` for copying structures, only manually dup'ing pointers where necessary.

vojtechtrefny:
> I'm using `memcpy()` for copying structures, only manually dup'ing pointers where necessary.

We already have this code on master in LVM. I was lazy and just copy-pasted it here :-)

PR #624, opened 2021-03-16
vfat: Add check for invalid characters in bd_fs_vfat_check_label()

tbzatek:
Backported from UDisks: udiskslinuxfilesystem.c:handle_set_label()

### DIFF
--- a/src/plugins/fs/vfat.c
+++ b/src/plugins/fs/vfat.c
@@ -133,9 +133,14 @@ BDExtraArg __attribute__ ((visibility ("hidden")))
 **bd_fs_vfat_mkfs_options (BDFSMkfsOptions *options, const BDExtraArg **extra) {
     GPtrArray *options_array = g_ptr_array_new ();
     const BDExtraArg **extra_p = NULL;
+    gchar *label;
 
-    if (options->label)
-        g_ptr_array_add (options_array, bd_extra_arg_new ("-n", options->label));
+    if (options->label) {
+        /* convert the label uppercase */
+        label = g_ascii_strup (options->label, -1);
+        g_ptr_array_add (options_array, bd_extra_arg_new ("-n", label));
+        g_free (label);
+    }
 
     if (options->uuid)
         g_ptr_array_add (options_array, bd_extra_arg_new ("-i", options->uuid));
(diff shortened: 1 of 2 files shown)


src/plugins/fs/vfat.c
```diff
@@ -288,12 +288,26 @@ gboolean bd_fs_vfat_set_label (const gchar *device, const gchar *label, GError *
...
+
     if (strlen (label) > 11) {
         g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_LABEL_INVALID,
                      "Label for VFAT filesystem must be at most 11 characters long.");
         return FALSE;
     }
 
+    /* VFAT does not allow some characters; as dosfslabel does not enforce this,
+     * check in advance; also, VFAT only knows upper-case characters, dosfslabel
+     * enforces this */
```
vojtechtrefny, line 302: > VFAT only knows upper-case characters, dosfslabel enforces this

`dosfslabel` only warns about it. Maybe we should be stricter here and force uppercase only?

```
$ sudo dosfslabel /dev/loop1 aaaa
fatlabel: warning - lowercase labels might not work properly on some systems
```

vojtechtrefny (approved):
Looks good. I took the label "limits" from blivet, maybe we should check all filesystems to be sure these are actually valid.

src/plugins/fs/vfat.c
tbzatek: Right, though I'm not sure if it's worth enforcing that - i.e. failing the check. UDisks always converts FAT labels uppercase without complaining.

Pushed a second commit that always converts labels uppercase, please check. The `bd_fs_vfat_mkfs()` function still allows specifying arbitrary arguments, added a docs note.

src/plugins/fs/vfat.c
```diff
@@ -274,7 +282,14 @@ gboolean bd_fs_vfat_set_label (const gchar *device, const gchar *label, GError *
             args[2] = "--reset";
     }
 
-    return bd_utils_exec_and_report_error (args, NULL, error);
+    /* forcefully convert the label uppercase */
+    if (!new_vfat && label)
```
vojtechtrefny, line 286: Why only for dosfstools < 4.2? You convert it for both the old and new version in `bd_fs_vfat_mkfs_options`.

tbzatek: That's a logical conundrum. `new_vfat` is only set when dosfstools >= 4.2 AND the label is empty.

vojtechtrefny: Oh, I see. But I think adding a new `is_empty`  boolean would make this much easier to read.

tbzatek: Attempt #2, please check.

vojtechtrefny (approved):
Looks good, thanks.

PR #630, opened 2021-04-27
Adapt to dosfstools 4.2 changes

vojtechtrefny:
Backport of #618, 2.x will be with us for some time so we need the fix here too.


tbzatek (approved):
good :-)

PR #631, opened 2021-04-29
kbd: Fix memory leak

vojtechtrefny:

PR #632, opened 2021-04-29
Memory leaks fixes backport

vojtechtrefny:
Backport of memoery leak fixes from #631 and #571


tbzatek (approved):
Good. I think've reviewed the code earlier already :-)

PR #641, opened 2021-07-01
exec: Fix deprecated glib function call

krieghof:
Glib will rename "g_spawn_check_exit_status()" to "g_spawn_check_wait_status()"
in version 2.69.

### DIFF
--- a/src/utils/exec.c
+++ b/src/utils/exec.c
@@ -229,7 +229,12 @@ gboolean bd_utils_exec_and_report_status_error (const gchar **argv, const BDExtr
     /* g_spawn_sync set the status in the same way waitpid() does, we need
        to get the process exit code manually (this is similar to calling
        WEXITSTATUS but also sets the error for terminated processes */
-    if (!g_spawn_check_exit_status (exit_status, &l_error)) {
+
+    #if !GLIB_CHECK_VERSION(2, 69, 0)
+    #define g_spawn_check_wait_status(x) (g_spawn_check_exit_status (x))
+    #endif
+
+    if (!g_spawn_check_wait_status (exit_status, &l_error)) {
         if (g_error_matches (l_error, G_SPAWN_ERROR, G_SPAWN_ERROR_FAILED)) {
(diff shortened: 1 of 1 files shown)


blivetghoul:

tbzatek:
Jenkins, ok to test.

tbzatek (approved):
Reference: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1967

Technically I'm okay with this, I'll leave it up to @vojtechtrefny for a stylistic consideration.

src/utils/exec.c
```diff
@@ -229,7 +229,12 @@ gboolean bd_utils_exec_and_report_status_error (const gchar **argv, const BDExtr
     /* g_spawn_sync set the status in the same way waitpid() does, we need
        to get the process exit code manually (this is similar to calling
        WEXITSTATUS but also sets the error for terminated processes */
-    if (!g_spawn_check_exit_status (exit_status, &l_error)) {
+
+    #if !GLIB_CHECK_VERSION(2, 69, 0)
+    #define g_spawn_check_wait_status(x) (g_spawn_check_exit_status (x))
```
tbzatek, line 234: This will need to take two arguments.

tbzatek (changes requested):

vojtechtrefny (approved):
Looks good to me now, thank you.

tbzatek (approved):
If @vojtechtrefny is okay with the `x,y` arguments, then I'm fine with this as well :-)

pothos:
Hello, would be nice if someone can put this into the `2.x-branch`. Currently building UDisks and libblockdev fails because both hardcode `-Wall` and UDisks `master` requires the old branch because it is still using the old wipe function.

vojtechtrefny:
> Hello, would be nice if someone can put this into the `2.x-branch`.

Ok. We already have quite a few backported bugfixes, so I'll check if we need something else from master and do a new upstream release next week.

tbzatek:
Alright, let's make `udisks-2.9.3` later next week.

PR #642, opened 2021-07-08
tests: Fix checking for XFS shrink support in tests

vojtechtrefny:
Support for shrinking was added in xfsprogs 5.12 not 5.1.12


tbzatek (approved):
:hand:

PR #644, opened 2021-07-09
Add new argument for bd_fs_wipe/clean to control force wipe

### LINKED ISSUE #483: bd_fs_wipe happily wipes mounted filesystems

vojtechtrefny:
Our current implementation allows to wipe mounted devices by
default, we should let users decide whether they want it or not.

Fixes: #483 

Related #484 and https://github.com/storaged-project/udisks/pull/901


tbzatek (approved):
Great, thanks, I'll tweak the UDisks part myself.

PR #645, opened 2021-07-09
configure.ac: Remove invalid email address for bug reports

vojtechtrefny:
Sorry @vpodzime :-) I've added project URL but I'm not sure how often is this actually used (AFAICT it's visible only in `configure --help`) so it might not be really useful.


tbzatek (approved):

vpodzime (approved):
:cry: ACK

PR #646, opened 2021-07-13
Support for LVM devices file

vojtechtrefny:
LVM 2.03.12 introduces a new config file `/etc/lvm/devices/system.devices`, all devices that LVM should have access to must be specified there (except for `pvcreate` that can work on any device and will add it to the file). The problem for blivet is that it currently heavily uses the `--config devices { filter=[""] }'` device filtering that is no longer supported, the behaviour of the filter is partially replaced by the `--devices` option where we can specify devices for LVM to use (this bypasses the config file).

This is a work in progress change for blivet with changes limited to the CLI plugin (we'll probably need to add the `lvmdevices` command to LVM DBus API, not sure about that yet) and I will probably change it a lot based on the progress of the blivet code for this new "feature".


tbzatek (approved):
Looks good!

PR #648, opened 2021-07-14
fs: Reset blkid probe after bd_fs_wipe() all signatures

tbzatek:
In rare circumstances not all signatures are wiped during
a single blkid probe pass. The probe needs to be reset and
another pass made.

Reproducible with util-linux-2.36.2, dosfstools-4.2, protective
MBR on unpartitioned block device and having the filesystem on
a nested partition freshly unmounted. Happens when wipe is
not forced (i.e. exclusive open not used) and wiping all signatures.

### DIFF
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -123,6 +123,7 @@ gboolean bd_fs_wipe (const gchar *device, gboolean all, gboolean force, GError *
     gchar *msg = NULL;
     guint n_try = 0;
     gint mode = 0;
+    gboolean reprobed = FALSE;
 
     msg = g_strdup_printf ("Started wiping signatures from the device '%s'", device);
     progress_id = bd_utils_report_started (msg);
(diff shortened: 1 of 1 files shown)


tbzatek:
This happens during UDisks `test_80_filesystem.VFATTestCase.test_protective_part_overwrite_mounted` tests (see https://github.com/storaged-project/udisks/pull/901) where the master block device is attempted to be erased right after a nested (protective) partition has just been unmounted.

No idea what was really going on, perhaps a delayed signature write but since it doesn't happen when wiping with exclusive lock held, my wild guess is that something is poking the device while the probe/erase process is in progress, possibly as a result of a uevent generated.

vojtechtrefny (changes requested):
I don't like this, the `blkid_reset_probe` really shouldn't be needed, did you report this to util-linux? I'm not strictly against adding a workaround to libblockdev, but I think we should try fixing libblkid first (if this is something that could be fixed there).

Anyway, I plan to refactor `bd_fs_wipe` soon (we have two different implementation of *wipefs* in libblockdev) so this definitely needs a test case to make sure we don't brake UDisks in the future.

tbzatek:
Possibly related: https://github.com/util-linux/util-linux/issues/2228

PR #649, opened 2021-07-22
README: Add information about 3.0 and branches

vojtechtrefny:
We already started merging some backward incompatible changes so we should address this in the README.


tbzatek (approved):
Agreed, good idea.

PR #650, opened 2021-07-22
exec: Fix deprecated glib function call

vojtechtrefny:
Glib will rename "g_spawn_check_exit_status()" to "g_spawn_check_wait_status()"
in version 2.69.

Backport of #641


tbzatek (approved):
That pedantic `-Werror`... :-)

PR #654, opened 2021-08-20
overrides: Fix translating exceptions in ErrorProxy

vojtechtrefny:
This fixes exception translation for plugins with multiple
custom exceptions/xrules (currently only swap plugin uses this
functionallity). Currently only the last xrule added to the
ErrorProxy is considered, this fix makes sure all rules are checked
and correct exception is raised.

### DIFF
--- a/tests/swap_test.py
+++ b/tests/swap_test.py
@@ -8,6 +8,7 @@ from gi.repository import BlockDev, GLib
 
 class SwapTest(unittest.TestCase):
     requested_plugins = BlockDev.plugin_specs_from_names(("swap",))
+    dev_size = 1024**3
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


tests/swap_test.py
```diff
@@ -110,6 +111,18 @@ def test_swapon_pagesize(self):
...
 
+    def test_swapon_wrong_size(self):
+        """Verify that activating swap with a wrong size fails with expected exception"""
+
+        # create swap bigger that the device (twice as big in 1024 sectors)
+        ret, out, err = run_command("mkswap -f %s %d" % (self.loop_dev, (self.dev_size * 2) / 1024))
+        if ret != 0:
+            self.fail("Failed to prepare swap for wrong size test: %s %s" % (out, err))
+
+        # activation should fail because swap is bigger that the underlying device
```
vpodzime, line 122: *than*

vpodzime (approved):
ACK

PR #656, opened 2021-08-24
tests: Do not check that swap flag is not support on DOS table

vojtechtrefny:
Support for setting the swap flag was added to libparted recently.
This was added as a downstream patch on rawhide so we can't simply
check libparted version for the support so removing this check
entirely is the best solution.


tbzatek (approved):
Likely it should read _"not supported"_ but looks good anyway.

PR #658, opened 2021-08-30
tests: Fix copy-paste in CryptoTestTrueCrypt and CryptoTestBitlk

vojtechtrefny:
The error paths should also use the tc/bitlk specific functions
not the copied luks ones. Also both bitlocker and truecrypt need
password because we don't support key file for those.


tbzatek (approved):
Yeah, looks sane.

PR #659, opened 2021-08-31
lvm: Fix reading statistics for VDO pools with VDO 8

vojtechtrefny:
The statistics are no longer available in /sys/kvdo, in the latest
version of kvdo we need to use /sys/block/<pool>/vdo/statistics.

Resolves: rhbz#1994220

### DIFF
--- a/src/plugins/vdo_stats.c
+++ b/src/plugins/vdo_stats.c
@@ -133,6 +133,23 @@ static void add_computed_stats (GHashTable *stats) {
     add_journal_stats (stats);
 }
 
+static gchar* _dm_node_from_name (const gchar *map_name, GError **error) {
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
The tests will still fail on CentOS 9 because LVM also uses `/sys/kvdo` for `lvs` (and we get `used_percent` and `saving_percent` from `lvs`).

tbzatek (approved):
Looks good. Alternative way of getting sysfs path for any block device would be to query udev db via gudev.

vojtechtrefny:
> Alternative way of getting sysfs path for any block device would be to query udev db via gudev.

UDev in libblockdev? Never!

PR #660, opened 2021-09-01
tests: Lower expected free space on newly created Ext filesystems

vojtechtrefny:
With e2fsprogs 1.46.4 we now see less than 90 % of free blocks on
newly created devices in our tests.


tbzatek (approved):
Hah. Perhaps a higher reserved blocks number in `e2fsprogs` defaults?

PR #663, opened 2021-09-21
Add support for creating and activating integrity devices

vojtechtrefny:
This adds support for create, open and close actions for standalone
integrity devices using cryptsetup.

----- 

I'm creating this as a draft, because I'm still not 100 % sure the API for `integrity_format` and `integrity_open` makes sense and I might change it a bit when working on the blivet part of the integrity support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -191,7 +191,7 @@ AS_IF([test "x$with_crypto" != "xno"],
       AS_IF([$PKG_CONFIG --atleast-version=2.0.3 libcryptsetup],
             [AC_DEFINE([LIBCRYPTSETUP_2])], [])
       AS_IF([$PKG_CONFIG --atleast-version=2.3.0 libcryptsetup],
-            [AC_DEFINE([LIBCRYPTSETUP_BITLK])], [])
+            [AC_DEFINE([LIBCRYPTSETUP_23])], [])
       AS_IF([$PKG_CONFIG --atleast-version=2.4.0 libcryptsetup],
             [AC_DEFINE([LIBCRYPTSETUP_24])], [])
       AS_IF([test "x$with_escrow" != "xno"],
--- a/src/lib/plugin_apis/crypto.api
+++ b/src/lib/plugin_apis/crypto.api
@@ -1,5 +1,6 @@
 #include <glib.h>
 #include <blockdev/utils.h>
+#include <libcryptsetup.h>
 
 #define BD_CRYPTO_LUKS_METADATA_SIZE G_GUINT64_CONSTANT (2097152ULL) // 2 MiB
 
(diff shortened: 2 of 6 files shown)


src/lib/plugin_apis/crypto.api
```diff
@@ -1,5 +1,6 @@
 #include <glib.h>
 #include <blockdev/utils.h>
+#include <libcryptsetup.h>
```
vpodzime, line 3: I wish this was not needed, but I guess it's better than having some magic numbers in the `enum` definition below.

src/plugins/crypto.c
```diff
@@ -53,6 +53,12 @@
 #define DEFAULT_LUKS_KEYSIZE_BITS 256
 #define DEFAULT_LUKS_CIPHER "aes-xts-plain64"
 
+#ifdef LIBCRYPTSETUP_23
+#define DEFAULT_INTEGRITY_TAG_SIZE 0
+#else
+#define DEFAULT_INTEGRITY_TAG_SIZE 4
+#endif
```
vpodzime, line 60: This is worth a comment.

configure.ac
```diff
@@ -191,7 +191,7 @@ AS_IF([test "x$with_crypto" != "xno"],
       AS_IF([$PKG_CONFIG --atleast-version=2.0.3 libcryptsetup],
             [AC_DEFINE([LIBCRYPTSETUP_2])], [])
       AS_IF([$PKG_CONFIG --atleast-version=2.3.0 libcryptsetup],
-            [AC_DEFINE([LIBCRYPTSETUP_BITLK])], [])
+            [AC_DEFINE([LIBCRYPTSETUP_23])], [])
```
vpodzime, line 194: Why not define both?

src/plugins/crypto.c
```diff
@@ -2224,6 +2267,207 @@ BDCryptoLUKSTokenInfo** bd_crypto_luks_token_info (const gchar *device, GError *
 }
 #endif
 
+static int _wipe_progress (guint64 size, guint64 offset, void *usrptr) {
+    gdouble progress = 50 + (((gdouble) offset / size) * 100) / 2;
```
vpodzime, line 2271: Deserves a comment.

src/python/gi/overrides/BlockDev.py
```diff
@@ -329,6 +329,30 @@ def crypto_keyring_add_key(key_desc, key):
 __all__.append("crypto_keyring_add_key")
 
 
+class CryptoIntegrityExtra(BlockDev.CryptoIntegrityExtra):
+    def __new__(cls, sector_size=0, journal_size=0, journal_watermark=0, journal_commit_time=0, interleave_sectors=0, tag_size=0, buffer_sectors=0):
+        ret = BlockDev.CryptoIntegrityExtra.new(sector_size, journal_size, journal_watermark, journal_commit_time, interleave_sectors, tag_size, buffer_sectors)
+        ret.__class__ = cls
```
vpodzime, line 335: Uuuuh, magic!

vpodzime (approved):
Looks good to me otherwise.

configure.ac
vojtechtrefny: I didn't want to add another variable and we already have `LIBCRYPTSETUP_2` and `LIBCRYPTSETUP_24` so `LIBCRYPTSETUP_23` looked like a better choice for both integrity and bitlk.

src/plugins/crypto.c
vojtechtrefny: I've added a comment explaining the defaults. The `4` for older versions of cryptsetup isn't really the best choice (it will work only with crc32, truncating the tag for "longer" hashes like sha256) but I don't want to add another logic for calculating the correct tag size for hashes, that should be up to the caller. (And this probably wouldn't be used with older cryptsetup anyway, I don't plan to backport this to 2.x).

tbzatek (approved):

src/lib/plugin_apis/crypto.api
```diff
@@ -251,6 +252,115 @@ GType bd_crypto_luks_extra_get_type () {
     return type;
 }
 
+#define BD_CRYPTO_TYPE_Integrity_EXTRA (bd_crypto_integrity_extra_get_type ())
```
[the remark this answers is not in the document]
vojtechtrefny: Fixed.

PR #664, opened 2021-10-07
crypto: Include libcryptsetup.h in crypto.h to fix compile errors

vojtechtrefny:
We use some libcryptsetup constants in crypto.h.

----

/me wonders if using the cryptsetup flags directly for `BDCryptoIntegrityOpenFlags` is the best idea. It might be better to "translate" those to the cryptsetup flags in `bd_crypto_integrity_open` and avoid including libcryptsetup.h here and in `crypto.api`. Thoughts?

### DIFF
--- a/src/plugins/crypto.h
+++ b/src/plugins/crypto.h
@@ -1,5 +1,6 @@
 #include <glib.h>
 #include <blockdev/utils.h>
+#include <libcryptsetup.h>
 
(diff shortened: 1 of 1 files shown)


vpodzime:
> Thoughts?

Didn't like it at the first glance in the previous PR. :) I think adding our own constants would be better.

PR #665, opened 2021-10-15
lvm: Report special error when system.devices file is not enabled

vojtechtrefny:
This can be disabled either in LVM by a compile time option or
by a lvm.conf option so we should report a specific error for this
case so users can distinguish between the feature not being enabled
and not being supported at all.

### DIFF
--- a/src/plugins/lvm.c
+++ b/src/plugins/lvm.c
@@ -3439,6 +3439,54 @@ BDLVMVDOStats* bd_lvm_vdo_get_stats (const gchar *vg_name, const gchar *pool_nam
     return stats;
 }
 
+/* check whether the LVM devices file is enabled by LVM
+ * we use the existence of the "lvmdevices" command to check whether the feature is available
+ * or not, but this can still be disabled either in LVM or in lvm.conf
+ */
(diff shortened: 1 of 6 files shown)


src/plugins/lvm.c
```diff
@@ -3439,6 +3439,51 @@ BDLVMVDOStats* bd_lvm_vdo_get_stats (const gchar *vg_name, const gchar *pool_nam
...
+    gint scanned = 0;
+
+    /* try current config first -- if we get something from this it means the feature is
+       explicitly enabled or disabled by system lvm.conf or using the --config option */
+    args[2] = "current";
+    ret = call_lvm_and_capture_output (args, NULL, &output, &loc_error);
+    if (ret) {
+        scanned = sscanf (output, "use_devicesfile=%u", &enabled);
+        if (scanned != 1)
+            return FALSE;
```
tbzatek, line 3461: Leaking `output` - just move the `g_free()` before the `if`. Three more places to fix.

vojtechtrefny: Oops, fixed.

Issue #667, opened 2021-10-29
Duplicate LVs returned from bd_lvm_lvs()


wfpudak:
Hi,

We've noticed that if an LV has several segments as noted by lvdisplay

  --- Logical volume ---
  LV Path                /dev/rootvg/test
  LV Name                test
  VG Name                rootvg
  LV UUID                TgjM5V-jhXX-tPTf-gPMa-c3w6-QEjo-7qkmVg
  LV Write Access        read/write
  LV Creation host, time du1, 2021-10-25 15:59:20 +0000
  LV Status              available
  "#" open                 1
  LV Size                68.00 MiB
  Current LE             68
  **Segments               2**
  Allocation             inherit
  Read ahead sectors     1024
  Block device           253:166

 the API method **bd_lvm_lvs()** will return several entries since **segtype** is used with the **-o** option.

It can easily be reproduced with the lvs command itself as well. This happens due to fragmentation as described in : https://askubuntu.com/questions/380424/understanding-output-of-lvdisplay-in-lvm-linux

Is the expectation from **bd_lvm_lvs()** to return those duplicates or shouldn't the lib code filter them out?

Example lvs output with your flags if there are several segments :

LVM2_VG_NAME=rootvg LVM2_LV_NAME=test LVM2_LV_UUID=TgjM5V-jhXX-tPTf-gPMa-c3w6-QEjo-7qkmVg LVM2_LV_SIZE=71303168 LVM2_LV_ATTR=-wi-ao---- LVM2_SEGTYPE=linear LVM2_ORIGIN= LVM2_POOL_LV= LVM2_DATA_LV= LVM2_METADATA_LV= LVM2_LV_ROLE=public LVM2_MOVE_PV= LVM2_DATA_PERCENT= LVM2_METADATA_PERCENT= LVM2_COPY_PERCENT= LVM2_LV_TAGS=
[...]

PR #671, opened 2021-11-11
lvm: Do not include duplicate entries in bd_lvm_lvs output

vojtechtrefny:
We use "-o segtypes" for the "lvs" command which means multisegment
LVs will be twice in the output.

See #667

### DIFF
--- a/src/plugins/lvm.c
+++ b/src/plugins/lvm.c
@@ -1988,8 +1988,22 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
         if (table && (num_items == 16)) {
             /* valid line, try to parse and record it */
             lvdata = get_lv_data_from_table (table, TRUE);
-            if (lvdata)
-                g_ptr_array_add (lvs, lvdata);
+            if (lvdata) {
+                /* ignore duplicate entries in lvs output, these are caused by multi segments LVs */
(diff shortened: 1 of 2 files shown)


src/plugins/lvm.c
```diff
@@ -1988,8 +1988,22 @@ BDLVMLVdata** bd_lvm_lvs (const gchar *vg_name, GError **error) {
         if (table && (num_items == 16)) {
             /* valid line, try to parse and record it */
             lvdata = get_lv_data_from_table (table, TRUE);
-            if (lvdata)
-                g_ptr_array_add (lvs, lvdata);
+            if (lvdata) {
+                /* ignore duplicate entries in lvs output, these are caused by multi segments LVs */
+                for (gsize i = 0; i < lvs->len; i++) {
+                    if (g_strcmp0 (((BDLVMLVdata *) g_ptr_array_index (lvs, i))->lv_name, lvdata->lv_name) == 0) {
```
tbzatek, line 1994: There's a convenient `g_ptr_array_find_with_equal_func()` function for searching through the array but it wouldn't help much to relieve the clutter I guess.

tbzatek (approved):
Fine with me.

src/plugins/lvm.c
vojtechtrefny: It was added in 2.54 and I didn't want to bump glib required version.

PR #672, opened 2021-11-12
Fix setting "empty" global config for LVM commands

vojtechtrefny:
This fixes issue where using `BlockDev.lvm_set_global_config("")` will cause all commands to be run with `--config` without parameter, so for example for `pvremove` we will call `pvremove --config /dev/sda`. It doesn't make sense to call lvm with empty config, `--config ""` doesn't replace/remove defaults from `/etc/lvm/lvm.conf` so if user set's it to an empty string we should just ignore it and set the internal variable to `NULL`.

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -3000,7 +3000,10 @@ gboolean bd_lvm_set_global_config (const gchar *new_config, GError **error UNUSE
     g_free (global_config_str);
 
     /* now store the new one */
-    global_config_str = g_strdup (new_config);
+    if (!new_config || g_strcmp0 (new_config, "") == 0)
+         global_config_str = NULL;
+    else
+        global_config_str = g_strdup (new_config);
 
     g_mutex_unlock (&global_config_lock);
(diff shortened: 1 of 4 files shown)


tbzatek (approved):
Good catch. What about unreadable/non-existing config file? Would we let `lvm` commands to scream an error?

Still, at the time `bd_lvm_set_global_config()` is called the file don't necessarily need to exist.

vojtechtrefny:
> What about unreadable/non-existing config file?

With `--config` you don't specify a file, but a list of config options to set or override from the default config. I'm not sure what happens when `/etc/lvm/lvm.conf` doesn't exist, but that isn't really our problem.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
The `lvm_dbus_tests.LvmTestLVcreateType.test_lvcreate_type` failure looks like something unrelated. It happens only on Debian where the LVM devices feature is not yet supported and happens also on today's daily build on master. I'm trying to reproduce the issue manually.

PR #674, opened 2021-11-22
vdo_stats: Default to 100 % savings for invalid savings values

vojtechtrefny:
We are currently using "-1" when VDO logical_blocks_used is 0
which doesn't match the LVM logic which returns 100 so to make
both values in vdo_info and vdo_stats we should return 100 too.

-----

Alternatively we could just adjust the test that currently checks that the value we get from `lvs` and our internal calculation of `saving_percent` is the same, but I think that having two different values would be more confusing.

I'm also not sure what and when changed in VDO, because this wasn't an issue before (`logical_blocks_used` had some valid value so we got the same result for both functions). Unfortunately the VDO kernel module as been broken/behind kernel for so long that we don't have a passing build in our CI where VDO tests weren't skipped.


tbzatek (approved):
I see, good point.

PR #677, opened 2021-11-26
lvm: Use "lvmconfig full" to get valid config instead of "current"

vojtechtrefny:
"lvmconfig current" doesn't work together with --config even if we
don't override the "use_devicefile" key. "lvmconfig full" seems to
be working in all cases.

----

This is unfortunately hard to test because there isn't a better way to check whether the `use_devicefile` feature is enabled or not (it's a compile time option and is different even with same versions of LVM on different versions of RHEL/CentOS).


tbzatek (approved):
:-/

PR #680, opened 2021-12-01
tests: Fix creating loop device for CryptoTestLuksSectorSize

vojtechtrefny:
I'm amazed that creating two loop devices backed with a single file and then changing sector size of one of them to 4k actually works, but it probably explains why the test randomly fails when trying change the sector size.


tbzatek (approved):
Oh, nice.

PR #683, opened 2021-12-15
tests: Wait for raid and mirrored LVs to be synced before removing

vojtechtrefny:
The test_lvcreate_type test case randomly fails on slower machines
when trying to remove the newly created mirror and RAID LVs which
is probably caused by the LV still being synced.

Resolves: rhbz#2030647


tbzatek (approved):
Hmm, nasty. Probably good enough for now. Would be nice to see this design shortcoming somehow solved on the lvm side.

tbzatek (approved):
Hah.

PR #684, opened 2021-12-20
tests: Use losetup to create 4k sector size loop device for testing

vojtechtrefny:
When creating the loop device using libblockdev (which doesn't
support specifying the sector size in bd_loop_setup) the loop
device is not yet ready when we try to change its sector size
with losetup which means the test setup fails.

-----

#680 wasn't enough.


tbzatek (approved):
> When creating the loop device using libblockdev (which doesn't support specifying the sector size in bd_loop_setup)

Well, time to break the API for `libblockdev-3.0`!

PR #687, opened 2021-12-29
fs: Return BD_FS_ERROR_NOFS on unknown filesystem

tbzatek:
According to the mount(2) syscall manpage the return code
for an unknown filesystem is ENODEV, not shared with any other
error condition.

### DIFF
--- a/src/plugins/fs/mount.c
+++ b/src/plugins/fs/mount.c
@@ -227,7 +227,7 @@ static gboolean get_mount_error_old (struct libmnt_context *cxt, int rc, MountAr
                     g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
                                  "Filesystem type not specified");
                 else
-                    g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_FAIL,
+                    g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_UNKNOWN_FS,
                                  "Filesystem type %s not configured in kernel.", args->fstype);
                 break;
             case EROFS:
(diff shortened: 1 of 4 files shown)


src/plugins/fs/mount.c
```diff
@@ -343,6 +343,9 @@ static gboolean get_mount_error_new (struct libmnt_context *cxt, int rc, MountAr
         if (permission)
             g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_AUTH,
                          "Operation not permitted.");
+        else if (syscall_errno == ENODEV)
+            g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_NOFS,
```
tbzatek, line 347: I was not sure whether to introduce a new error code, kinda liked the existing `BD_FS_ERROR_NOFS`.

src/plugins/fs/mount.c
```diff
@@ -343,6 +343,9 @@ static gboolean get_mount_error_new (struct libmnt_context *cxt, int rc, MountAr
         if (permission)
             g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_AUTH,
                          "Operation not permitted.");
+        else if (syscall_errno == ENODEV)
```
vojtechtrefny, line 346: You should add this to `get_mount_error_old` too. Nobody is going to use ntfs3 with the old libmount, but we should still make sure the errors are same for both cases.

src/plugins/fs/mount.c
vojtechtrefny: A new error code sounds better. Maybe you could add the new error code only on master and use `BD_FS_ERROR_NOFS` on 2.x?

tbzatek: Not an option, I might need it for a stable udisks branch. Things will need to be consistent.

vojtechtrefny: I meant using `BD_FS_ERROR_NOFS` with libblockdev 2.x and using the new error code with 3.0. To avoid ABI changes and new error code on "stable" libblockdev. But if you think this is better, I'm ok with it.

tbzatek: When backporting to 2.x branch, the new error code would need to be specified as `BD_FS_ERROR_UNKNOWN_FS = 12` to maintain code numbers.

vojtechtrefny (approved):
Looks good to me, thanks.

src/plugins/fs/mount.c
vojtechtrefny: > When backporting to 2.x branch, the new error code would need to be specified as `BD_FS_ERROR_UNKNOWN_FS = 12` to maintain code numbers.

Please prepare the 2.x backport PR too, thanks. We already have quite a few bugfixes there so I plan to do a new release soon.

PR #690, opened 2022-01-19
Allow passing NULL for error (fs and btrfs plugins)

vojtechtrefny:
One step closer to fixing #527

PR #692, opened 2022-01-21
Fix multiple error path issues in kbd plugin

vojtechtrefny:
Discovered while working on #690


src/plugins/kbd.c
```diff
@@ -812,6 +812,7 @@ gboolean bd_kbd_bcache_create (const gchar *backing_device, const gchar *cache_d
         if (!present) {
             g_set_error (error, BD_KBD_ERROR, BD_KBD_ERROR_BCACHE_NOEXIST,
                          "Failed to locate uuid symlink '%s'", device_uuid[i]);
+            bd_utils_report_finished (progress_id, (*error)->message);
```
tbzatek, line 815: I suppose this will need to be rebased after #690, right?

tbzatek (approved):

src/plugins/kbd.c
vojtechtrefny: No, kbd plugin is not part of #690. I'll merge both and kbd will be next to be fixed.

PR #696, opened 2022-01-26
s390: Remove double fclose in bd_s390_dasd_online (#2045784)

vojtechtrefny:


tbzatek (approved):
Oh so yummy new compiler and added warnings!

PR #698, opened 2022-01-26
Allow passing NULL for error (kbd and loop plugins)

vojtechtrefny:


tbzatek (approved):
More or less good otherwise.

PR #699, opened 2022-01-28
Skip NVMe tests on Debian and CentOS 9

vojtechtrefny:


tbzatek (approved):
Interesting...

The c9s builds are not critical at this moment, the environment hasn't diverged much.

PR #700, opened 2022-02-03
Allow passing NULL for error (part3) 

vojtechtrefny:
This includes mdraid, mpath, nvdimm, s390 and swap plugins. Only part (where I plan to do some unrelated bigger changes first) and lvm (which I'm working on right now) plugins are left.

### DIFF
--- a/src/plugins/mdraid.c
+++ b/src/plugins/mdraid.c
@@ -215,7 +215,7 @@ void bd_md_close (void) {
  * bd_md_is_tech_avail:
  * @tech: the queried tech
  * @mode: a bit mask of queried modes of operation for @tech
- * @error: (out): place to store error (details about why the @tech-@mode combination is not available)
+ * @error: (out) (allow-none): place to store error (details about why the @tech-@mode combination is not available)
  *
  * Returns: whether the @tech-@mode combination is available -- supported by the
  *          plugin implementation and having all the runtime dependencies available
@@ -277,7 +277,7 @@ static GHashTable* parse_mdadm_vars (const gchar *str, const gchar *item_sep, co
     return table;
 }
 
-static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table, G_GNUC_UNUSED GError **error) {
+static BDMDExamineData* get_examine_data_from_table (GHashTable *table, gboolean free_table) {
     BDMDExamineData *data = g_new0 (BDMDExamineData, 1);
     gchar *value = NULL;
     gchar *first_space = NULL;
@@ -465,7 +465,7 @@ static BDMDDetailData* get_detail_data_from_table (GHashTable *table, gboolean f
 /**
  * get_sysfs_name_from_input: (skip)
  * @input: either RAID name or node name
- * @error: (out): place to store error (if any)
+ * @error: (out) (allow-none): place to store error (if any)
  *
  * Returns: (transfer full): RAID node name
  */
@@ -493,7 +493,7 @@ static gchar* get_sysfs_name_from_input(const gchar *input, GError **error) {
 /**
  * get_mdadm_spec_from_input: (skip)
  * @input: RAID specification from user
- * @error: (out): place to store error (if any)
+ * @error: (out) (allow-none): place to store error (if any)
  *
  * Returns: (transfer full): RAID specification for mdadm
  *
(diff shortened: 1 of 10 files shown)


src/plugins/s390.c
```diff
@@ -598,25 +603,26 @@ gboolean bd_s390_zfcp_online (const gchar *devno, const gchar *wwpn, const gchar
     /* part 01: make sure device is available/not on device ignore list */
     fd = fopen (online, "r");
     if (!fd) {
-        boolrc = bd_utils_exec_and_report_error_no_progress (zfcp_cio_free, NULL, error);
-
+        boolrc = bd_utils_exec_and_report_error_no_progress (zfcp_cio_free, NULL, NULL);
```
vojtechtrefny, line 606: We were leaking errors on multiple places in the s390 plugin. I was thinking about fixing this by prefixing the leaking error instead of overwriting it by `g_set_error`, but in the end I decided to just keep ignoring the error from the calls and not changing the logic here.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

src/plugins/mdraid.c
```diff
@@ -993,9 +991,8 @@ BDMDExamineData* bd_md_examine (const gchar *device, GError **error) {
         return NULL;
     }
 
-    ret = get_examine_data_from_table (table, TRUE, error);
+    ret = get_examine_data_from_table (table, TRUE);
     if (!ret)
-        /* error is already populated */
```
tbzatek, line 998: Hmm, wait - if `NULL` is a valid return value, it should be noted in the doc strings. However I would advise to keep the usual semantics of _return NULL means error is set_. Breaking the API is one option we can afford.

src/plugins/mpath.c
```diff
@@ -345,7 +346,7 @@ static gchar** get_map_deps (const gchar *map_name, guint64 *n_deps, GError **er
 /**
  * bd_mpath_is_mpath_member:
  * @device: device to test
- * @error: (out): place to store error (if any)
+ * @error: (out) (allow-none): place to store error (if any)
  *
  * Returns: %TRUE if the device is a multipath member, %FALSE if not or an error
```
tbzatek, line 351: This API will then hide the difference between an error scenario and valid `FALSE` result.

src/plugins/mpath.c
```diff
@@ -452,13 +455,14 @@ gchar** bd_mpath_get_mpath_members (GError **error) {
...
 
     progress_id = bd_utils_report_started ("Started getting mpath members");
 
     if (geteuid () != 0) {
-        g_set_error (error, BD_MPATH_ERROR, BD_MPATH_ERROR_NOT_ROOT,
+        g_set_error (&l_error, BD_MPATH_ERROR, BD_MPATH_ERROR_NOT_ROOT,
                      "Not running as root, cannot query DM maps");
-        bd_utils_report_finished (progress_id, (*error)->message);
+        bd_utils_report_finished (progress_id, l_error->message);
         return NULL;
```
tbzatek, line 466: Leaking `l_error`. Did you intend to propagate it up?

src/plugins/s390.c
tbzatek: I think this looks correct. The detailed error message might be lost, not a big deal I guess.

src/plugins/swap.c
```diff
@@ -392,7 +408,7 @@ gboolean bd_swap_swapoff (const gchar *device, GError **error) {
 /**
  * bd_swap_swapstatus:
  * @device: swap device to get status of
- * @error: (out): place to store error (if any)
+ * @error: (out) (allow-none): place to store error (if any)
  *
  * Returns: %TRUE if the swap device is active, %FALSE if not active or failed
  * to determine (@error) is set not a non-NULL value in such case)
```
tbzatek, line 414: Same here, I think we can afford to change the API as well.

tbzatek (commented):
Generally looks good, please think about fine-tuning the API for libblockdev-3.0 to fix the ambiguous scenarios.

src/plugins/mdraid.c
vojtechtrefny: `get_examine_data_from_table` never set the error and really couldn't return `NULL` anyway so the this check was always questionable. But I've added error here in case we change `get_examine_data_from_table` in the future.

src/plugins/mpath.c
vojtechtrefny: I asked about this in the [original issue](https://github.com/storaged-project/libblockdev/issues/527) :-)

We have two options:

1. Add an out argument.
2. Tell user to provide error if they are interested in the difference between "not member" and "error occurred".

From the API user view I don't think there is a big difference between these two situations, you need to provide an extra out variable either way. If you think 2 is the better way, we can do that, but probably in a separate PR later (to avoid changing API for UDisks for as long as possible).

tbzatek: That was merely two years ago! :-) My point is, when we have a chance right now to break the API, whether to change this or not...

E.g. the prototype would then look like this:
```C
gboolean bd_mpath_is_mpath_member (const gchar *device, gboolean *result, GError **error)
```

vojtechtrefny: Ok, lets deal with these functions separately so we can merge this without changing API. The functions that need to be changed are tracked in the original issue.

tbzatek (approved):

src/plugins/mpath.c
tbzatek: OK, let's merge this one now.

PR #701, opened 2022-02-03
Provide replacement function for strerror_l()

ffontaine:
`strerror_l()` is not implemented in some C libraries, such as uClibc, so let's provide a simple replacement function that falls back on `strerror()`.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>
[Retrieved from: https://git.buildroot.net/buildroot/tree/package/libblockdev/0001-Provide-replacement-function-for-strerror_l.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -122,6 +122,8 @@ AC_CHECK_HEADERS([dlfcn.h string.h unistd.h sys/fcntl.h sys/ioctl.h linux/random
                  [LIBBLOCKDEV_SOFT_FAILURE([Header file $ac_header not found.])],
                  [])
 
+AC_CHECK_FUNCS([strerror_l])
+
 AC_ARG_WITH([bcache],
     AS_HELP_STRING([--with-bcache], [support bcache @<:@default=yes@:>@]),
     [],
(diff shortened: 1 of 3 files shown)


blivetghoul:

tbzatek:
Interesting, thanks for the patch. However this is not a direct replacement since it handles locales differently - we use it to return non-localized string. Is there anything similar available in uClibc? Alternatively, this might be backported from other libc implementations.

Out of curiosity, is there anything else missing for successful libblockdev compilation?

ffontaine:
To my knowledge, there is nothing similar in uclibc-ng. `strerror_l` declaration was added back in 2012 with https://github.com/wbx-github/uclibc-ng/commit/0769334e2e1c8848757dcf6ca265b6c1dcf98e5d but it was never implemented. It should also be noted that locale support can also be disabled by the user in uclibc-ng. Indeed, this feature isn't really needed on embedded systems. 

Finally, this patch was the only one needed to integrate libblockdev in buildroot in 2020.

tbzatek:
To me it looks like an issue that should be fixed on uclibc side.

Further inspiration, not sure how do they handle uclibc though: https://lists.infradead.org/pipermail/libnl/2016-August/002193.html

PR #707, opened 2022-02-09
install-test-dependencies: Do not install reiserfs-utils on Fedora

vojtechtrefny:
The package is no longer available on Fedora 35 and newer.


vojtechtrefny:
> What??! :-o

Yep, it was orphaned for long time and retired in F34. I plan to unretire it, but haven't found time to do that yet.

PR #713, opened 2022-03-01
tests: Use NILFS2 instead of NTFS to test mount helper execution

vojtechtrefny:
NTFS is now in the kernel so we need a different filesystem with
a mount.<type> helper to test.

### DIFF
--- a/tests/fs_tests/mount_test.py
+++ b/tests/fs_tests/mount_test.py
@@ -222,15 +222,15 @@ class MountTestCase(FSTestCase):
             BlockDev.fs_unmount(self.loop_dev, run_as_uid=uid, run_as_gid=gid)
         self.assertTrue(os.path.ismount(tmp))
 
-    def test_mount_ntfs(self):
-        """ Test basic mounting and unmounting with NTFS filesystem"""
-        # using NTFS because it uses a helper program (mount.ntfs) and libmount
+    def test_mount_nilfs(self):
+        """ Test basic mounting and unmounting with NILFS2 filesystem"""
+        # using NILFS2 because it uses a helper program (mount.nilfs2) and libmount
         # behaves differently because of that
 
-        if not self.ntfs_avail:
-            self.skipTest("skipping NTFS: not available")
+        if not self.nilfs2_avail:
+            self.skipTest("skipping NILFS mount test: not available")
 
-        succ = BlockDev.fs_ntfs_mkfs(self.loop_dev, None)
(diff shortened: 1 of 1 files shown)


tbzatek:
Hmm, looking at the kernel config differences between `5.15.18-200.fc35` and `5.17.0-0.rc4.96.fc36` I see no change:

```
# CONFIG_NTFS_FS is not set
CONFIG_NTFS3_FS=m
```

I wonder what was the real change and where.

vojtechtrefny:
> I wonder what was the real change and where.

We have `ntfs-3g` installed on Fedora 35.

vpodzime (approved):
ACK

vojtechtrefny:
> > We have `ntfs-3g` installed on Fedora 35.
> 
> So that also makes the `ntfs` FS work and be supported? Without it, only `ntfs3` is working?

Yes, naming the kernel driver `ntfs3` was a really bad idea. We have a [UDisks issue](https://github.com/storaged-project/udisks/issues/932) for that, but even `mount` from command line doesn't work if you don't specify `-t ntfs3`:

```
$ sudo mount /dev/sdb1 /mnt/test
mount: /mnt/test: unknown filesystem type 'ntfs'.
       dmesg(1) may have more information after failed mount system call.
```

tbzatek:
There's also this udisks packaging issue: https://bugzilla.redhat.com/show_bug.cgi?id=2058506 and I'm planning to make `ntfs-3g` a hard dep for udisks (F36+).

Technically `ntfs3` is a completely separate filesystem and perhaps deserves its own set of test cases ;-)

That said I'd suggest `libblockdev` tests installing `ntfs-3g` as well...

tbzatek (approved):
I guess this is equally good enough solution.

vojtechtrefny:
> That said I'd suggest `libblockdev` tests installing `ntfs-3g` as well...

Libblockdev mostly doesn't really test `ntfs-3g`, we test `ntfsprogs`. The mount test should be mostly generic, we used NTFS only to test our code works when libmount uses the helper (e.g. `mount.ntfs-3g`) and if we can test that with nilfs2 and avoid NTFS, I think it's better to do that.

tbzatek:
> Libblockdev mostly doesn't really test `ntfs-3g`, we test `ntfsprogs`. The mount test should be mostly generic, we used NTFS only to test our code works when libmount uses the helper (e.g. `mount.ntfs-3g`) and if we can test that with nilfs2 and avoid NTFS, I think it's better to do that.

OK, agreed.

PR #717, opened 2022-03-04
Add missing functions to libblockdev-sections.txt

vojtechtrefny:
It would be nice to have a test for this in the CI to make sure this doesn't happen again. Any volunteers to write a GH action for it? :-)


tbzatek (approved):
Hmmm, that's a good idea for the extended checks! Still missing the volunteer though.

Issue #676, opened 2021-11-24
Replace deprecated GObject-Introspection annotations


vojtechtrefny:
`(allow-none)` is deprecated and should be replaced by `(nullable)` and `(optional)`, see [GObject Introspection documentation](https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations#Deprecated_GObject-Introspection_annotations) for more details.

PR #718, opened 2022-03-04
Use nullable and optional GI annotations instead of allow-none

vojtechtrefny:
allow-none is deprecated since 1.42.

Fixes: #676


tbzatek (approved):
My browser is choking, I refuse to review this!

I vaguely remember some issues about `(optional)` but I can't remember what it was exactly. Just make sure the generated docs are correct, g-i translations will be covered by Python tests.

PR #719, opened 2022-03-08
tests: Load the nvme-fabrics kernel module for nvme tests

tbzatek:
For some reason it doesn't get autoloaded even after setting up
a local loop nvme target and nvme-cli fails to scan the topology
afterwards.


vojtechtrefny (commented):
The tests on rawhide are still failing so I assume the problem is somewhere else?

tbzatek:
> The tests on rawhide are still failing so I assume the problem is somewhere else?

Indeed, the problem was in incorrect `modprobe` usage...

tbzatek:
Pushed more fixes on the `libblockdev` side, the rest of the failures are `nvme-cli` which I'm working on right now.

vojtechtrefny (approved):
If we won't be able to get `nvme-cli` fixed in rawhide, we should just skip the tests. But these changes looks OK so you can merge this now if you want.

tbzatek:
> If we won't be able to get `nvme-cli` fixed in rawhide, we should just skip the tests. But these changes looks OK so you can merge this now if you want.

As it turned out, `nvme-cli` 2.0 series differ in JSON output, more adaptations would be needed for `libblockdev` tests. It's not only about `nvme-cli` returning wrong return codes and segfaulting at places.

tbzatek:
Jenkins, please test this.

tests/nvme_test.py
```diff
@@ -17,6 +17,9 @@ def setUpClass(cls):
             raise unittest.SkipTest("nvme executable (nvme-cli package) not found in $PATH, skipping.")
         if not find_executable("nvmetcli"):
             raise unittest.SkipTest("nvmetcli executable not found in $PATH, skipping.")
+        ret, out, err = run_command("modprobe nvme-fabrics")
```
vojtechtrefny, line 20: It might be useful to create some helper "modprobe" function in utils if we are going to use this at more places.

tests/nvme_test.py
```diff
@@ -277,12 +277,12 @@ def test_sanitize(self):
...
 
 class NVMeFabricsTestCase(NVMeTest):
-    HOSTNQN = 'libblockdev_nvmeof_hostnqn'
-    SUBNQN = 'libblockdev_nvmeof_subnqn'
+    SUBNQN = 'libblockdev_nvme'
 
     def setUp(self):
         self.loop_devs = []
         self.dev_files = []
+        self.hostnqn = get_nvme_hostnqn()
```
vojtechtrefny, line 285: If the hostnqn doesn't change between tests, we should move this to `setUpClass`.

tbzatek: Good point. We might want to extend the tests with `hostnqn` changes in the future. This has been historically source of issues in `libnvme`. I'd leave this as is for the moment.

PR #723, opened 2022-03-27
fs: Remove support for ReiserFS

vojtechtrefny:
The userspace tools have been removed from Fedora and ReiserFS
is deprecated and scheduled to be removed from kernel in 2025.


tbzatek (approved):
www.nooooooooooooooo.com

PR #725, opened 2022-03-28
part: Align end sector when resizing partition to max size

vojtechtrefny:
Latest libfdisk added end alignment for newly created partitions
to make sure the partitions is aligned to 4k sectors on 4k disks.
This means partitions created by us are aligned but partitions
resized to the max size are not because we are calculating the
size and not libfdisk. This change makes sure the end sector is
aligned the same way libfdisk does it for new partitions.


tbzatek (approved):
Interesting...

PR #727, opened 2022-03-29
Add support for setting label and UUID for LUKS devices

vojtechtrefny:


src/plugins/crypto.c
```diff
@@ -2111,6 +2115,201 @@ gboolean bd_crypto_luks_header_restore (const gchar *device, const gchar *backup
...
+
+static gint synced_close (gint fd) {
+    gint ret = 0;
+    ret = fsync (fd);
+    if (close (fd) != 0)
+        ret = 1;
+    return ret;
+}
+
+static gboolean get_subsystem_label (const gchar *device, gchar **subsystem, gchar **label, GError **error) {
```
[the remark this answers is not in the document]
tbzatek: Agree, this is somewhat complex and might deserve separation. Not necessarily right now though.

src/plugins/crypto.c
```diff
@@ -2273,7 +2372,20 @@ BDCryptoLUKSInfo* bd_crypto_luks_info (const gchar *device, GError **error) {
...
 
+    if (info->version == BD_CRYPTO_LUKS_VERSION_LUKS2) {
+        success = get_subsystem_label (crypt_get_device_name (cd) , &(info->subsystem), &(info->label), error);
+        if (!success) {
+            crypt_free (cd);
+            bd_crypto_luks_info_free (info);
+            return NULL;
+        }
+    } else {
+        info->label = g_strdup ("");
```
tbzatek, line 2383: Does it need to be empty string? Is use of NULL a problem somewhere?

tbzatek (approved):
Looks good, some questions posted.

src/plugins/crypto.c
vojtechtrefny: We use empty string for the filesystem labels so I just wanted to be consistent. I guess having it always a valid string is easier for people that just want to print/show the label, no need for an extra check to make sure it's not `NULL`/`None`.

tbzatek: Okay then, fair enough.

PR #730, opened 2022-04-08
lvm: Enhancements for LVM2 RAID support

mvollmer:
This exposes the health status etc and allows for repairs.

This is part of supporting LVM2 RAID in Cockpit:

https://github.com/storaged-project/udisks/pull/969
https://github.com/cockpit-project/cockpit/pull/17226

There is a first demo already: https://www.youtube.com/watch?v=86veT8waLZE
Please check the cockpit PR for updates.

### DIFF
--- a/src/lib/plugin_apis/lvm.api
+++ b/src/lib/plugin_apis/lvm.api
@@ -123,6 +123,7 @@ typedef struct BDLVMPVdata {
     guint64 vg_free_count;
     guint64 vg_pv_count;
     gchar **pv_tags;
+    gboolean missing;
 } BDLVMPVdata;
 
 /**
@@ -151,6 +152,7 @@ BDLVMPVdata* bd_lvm_pvdata_copy (BDLVMPVdata *data) {
     new_data->vg_free_count = data->vg_free_count;
     new_data->vg_pv_count = data->vg_pv_count;
     new_data->pv_tags = g_strdupv (data->pv_tags);
+    new_data->missing = data->missing;
 
     return new_data;
 }
@@ -1113,6 +1115,21 @@ gboolean bd_lvm_lvrename (const gchar *vg_name, const gchar *lv_name, const gcha
  * Tech category: %BD_LVM_TECH_BASIC-%BD_LVM_TECH_MODE_MODIFY
  */
 gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 size, const BDExtraArg **extra, GError **error);
+
+/**
+ * bd_lvm_lvrepair:
+ * @vg_name: name of the VG containing the to-be-repaired LV
+ * @lv_name: name of the to-be-repaired LV
+ * @pv_list: (array zero-terminated=1): list of PVs to be used for the repair
+ * @extra: (nullable) (array zero-terminated=1): extra options for the LV repair
+ *                                                 (just passed to LVM as is)
+ * @error: (out) (optional): place to store error (if any)
+ *
+ * Returns: whether the @vg_name/@lv_name LV was successfully repaired or not
+ *
+ * Tech category: %BD_LVM_TECH_BASIC-%BD_LVM_TECH_MODE_MODIFY
+ */
+gboolean bd_lvm_lvrepair (const gchar *vg_name, const gchar *lv_name, const gchar **pv_list, const BDExtraArg **extra, GError **error);
 
 /**
  * bd_lvm_lvactivate:
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -989,6 +989,7 @@ static BDLVMPVdata* get_pv_data_from_props (GVariant *props, GError **error) {
     g_variant_dict_lookup (&dict, "FreeBytes", "t", &(data->pv_free));
     g_variant_dict_lookup (&dict, "SizeBytes", "t", &(data->pv_size));
     g_variant_dict_lookup (&dict, "PeStart", "t", &(data->pe_start));
+    g_variant_dict_lookup (&dict, "Missing", "b", &(data->missing));
 
     value = g_variant_dict_lookup_value (&dict, "Tags", (GVariantType*) "as");
     if (value) {
@@ -2443,6 +2444,25 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
     g_variant_builder_clear (&builder);
 
     return call_lv_method_sync (vg_name, lv_name, "Resize", params, NULL, extra, TRUE, error);
+}
+
+/**
+ * bd_lvm_lvrepair:
+ * @vg_name: name of the VG containing the to-be-repaired LV
+ * @lv_name: name of the to-be-repaired LV
+ * @pv_list: (array zero-terminated=1): list of PVs to be used for the repair
+ * @extra: (nullable) (array zero-terminated=1): extra options for the LV repair
+ *                                                 (just passed to LVM as is)
+ * @error: (out) (optional): place to store error (if any)
+ *
+ * Returns: whether the @vg_name/@lv_name LV was successfully repaired or not
+ *
+ * Tech category: %BD_LVM_TECH_BASIC-%BD_LVM_TECH_MODE_MODIFY
+ */
+gboolean bd_lvm_lvrepair (const gchar *vg_name UNUSED, const gchar *lv_name UNUSED, const gchar **pv_list UNUSED, const BDExtraArg **extra UNUSED, GError **error) {
+  g_set_error (error, BD_LVM_ERROR, BD_LVM_ERROR_TECH_UNAVAIL,
+               "lvrepair is not supported by this plugin implementation.");
+  return FALSE;
 }
 
 /**
--- a/src/plugins/lvm.c
+++ b/src/plugins/lvm.c
@@ -89,6 +89,7 @@ BDLVMPVdata* bd_lvm_pvdata_copy (BDLVMPVdata *data) {
     new_data->vg_free_count = data->vg_free_count;
     new_data->vg_pv_count = data->vg_pv_count;
     new_data->pv_tags = g_strdupv (data->pv_tags);
+    new_data->missing = data->missing;
 
     return new_data;
 }
@@ -542,6 +543,9 @@ static BDLVMPVdata* get_pv_data_from_table (GHashTable *table, gboolean free_tab
         data->pv_tags = g_strsplit (value, ",", -1);
     else
         data->pv_tags = NULL;
+
+    value = (gchar*) g_hash_table_lookup (table, "LVM2_PV_MISSING");
+    data->missing = (g_strcmp0 (value, "missing") == 0);
 
     if (free_table)
         g_hash_table_destroy (table);
@@ -1175,7 +1179,7 @@ BDLVMPVdata* bd_lvm_pvinfo (const gchar *device, GError **error) {
     const gchar *args[10] = {"pvs", "--unit=b", "--nosuffix", "--nameprefixes",
                        "--unquoted", "--noheadings",
                        "-o", "pv_name,pv_uuid,pv_free,pv_size,pe_start,vg_name,vg_uuid,vg_size," \
-                       "vg_free,vg_extent_size,vg_extent_count,vg_free_count,pv_count,pv_tags",
+                       "vg_free,vg_extent_size,vg_extent_count,vg_free_count,pv_count,pv_tags,pv_missing",
                        device, NULL};
     GHashTable *table = NULL;
     gboolean success = FALSE;
@@ -1194,7 +1198,7 @@ BDLVMPVdata* bd_lvm_pvinfo (const gchar *device, GError **error) {
 
     for (lines_p = lines; *lines_p; lines_p++) {
         table = parse_lvm_vars ((*lines_p), &num_items);
-        if (table && (num_items == 14)) {
+        if (table && (num_items == 15)) {
             g_clear_error (error);
             g_strfreev (lines);
             return get_pv_data_from_table (table, TRUE);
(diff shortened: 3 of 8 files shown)


blivetghoul:

mvollmer:
I guess I need to write some tests... :-)

vojtechtrefny:
Jenkins, ok to test.

src/lib/plugin_apis/lvm.api
```diff
@@ -307,6 +309,9 @@ typedef struct BDLVMLVdata {
     guint64 metadata_percent;
     guint64 copy_percent;
     gchar **lv_tags;
+    gchar *health_status;
+    gchar **devices;
```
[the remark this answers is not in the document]
vojtechtrefny: Btw. `lvs -o devices` also doesn't contain metadata which can be located on different PVs. I think we'll need something way more complicated to this unfortunately.

mvollmer:
> The biggest issue I have with this is the DBus plugin

Ah, I have ignored it completely. Should we agree on how we want to extend the "lvm" plugin exactly, and then I port that over to the "lvm-dbus" plugin?

src/plugins/lvm.c
```diff
@@ -1175,7 +1198,7 @@ BDLVMPVdata* bd_lvm_pvinfo (const gchar *device, GError **error) {
     const gchar *args[10] = {"pvs", "--unit=b", "--nosuffix", "--nameprefixes",
                        "--unquoted", "--noheadings",
                        "-o", "pv_name,pv_uuid,pv_free,pv_size,pe_start,vg_name,vg_uuid,vg_size," \
-                       "vg_free,vg_extent_size,vg_extent_count,vg_free_count,pv_count,pv_tags",
+                       "vg_free,vg_extent_size,vg_extent_count,vg_free_count,pv_count,pv_tags,missing",
```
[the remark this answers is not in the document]
mvollmer: Thanks, done!

src/lib/plugin_apis/lvm.api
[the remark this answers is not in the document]
mvollmer: > I think this needs some explanation what it contains.

Ultimately, it is meant to show which PVs contribute space for a given LV.  My plan right now is to let the user select which PVs to use when creating a LV, and then it is only approriate to show what they are also afterwards.

I realize only now that I haven't linked this PR back to the full story, sorry.  I'll correct that.

> If the idea here is to get "list of PVs this LV is allocated on", we should provide a better results than lvs -o+devices does.

Yes, and currently this is produced in UDisks: https://github.com/storaged-project/udisks/pull/969/files#diff-6d18017b318477b8f7b0121bae3305c5c35f3ebeed6da974725a51a6a42f2961R229

If this should happen in libblockdev instead, I am all for it.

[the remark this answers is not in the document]
mvollmer: > Btw. `lvs -o devices` also doesn't contain metadata which can be located on different PVs so we'd need to use both `devices` and `metadata_devices`. I think we'll need something way more complicated to this unfortunately.

Interesting. Thanks!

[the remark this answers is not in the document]
mvollmer: Also, `devices` and `metadata_devices` are actually `Linear Volume Segment Fields`, and if we ask for them with `-o` we get one line of output for each output for each segment of a LV, not the expected single one.  The same is true for `segtype`, which is already in the list...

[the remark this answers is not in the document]
mvollmer: Or we omit `devices` etc completely here.  It is only used for informational purposes in Cockpit, and not really required to create and repair RAID LVs.

vojtechtrefny:
> While writing the test, I noticed wait_for_sync, which waits for copy_percent to reach 100... What is the relation between copy_percent and sync_percent? Do we maybe only need one?

Good question, we probably need to ask someone from the LVM team. The description for `copy_percent` and `sync_percent` says exactly the same:

```
copy_percent                - For Cache, RAID, mirrors and pvmove, current percentage in-sync.
sync_percent                - For Cache, RAID, mirrors and pvmove, current percentage in-sync.
```

vojtechtrefny:
> > The biggest issue I have with this is the DBus plugin
> 
> Ah, I have ignored it completely. Should we agree on how we want to extend the "lvm" plugin exactly, and then I port that over to the "lvm-dbus" plugin?

Ok, good point. Let's deal with the DBus plugin later.

src/lib/plugin_apis/lvm.api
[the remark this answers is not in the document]
vojtechtrefny: > > It would be nice to have a full "map" of the segments with ranges in libblockdev.
> 
> Yes, I see. But doesn't this contradict "We should definitely move [pv collection] to libblockdev."? Either libblockdev gives a list of segments, or it gives a pre-cooked list of PVs, no? Or do you want it to do both?
> 
> Would it make sense to exclude `devices` from this PR and move this whole topic to a separate one? From Cockpit's point of view `devices` is only "nice to have".

If it's just a nice-to-have, it would be better to implement this separately. I was thinking about having both segments and PVs, probably as a separate struct. We can create a separate RFE for this a discuss it more later.

mvollmer:
> Good question, we probably need to ask someone from the LVM team.

I am collecting questions here: https://github.com/cockpit-project/cockpit/pull/17226#issue-1195618741, I'll add this to the list.

src/lib/plugin_apis/lvm.api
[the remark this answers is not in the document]
mvollmer: Roger.

mvollmer:
I now see that `lv_health_status` is already encoded in `lv_attr[9]`, so we don't need it either...

mvollmer:
I think the test is failing because `/tmp` runs out of memory while the sparse files are being written to. I guess the new test is different in that the files are actually written fully and become unsparse when the mirror halves are synced up...

mvollmer:
> Bad news is that [the lvm2 d-bus api] doesn't provide a function for Repair, but I can live with that (but we still need an "empty" function to return the "not implemented" error in the plugin, it wouldn't be the first

@vojtechtrefny, can you point to an example of such a empty function in lvm-dbus.c ? I can't find any... However, bd_lvm_devices_add in lvm-dbus.c spawns "lvmdevices". Should we maybe just spawn "lvconvert --repair" in lvm-dbus.c as well?

> and we can add it to lvmdbusd later).

Maybe lvmdbusd can just call libblockdev for that. :-)

mvollmer:
`lvm_dbus_tests.py` doesn't seem to run during CI, right?  (Because it doesn't match "*_test.py"? :-)

mvollmer:
I think I have reached a stable point here. I have squashed the FIXUPs, please have another look!

vojtechtrefny:
> I think the test is failing because `/tmp` runs out of memory while the sparse files are being written to. I guess the new test is different in that the files are actually written fully and become unsparse when the mirror halves are synced up...

Yes, I had the same issue few months ago https://github.com/storaged-project/libblockdev/commit/4562941ced82efa55d6515f8470c8f9bc46fb414

> can you point to an example of such a empty function in lvm-dbus.c

I actually fixed the missing functionality in lvm-dbus few months ago. Example from the older code: https://github.com/storaged-project/libblockdev/blob/2.26-1/src/plugins/lvm-dbus.c#L3529 it simply calls `bd_lvm_is_tech_avail` to return error for unsupported technologies.

> However, bd_lvm_devices_add in lvm-dbus.c spawns "lvmdevices". Should we maybe just spawn "lvconvert --repair" in lvm-dbus.c as well?

I don't think it's necessary. UDisks forces the CLI plugin so it will never use the DBus one, we can deal with this in the future.

> Maybe lvmdbusd can just call libblockdev for that. :-)

+1 Everyone loves circular dependencies :-)

> lvm_dbus_tests.py doesn't seem to run during CI, right? (Because it doesn't match "*_test.py"? :-)

We use `pattern='*_test*.py'`. There was some reason for naming the LVM DBus test file differently, I don't remember why we did that (maybe to be able to run all tests except these in some cases?), we should probably rename it to make it less confusing.

> I think I have reached a stable point here. I have squashed the FIXUPs, please have another look!

Great, I'll try to do the review as soon as possible.

mvollmer:
> > However, bd_lvm_devices_add in lvm-dbus.c spawns "lvmdevices". Should we maybe just spawn "lvconvert --repair" in lvm-dbus.c as well?
> 
> I don't think it's necessary. UDisks forces the CLI plugin so it will never use the DBus one, we can deal with this in the future.

I copy/pasted the code from lvm.c and it seems to work. *shrug*

src/lib/plugin_apis/lvm.api
```diff
@@ -1114,6 +1116,21 @@ gboolean bd_lvm_lvrename (const gchar *vg_name, const gchar *lv_name, const gcha
  */
 gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 size, const BDExtraArg **extra, GError **error);
 
+/**
+ * bd_lvm_lvrepair:
+ * @vg_name: name of the VG containing the to-be-repaired LV
+ * @lv_name: name of the to-be-repaired LV
+ * @pv_list: (array zero-terminated=1): list of PVs to be used for the repair
+ * @extra: (nullable) (array zero-terminated=1): extra options for the LV resize
```
vojtechtrefny, line 1124: *extra options for the LV resize* -> *repair*

src/plugins/lvm-dbus.c
```diff
@@ -2445,6 +2446,41 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
...
+ * @pv_list: (array zero-terminated=1): list of PVs to be used for the repair
+ * @extra: (nullable) (array zero-terminated=1): extra options for the LV resize
+ *                                                 (just passed to LVM as is)
+ * @error: (out) (optional): place to store error (if any)
+ *
+ * Returns: whether the @vg_name/@lv_name LV was successfully repaired or not
+ *
+ * Tech category: %BD_LVM_TECH_BASIC-%BD_LVM_TECH_MODE_MODIFY
+ */
+gboolean bd_lvm_lvrepair (const gchar *vg_name, const gchar *lv_name, const gchar **pv_list, const BDExtraArg **extra, GError **error) {
```
vojtechtrefny, line 2462: I still don't like the idea of using LVM cli in the DBus plugin. I know that it's not possible to have lvmdbusd without lvm so this isn't really adding a new dependency, but it still makes `bd_lvm_is_tech_avail` a little bit meaning less for the DBus plugin in this case. I would prefer not using the cli here and just making the function return an error, because LVM DBus doesn't (yet) support repair.

@tbzatek @vpodzime thoughts?

tests/lvm_dbus_tests.py
```diff
@@ -1172,6 +1189,88 @@ def _clean_up(self):
...
+        info = BlockDev.lvm_lvinfo("testVG", "testLV")
+        self.assertTrue(info)
+        self.assertEqual(info.attr[8], "-")
+
+        # Disconnect the second PV, this should cause it to be flagged
+        # as missing, and testLV to be reported as "partial".
+        delete_lio_device(self.loop_dev2)
+
+        # HACK - Kick lvmdbusd so that it notices the missing PV. Needing to do
+        #        this might count as a bug in lvmdbusd.
```
vojtechtrefny, line 1237: lvmdbusd unfortunately doesn't react to changes from udev by default, there is a `--dbus` option but that's intentionally not used in the service file.

vojtechtrefny (approved):
Looks good to me in general. Let's wait for other opinions about the LVM DBus plugin repair support.

src/plugins/lvm-dbus.c
vpodzime: I completely agree. That's why we have the flags and checks.

vpodzime (commented):
Thanks for the bold move!

src/plugins/lvm-dbus.c
mvollmer: Roger!

src/lib/plugin_apis/lvm.api
mvollmer: Thanks!

tests/lvm_dbus_tests.py
mvollmer: Understood, thanks!

src/plugins/lvm-dbus.c
mvollmer: It's not clear to me how this should be wired up, unfortunately... so I have just put a g_set_error in there. Could you explain how this should look properly?

vojtechtrefny: Just set error to something like

```
g_set_error (error, BD_LVM_ERROR, BD_LVM_ERROR_TECH_UNAVAIL, "lvrepair is not supported by this plugin implementation.");
```
and return `FALSE` from the function.

vojtechtrefny (approved):
Looks good to me, thank you. If this is the final version, I think we can merge it now.

src/plugins/lvm-dbus.c
```diff
@@ -2445,6 +2446,29 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
...
+ * @extra: (nullable) (array zero-terminated=1): extra options for the LV repair
+ *                                                 (just passed to LVM as is)
+ * @error: (out) (optional): place to store error (if any)
+ *
+ * Returns: whether the @vg_name/@lv_name LV was successfully repaired or not
+ *
+ * Tech category: %BD_LVM_TECH_BASIC-%BD_LVM_TECH_MODE_MODIFY
+ */
+gboolean bd_lvm_lvrepair (const gchar *vg_name, const gchar *lv_name, const gchar **pv_list, const BDExtraArg **extra, GError **error) {
+  (void) vg_name;
```
tbzatek, line 2463: yay, a victim of `-Werror`? :face_with_head_bandage:

tbzatek (approved):

src/plugins/lvm-dbus.c
vojtechtrefny: Oops, I completely missed this. Please use the `UNUSED` macro, we are already using it in lvm-dbus.

mvollmer: Oh, much better, thanks!  Done.

PR #732, opened 2022-04-21
tests: Fix checking hostnqn length in get_nvme_hostnqn

vojtechtrefny:


tbzatek (approved):
Eh, good catch, thanks

PR #742, opened 2022-05-26
tests: Fix expected extended partition flags with new parted

vojtechtrefny:
Latest libparted now creates all extended partitions as
PARTITION_EXT_LBA with the lba flag enabled by default.

------

This probably isn't the best solution, but I don't really want to deal with parted changes. Lets hope we'll release 3.0 with libfdisk part plugin soon.


tbzatek (approved):
wicked

PR #746, opened 2022-06-03
nvme: Host NQN and Host ID

tbzatek:
Might be useful for blivet!

### DIFF
--- a/src/plugins/nvme/nvme-fabrics.c
+++ b/src/plugins/nvme/nvme-fabrics.c
@@ -27,6 +27,7 @@
 #include <fcntl.h>
 #include <malloc.h>
 #include <linux/fs.h>
+#include <glib/gstdio.h>
 
 #include <libnvme.h>
 #include <uuid/uuid.h>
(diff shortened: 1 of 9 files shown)


tbzatek:
FYI, the tests are failing simply due to a difference of `libnvme` reading files from `/etc/nvme` while the in-tree `libblockdev` sources are configured with a default `/usr/local/etc` `sysconfdir`.

src/plugins/nvme/nvme-fabrics.c
```diff
@@ -766,3 +767,135 @@ gchar ** bd_nvme_find_ctrls_for_ns (const gchar *ns_sysfs_path, const gchar *sub
...
     return (gchar **) g_ptr_array_free (ptr_array, FALSE);
 }
+
+
+/**
+ * bd_nvme_get_host_nqn:
+ * @error: (out) (nullable): Place to store error (if any).
+ *
+ * Reads the Host NQN (NVM Qualified Name) value from the `/etc/nvme/hostnqn` file
+ * or generates new one from system DMI data.
```
vojtechtrefny, line 781: This feels kinda unintuitive, especially because `bd_nvme_get_host_id` doesn't generate new ID when the file doesn't exist, it might be better to have two functions -- one to read nqn from the file and one to generate id.

vojtechtrefny (commented):
Looks good to me in general, but we should figure out the tests somehow.

vojtechtrefny:
> FYI, the tests are failing simply due to a difference of `libnvme` reading files from `/etc/nvme` while the in-tree `libblockdev` sources are configured with a default `/usr/local/etc` `sysconfdir`.

So running the tests with `configure --prefix=/usr` would fix this? I guess we could do that in the CI, but it would still feel weird to have a test that fails when you run `./configure && make && make ci`.

src/plugins/nvme/nvme-fabrics.c
tbzatek: Separated, after mutual agreement.

tbzatek:
I guess the best way for the moment would be to skip the tests if the prefix is different. There are external factors that we can't change.

tests/nvme_test.py
```diff
@@ -530,6 +538,11 @@ def test_host_nqn(self):
         FAKE_HOSTID1 = 'eeeeeeee-eeee-eeee-eeee-eeeeeeeeeeee'
         FAKE_HOSTID2 = 'beefbeef-beef-beef-beef-beefdeadbeef'
 
+        # libnvme might have been configured with a different prefix than libblockdev
+        sysconf_dir = self._get_sysconf_dir()
+        if sysconf_dir != '/etc':
+            raise unittest.SkipTest("libblockdev was not configured with standard prefix (/usr)")
```
vojtechtrefny, line 544: Please use `self.skipTest` here (we have `self` here and it produces better output).

tests/nvme_test.py
```diff
@@ -324,6 +324,14 @@ def _nvme_disconnect(self, subnqn, ignore_errors=False):
         if not ignore_errors and (ret != 0 or 'disconnected 0 ' in out):
             raise RuntimeError("Error disconnecting the '%s' subsystem NQN: '%s %s'" % (subnqn, out, err))
 
+    def _get_sysconf_dir(self):
+        try:
+            makefile = read_file(os.path.join(os.environ['LIBBLOCKDEV_PROJ_DIR'], 'Makefile'))
+            r = re.search(r'sysconfdir = (.*)', makefile)
```
vojtechtrefny, line 330: I don't have a better idea how to check for *sysconfigdir* here, so I'm just going to pretend I didn't see this.

vojtechtrefny (approved):
Looks good to me, thanks.

tests/nvme_test.py
tbzatek: Fixed.

tests/nvme_test.py
tbzatek: The other way would be to generate some _config.py_ file that would include such string. Not much easier.

PR #752, opened 2022-07-25
Use crypt_keyfile_read to read key files and cryptsetup required version bump

vojtechtrefny:
Fixes: #471

New version of #540, there was a lot of merge conflicts so creating a new PR. In the original discussion we agreed to wait because cryptsetup 2.3.0 was not yet widely available, but this has now changed -- 2.3.0 is available in Debian stable, RHEL 8 and 9 and all supported versions of Fedora so bumping the requirement shouldn't be a problem now.


vojtechtrefny:
> so bumping the requirement shouldn't be a problem now.

Famous last words. LGTM is using Ubuntu 20.04...

src/plugins/crypto.c
```diff
@@ -1507,29 +1575,67 @@ gboolean bd_crypto_luks_remove_key_blob (const gchar *device, const guint8 *pass
...
 
+    msg = g_strdup_printf ("Started removing key from the LUKS device '%s'", device);
+    progress_id = bd_utils_report_started (msg);
+    g_free (msg);
+
+    ret = crypt_init (&cd, device);
+    if (ret != 0) {
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_DEVICE,
+                     "Failed to initialize device: %s", strerror_l (-ret, c_locale));
+        bd_utils_report_finished (progress_id, (*error)->message);
```
tbzatek, line 1594: Probably better to use the `l_error` local intermediary result variable so that `error` can be passed in as `NULL`.

Alternatively, since this patter is repeating quite often, might be useful to think about some local `bd_` wrapper for the related job reporting. Just an idea, could be a macro even.

tbzatek (approved):
Nice cleanup. Beware of dereferencing `NULL` `GError`.

src/plugins/crypto.c
vojtechtrefny: Oops, forgot to fix it here.

vojtechtrefny:
> > so bumping the requirement shouldn't be a problem now.
> 
> Famous last words. LGTM is using Ubuntu 20.04...

Ok, let's wait few more months and if LGTM won't upgrade to 22.04, we'll disable it.

vojtechtrefny:
Rebased to latest master. The issue with LGTM resolved itself -- it is being replaced by GitHubs CodeQL (see #805) which runs in GH actions where we can use Ubuntu 22.04.

PR #756, opened 2022-07-29
nvme: Differentiate fabrics connection error codes

tbzatek:
The kernel nvme host driver maps errors to standard errno values,
libnvme maps them back to own internal codes and libblockdev does
the mapping once again for GError. The kernel host driver often
spits more information into the kernel log, but that's very
difficult for userspace to catch for the particular call.

Some codes are available only with libnvme >= 1.1


vojtechtrefny (approved):
Looks good to me.

PR #759, opened 2022-08-03
fs: Fix optional error handling in fs_mount helper function

vojtechtrefny:
This function claims to allow passing NULL instead of error so
we need to actually allow it.

These types of bugs are unfortunately hard to catch, because the Python tests always provide errors, I found this one when working on https://github.com/storaged-project/udisks/pull/763


tbzatek (approved):
Great!

PR #763, opened 2022-08-09
Remove KBD plugin

vojtechtrefny:
The plugin is not being used (UDisks is removing support for
bcache and zRAM). Bcache support is broken everywhere and zRAM is
now managed by systemd services.


vojtechtrefny:
So @vpodzime  isn't a fan of removing working code so I am now considering removing only the bcache code (which doesn't really work) and keeping the zram support even though it isn't that much useful in my opinion. I'll probably "rename" the plugin to just `zram`. Thoughts?

tbzatek:
Hmm. The UDisks `zram` plugin definitely needs rewrite as it's horribly broken (and messes up with your system) and until we find a maintainer that would bring it back with obligation of long-term support of wide range of distributions and their custom zram generators, it's better to just let it go.

That would leave `libblockdev` with no user of the `kbd` plugin. The `libblockdev` API looks rather sane and simple but will necessarily bring some maintainance burden (although rather small).

If the final decision is not to drop the `kbd` plugin, I'd keep the UDisks plugins around, marking them as deprecated, disabled by default with a big fat warning that they're broken and use is on consumer's risk.

tbzatek:
So? Has the jury came to a conclusion?

vojtechtrefny:
> So? Has the jury came to a conclusion?

I discussed this with @vpodzime and we agreed that having a libblockdev plugin just for zram is probably not worth but it would still be nice to keep the code somewhere, so I'll create a new separate *libzram* library separate from libblockdev. I'll rebase this to the latest master and we can merge this after removing the modules from udisks in https://github.com/storaged-project/udisks/pull/997

tbzatek (approved):

PR #764, opened 2022-08-09
lvm: Fix bd_lvm_get_supported_pe_sizes in Python on 32bit

vojtechtrefny:
Pygobject fails to get size of the returned zero terminated array
on 32bit systems. With older versions we get only the first value
from the list and with newer versions we get an abort. The number
of supported PE sizes is constant so we can use the fixed-size
annotation to workaround this issue without changing API.


tbzatek (approved):
This is crazy.

PR #765, opened 2022-08-09
lvm: Fix bd_lvm_get_supported_pe_sizes in Python on 32bit

vojtechtrefny:
Pygobject fails to get size of the returned zero terminated array
on 32bit systems. With older versions we get only the first value
from the list and with newer versions we get an abort. The number
of supported PE sizes is constant so we can use the fixed-size
annotation to workaround this issue without changing API.

#764 for master, we might want to change the API here to make it something that pygobject understands better (e.g. using array of pointers or adding a new out argument with the number of elements).


tbzatek (approved):
> we might want to change the API here to make it something that pygobject understands better (e.g. using array of pointers or adding a new out argument with the number of elements).

Subject to discussion, having something else than the magic number _25_ would be nice.
Good enough fix for now.

Issue #768, opened 2022-08-10
Add an API for partition IDs and GUIDs for filesystems and other use cases


vojtechtrefny:
UDisks currently has a list of partition IDs and GUIDs for known filesystems and formats, we should add this to libblockdev, either to the filesystem plugin or to the part plugin (or to both). With the planned parted-inspired `BDPartFlag` removal, the only API to set these will be `bd_part_set_part_type` for GPT and `bd_part_set_part_id` for MSDOS. We should add a simpler API so our users don't need their own lists of partition types/ids to set.

Related: #760

PR #771, opened 2022-08-11
fs: Add partition IDs/types to BDFSFeatures

vojtechtrefny:
UDisks needs an API to get partiton IDs and types to set when
creating a new format.

Fixes: #768


src/lib/plugin_apis/fs.api
```diff
@@ -2540,13 +2540,17 @@ typedef enum {
  * @fsck: support for fsck operations (check and repair)
  * @configure: support for changing properties of an existing filesystem
  * @features: other supported features
+ * @partition_id: partition ID used for this filesystem on MSDOS partitions
+ * @partition_type: partition type/GUID used for this filesystem GPT partitions
```
tbzatek, line 2544: Hmm, this is misleading on a first look. Any reason not to use something like `partition_id_mbr` and `partition_uuid_gpt`?

tbzatek (approved):
Cool, thanks for adding this :-)

src/lib/plugin_apis/fs.api
vojtechtrefny: `id` and `type` are "official" names for the entries in MSDOS and GPT partition tables, our part plugin also calls those like that. It might be confusing, but it's an advanced API, I'd expect users to know what they are doing.

tbzatek: Alright, fair enough.

PR #778, opened 2022-08-16
Fix calling bd_fs_mkfs without options from Python 

vojtechtrefny:


tbzatek (approved):
Hmm, interesting. :cat2:

PR #779, opened 2022-08-16
fs: Fix nilfs2 mkfs util in BDFSInfo

vojtechtrefny:
Correct one is mkfs.nilfs2, not mkfs.nilfs.


tbzatek (approved):
:wolf:

vojtechtrefny:
We need to look at how we skip tests if we don't have the filesystem tools -- we had a wrong mkfs tool for nilfs2 and because of that we skipped the mkfs test and because of that we didn't find out that `bd_fs_nilfs2_mkfs_options` was missing the dry run option.

tbzatek (approved):
:ok_hand:

PR #782, opened 2022-08-23
spec: Require the same version utils from plugins

vojtechtrefny:
We don't want to deal with utils backward compatibility, especially
now with 3.0 around the corner, and rpminspect is also complaining
about this.

Resolves: rhbz#2116544


tbzatek (approved):
:satisfied:

PR #783, opened 2022-08-24
fs/ext: Use libext2fs instead of spawning dumpe2fs

mvollmer:
Spawning and parsing is more expensive, and dumpe2fs has the annoying
bug that it triggers udev events when the MMP feature is enabled on a
filesystem.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -20,5 +20,6 @@ extraction:
       - "libbytesize-dev"
       - "libnss3-dev"
       - "libndctl-dev"
+      - "libext2fs-dev"
     configure:
       command: "./autogen.sh && ./configure --without-nvme"
--- a/configure.ac
+++ b/configure.ac
@@ -233,7 +233,8 @@ AS_IF([test "x$with_fs" != "xno"],
        LIBBLOCKDEV_PKG_CHECK_MODULES([MOUNT], [mount >= 2.23.0])
        # new versions of libmount has some new functions we can use
        AS_IF([$PKG_CONFIG --atleast-version=2.30.0 mount],
-             [AC_DEFINE([LIBMOUNT_NEW_ERR_API])], [])]
+             [AC_DEFINE([LIBMOUNT_NEW_ERR_API])], [])
+       LIBBLOCKDEV_PKG_CHECK_MODULES([EXT2FS], [ext2fs e2p])]
       [])
 
 AS_IF([test "x$with_tools" != "xno"],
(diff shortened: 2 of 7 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Please also add the `e2fsprogs-devel` (`libext2fs-dev` for Ubuntu/Debian) to [install-test-dependencies.yml](https://github.com/storaged-project/libblockdev/blob/master/misc/install-test-dependencies.yml) and [.lgtm.yml](https://github.com/storaged-project/libblockdev/blob/master/.lgtm.yml) to make the CI happy.

tbzatek (approved):
I like it! I'm always for a clean library access instead of parsing commandline output.

Please add spaces before opening parentheses, otherwise looks good.

vojtechtrefny (approved):
I'm kinda conflicted about this. On one hand, removing the tool call and avoiding output parsing is great, on the other hand, people already hate us for adding more build time dependencies (even small things like *libkmod* can be controversial). And problem with filesystem plugin is that the filesystem support is not "optional" -- it's perfectly normal use case to have a system without ext and now we are forcing users to install e2fsprogs(-devel) without option to disable it (which is possible for other plugins: if you don't want LUKS support you can build libblockdev without the crypto plugin and without cryptsetup, but if you want XFS support and don't need ext, you won't have the option now).

But I guess the positives of getting rid of `dumpe2fs` will win me over on this one.

pothos:
Dependency changes are not really something special to complain about, I think the question is rather how easy the dependency can be installed and how (in)sane the build system is ;)

mvollmer:
> But I guess the positives of getting rid of `dumpe2fs` will win me over on this one.

A simple alternative would be to make a new `dumpe2fs.libblockdev` executable and spawn that instead of `dumpe2fs`. That would fix the bug of dumpe2fs triggering udev just fine, but would increase the packaging complexity, I guess.

mvollmer:
https://github.com/tytso/e2fsprogs/pull/122 just got merged, so the bug in `dumpe2fs` will go away eventually. Nice!

vpodzime (approved):
I like this. Code reduction, resource reduction (less overhead). :+1: :clap: I don't think requiring things for ext is a problem, ext is pretty much everywhere. For more exotic FSs it might a problem, but not here, AFAICT. But if we find this problematic, we can actually try to load the library dynamically.

Issue #769, opened 2022-08-10
lvm-dbus.c: Fix GError overwrites


tbzatek:
_(opening as a TODO, separated from #740)_
https://github.com/storaged-project/libblockdev/blob/74a6107a6feab960a0254c0f05c1dcdfc4b4ff8a/src/plugins/lvm-dbus.c#L2823-L2824

Overwriting `l_error` without checking, possibly multiple places in the code. Figure out at what point to fail the function or ignore the error.
Re-evaluate once #740 is fixed.

Similar case:
https://github.com/storaged-project/libblockdev/blob/74a6107a6feab960a0254c0f05c1dcdfc4b4ff8a/src/plugins/lvm-dbus.c#L2638-L2639

PR #784, opened 2022-08-25
lvm-dbus: Fix various GError overwrites

vojtechtrefny:
In cases where we don't care about the error we can simply pass
NULL now.

Fixes: #769

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -453,6 +453,14 @@ static gchar** get_existing_objects (const gchar *obj_prefix, GError **error) {
     return ret;
 }
 
+
+/**
+ * get_object_path:
+ * @obj_id: get object path for an LVM object (vgname/lvname)
+ * @error: (out) (optional): place to store error (if any)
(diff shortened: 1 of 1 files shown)


src/plugins/lvm-dbus.c
```diff
@@ -453,6 +453,14 @@ static gchar** get_existing_objects (const gchar *obj_prefix, GError **error) {
     return ret;
 }
 
+
+/**
+ * get_object_path
```
tbzatek, line 458: I think you need to add a trailing colon `:` character... But then, these are only static private functions.

tbzatek (approved):
Haven't gone through the whole file, looks good on a first sight.

src/plugins/lvm-dbus.c
vojtechtrefny: Fixed, I added the docstrings mostly because this isn't the first I was looking at those functions wondering whether it is safe to pass `NULL` for `error` or not. So now I don't need to wonder next time :)

Issue #767, opened 2022-08-10
Add a option to disable partition table creating for bd_fs_mkfs


vojtechtrefny:
Currently two filesystems (UDF and VFAT) create a partition table when formatting a raw disk. We need a special feature flag for this so udisks can detect that the partition table will (possibly) be created and also a mkfs option to disable this behaviour.

PR #786, opened 2022-08-25
fs: Add a mkfs option to disable creating partition table

vojtechtrefny:
Currectly UDF and VFAT create (sometimes) a partition table on
the device when running mkfs and only VFAT allows controlling this
behaviour.

Fixes: #767


tbzatek (approved):
Looks good and works fine in UDisks.

From symbol naming perspective, an alternate option could be `@no_prot_pt`, `BD_FS_MKFS_NO_PROT_PT` etc. or just `@no_prot_table`, `BD_FS_MKFS_NO_PROT_TABLE`. Neither naming is elegant.

Issue #766, opened 2022-08-10
Add force option to BDFSMkfsOptions


vojtechtrefny:
Running mkfs with `--force` is a common use case because some mkfs tools do not allow creating the file system on a raw disk or other non-partition devices by default. We should add the option to `BDFSMkfsOptions` to allow easily running `bd_fs_mkfs` with the force argument.

We should also reconsider whether we want to use the force option by default for the various  `bd_fs_<FS>_mkfs` functions. Currently we use the option by default for Ext, NTFS, NILFS2 and Btrfs.

vojtechtrefny:
So currently we call `mkfs` with `--force` (or similar) for these filesystems

- **btrfs**, where force means _Forcibly  overwrite  the  block devices when an existing filesystem is detected._
- **ext**, where it means _Force mke2fs to create a file system, even if the specified device is not a partition on a block special device, or if other parameters do not make sense._
- **nilfs2**, where it means _Force  overwrite  when  an  existing filesystem is detected on the device._
- **ntfs**, where it means _Force mkntfs to run, even if the specified device is not a block special device, or appears to be mounted._
- **vfat**, where it means _Ignore and disable safety checks._

So we call it with `--force` either to ignore existing formats (btrfs, nilfs2, ntfs, vfat) or to allow creating the format on non-partition device (ntfs, ext*).

*) ext man page claims you need to call it with `-F` to allow formatting non-partition devices, but even older versions allow creating ext on a disk

So the only "problematic" filesystem here is NTFS where `--force` can mean two different options.

I would recommend removing the `--force` option from all filesystems except NTFS and letting the caller to decide whether they want to call `bd_fs_mkfs` with `--force` or not using the new option. Thoughts @tbzatek

vojtechtrefny:
So not only `mke2fs` man page isn't telling the truth (it will create ext on whole disk without `-F`), it also allows overwriting preexisting filesystem when stdin is closed.

tbzatek:
For general purpose use I would recommend not to use force by default. It is a new API and an option to make use of the force is sufficient. Let the `mkfs` tools decide upon themselves with their defaults.

[...]

PR #787, opened 2022-08-26
Add a "force" mkfs option for bd_fs_mkfs

vojtechtrefny:
Fixes: #766 

Two "issues" with this one:

- *NTFS*: I've decided to keep running `mkfs.ntfs` with the `-F` option by default and pretend the option doesn't exist for NTFS. `mkfs.ntfs` will overwrite preexisting filesystems without asking so the `-F` option mostly allows us to format non-partition devices.
- *ext*: The `-F` option is basically no-op for most cases, because running `mke2fs` with closed stdin answers *yes* to the "overwrite" question, but I decided to keep the option here, someone might want to use it for the other use cases.


tbzatek (approved):
May the Force be with you. :merman:

PR #788, opened 2022-08-26
fs: Remove filesystem specific wipe function

vojtechtrefny:
We don't really need these and most of them are somewhat broken
because exfat, vfat, ntfs and udf have multiple signatures and
we currently remove only the "main" signature and not the msdos
partition table signature that is also created by mkfs.

---

Second step will be looking at `bd_fs_clean` and `bd_fs_wipe` and clean their API and possibly add either an option to one of these functions or an entirely new function that would allow wiping a specific signature from the device.


tbzatek (approved):
Now this is what I call cleanup!

PR #789, opened 2022-08-26
Various documentation related fixes

vojtechtrefny:
And my annual reminder that _someone_ should write a test for this so we don't forget to add new functions to `docs/libblockdev-sections.txt` all the time.


tbzatek (approved):
Good catches at places!

PR #796, opened 2022-09-16
nvme: Extend BDNVMELBAFormat with metadata_size

tbzatek:
The metadata size field was originally intentionally omitted for the libblockdev API, hiding low-level implementation details that seemed unimportant. Turned out that some drives expose several metadata sizes for the same lba data size, leading to seemingly duplicate entries in the list, causing confusion in upper layers. As this is used for namespace formatting that is a destructive operation, it's wiser to expose this detail to higher levels after all.


vojtechtrefny (approved):
Looks good to me.

PR #802, opened 2022-10-07
nvme: Rework the disconnect device topology crawl

tbzatek:
This is more sane regarding to memory ownership and the scanned topology tree.


vojtechtrefny (approved):
Looks good to me.

PR #805, opened 2022-10-11
ci: Add GH actions CodeQL static analysis workflow

vojtechtrefny:
Replacement for LGTM.com which is scheduled to shut down in December.


tbzatek (approved):
I don't understand a word here but it looks nice :-]

PR #808, opened 2022-10-21
spec: Obsolte the kbd and kbd-devel subpackages

vojtechtrefny:
Without obsoleting the subpackages upgrading to the latest build won't work without "--best --allowerasing".


tbzatek (approved):
What about `-devel` subpackages?

vojtechtrefny:
> What about `-devel` subpackages?

Oops, the second `Obsoletes` was supposed to be `libblockdev-kbd-devel`, not `libblockdev-kbd-plugin` :facepalm:

PR #818, opened 2022-11-28
configure: Add missing check for libuuid to the NVMe plugin

vojtechtrefny:


tbzatek (approved):
Ah, recent `libnvme` removed the `libuuid` dependency...

Issue #685, opened 2021-12-29
Add support for advanced options for bd_loop_setup


vojtechtrefny:
4k sectors is an obvious candidate (see #684) but we could also add option to enable/disable direct IO and other options available when using losetup from the command line.

PR #819, opened 2022-11-29
loop plugin API changes

vojtechtrefny:
New argument for `bd_loop_setup` and a new function to get information about loop devices. UDisks needs to be adjusted to the new API before merging this.

Fixes: #685

### DIFF
--- a/docs/3.0-api-changes.xml
+++ b/docs/3.0-api-changes.xml
@@ -78,4 +78,13 @@
   </para>
 
   </simplesect>
+
+  <simplesect><title>Loop plugin</title>
+  <para>
+    <literal>bd_loop_setup</literal> and <literal>bd_loop_setup_from_fd</literal> has a new parameter <literal>sector_size</literal>.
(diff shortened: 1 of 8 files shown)


tbzatek (approved):
Looks good

vojtechtrefny:
Moved removing of bd_loop_get_backing_file to a separate PR #822 so this could be merged without breaking blivet.

tbzatek (approved):
Ok then.

vojtechtrefny:
Last change, I promise :) Because we need the removed function only for Blivet and Anaconda, I decided to put a little "helper" one in the overrides which uses the new `bd_loop_info` function to get the backing file. This way removing it later will be just a small change in the python overrides.

tbzatek:
> Last change, I promise :) Because we need the removed function only for Blivet and Anaconda, I decided to put a little "helper" one in the overrides which uses the new `bd_loop_info` function to get the backing file. This way removing it later will be just a small change in the python overrides.

Oh, interesting approach. I like it. Therefore: ship it!

Issue #825, opened 2022-12-07
Add support for setting PARTUUID


vojtechtrefny:
See https://github.com/storaged-project/udisks/issues/1052

PR #826, opened 2022-12-08
part: Add function to set partition UUID

vojtechtrefny:
Fixes: #825

### DIFF
--- a/src/plugins/part.c
+++ b/src/plugins/part.c
@@ -59,6 +59,7 @@ BDPartSpec* bd_part_spec_copy (BDPartSpec *data) {
     ret->path = g_strdup (data->path);
     ret->name = g_strdup (data->name);
     ret->id = g_strdup (data->id);
+    ret->uuid = g_strdup (data->uuid);
     ret->type_guid = g_strdup (data->type_guid);
     ret->type = data->type;
     ret->start = data->start;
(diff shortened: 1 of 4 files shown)


src/plugins/part.c
```diff
@@ -2006,6 +2013,119 @@ gboolean bd_part_set_part_id (const gchar *disk, const gchar *part, const gchar
     return TRUE;
 }
 
+/**
+ * bd_part_set_part_uuid:
+ * @disk: device the partition belongs to
+ * @part: partition the UUID should be set for
+ * @uuid: partition UUID to set
```
tbzatek, line 2020: So does it mean that when `uuid == NULL`, it means the UUID value will be cleared from the partition table? Another option is to generate a new random value (we might do that in UDisks).

tbzatek (approved):
Nice, thanks!

src/plugins/part.c
vojtechtrefny: I don't think it's possible to remove the UUID. The function now doesn't accept `NULL`, none of the `set_X` functions in the part plugin take `NULL`, if you want to remove the value (for things that can be removed, like name), you need to use an empty string for that. We can change it to generate a new random one, that would be similar to how the `bd_fs_*_set_uuid` functions work for filesystem.

tbzatek: Hmm, each of the `bd_fs_*_set_uuid` functions take slightly different "magic" values according to the `mkfs` tool used. And the `part` plugin setters only seems to just pass the `NULL` value further down the `libfdisk` code to see what happens. Okay, let's keep it as proposed, at least it's consistent with the others.

PR #828, opened 2022-12-18
crypto: Add support for opening FileVault encrypted devices

vojtechtrefny:
cryptsetup 2.6.0 added support for macOS FileVault2 devices so we can now support these too.


src/plugins/crypto.c
```diff
@@ -425,6 +425,19 @@ gboolean bd_crypto_is_tech_avail (BDCryptoTech tech, guint64 mode, GError **erro
                 return FALSE;
             } else
                 return TRUE;
+        case BD_CRYPTO_TECH_FVAULT2:
+#ifndef LIBCRYPTSETUP_26
+            g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_TECH_UNAVAIL,
+                         "FVAUL2 technology requires libcryptsetup >= 2.6.0");
```
tbzatek, line 431: missing 'T', two occurences.

tbzatek (approved):
Nice addition!

src/plugins/crypto.c
vojtechtrefny: Fixed.

PR #830, opened 2023-01-04
btrfs: Fix parsing snapshot info without valid otime

vojtechtrefny:
With latest btrfs-progs the "otime" value for snapshots can be empty so we need to adjust our regex for parsing the subvolume information.

With btrfs 6.1:
ID 256 ...level 5 otime - path snap1

With btrfs 5.18:
ID 256 ... level 5 otime 2023-01-04 14:11:41 path snap1


tbzatek (approved):
Ugh.

PR #833, opened 2023-01-06
fs: Fix parsing F2FS info from dump.f2fs >= 1.15 output

vojtechtrefny:
New version of #831 I wasn't able to convince regex to make the sector size optional, so more traditional parsing is it.

### DIFF
--- a/src/plugins/fs/f2fs.c
+++ b/src/plugins/fs/f2fs.c
@@ -279,43 +279,18 @@ BDFSF2FSInfo* bd_fs_f2fs_get_info (const gchar *device, GError **error) {
     const gchar *argv[3] = {"dump.f2fs", device, NULL};
     gchar *output = NULL;
     gboolean success = FALSE;
-    gchar const * const pattern = "[\\S\\s]+" \
-                                  "Info:\\ssector\\ssize\\s=\\s(?P<ssize>\\d+)\\s+" \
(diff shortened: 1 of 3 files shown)


src/plugins/fs/f2fs.c
```diff
@@ -325,28 +300,57 @@ BDFSF2FSInfo* bd_fs_f2fs_get_info (const gchar *device, GError **error) {
...
+    }
 
-    item = g_match_info_fetch_named (match_info, "sectors");
-    ret->sector_count = g_ascii_strtoull (item, NULL, 0);
-    g_free (item);
+    line_p = lines;
+    while (line_p && *line_p && !g_str_has_prefix (*line_p, "Info: total FS sectors"))
+        line_p++;
+    if (!line_p || !(*line_p)) {
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_PARSE, "Failed to parse NILFS2 file system information");
```
tbzatek, line 327: NILFS2?

tbzatek (approved):
Well, the regexp approach wasn't necessarily bad, but it was a maintenance nightmare I believe.

src/plugins/fs/f2fs.c
vojtechtrefny: Oops. You can guess from which file system I copied the code :-)

PR #837, opened 2023-01-09
Backport exFAT support on top of 2.28 

zyga:
It looks like master is brewing for a 3.x release, with API changes and further changes required in downstream projects, and 2.28 is the last stable release. Since I wanted to have exFAT support without waiting for the upcoming release, I've back-ported a patch to a branch based on 2.28-1.

### DIFF
--- a/docs/libblockdev-sections.txt
+++ b/docs/libblockdev-sections.txt
@@ -592,6 +592,15 @@ bd_fs_ntfs_repair
 bd_fs_ntfs_resize
 bd_fs_ntfs_set_label
 bd_fs_ntfs_wipe
+BDFSExfatInfo
+bd_fs_exfat_check
+bd_fs_exfat_get_info
+bd_fs_exfat_info_copy
+bd_fs_exfat_info_free
+bd_fs_exfat_mkfs
+bd_fs_exfat_repair
+bd_fs_exfat_set_label
+bd_fs_exfat_wipe
 </SECTION>
 
 <SECTION>
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -343,7 +343,8 @@ typedef enum {
     BD_FS_TECH_EXT4,
     BD_FS_TECH_XFS,
     BD_FS_TECH_VFAT,
-    BD_FS_TECH_NTFS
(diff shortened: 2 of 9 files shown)


StorageGhoul:

tbzatek:
Hmm, we do hope to release the final 3.0 reasonably soon, right @vojtechtrefny? :-)

zyga:
I hope so too, I'm also sharing this to see if anyone else might need this. Since Yocto Kirkstone has 2.26, and 2.28 is pretty close, it does make the patch useful.

N.B.: I'm still working on testing, this is still a draft.

zyga:
I've marked this as ready, it works for my use-cases.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
> Hmm, we do hope to release the final 3.0 reasonably soon, right @vojtechtrefny? :-)

Yes, we've been planning to release 3.0 "soon" for two years now :-)

vojtechtrefny (commented):
The code looks OK, the tests are also passing, so this should be fine. My only potential problem is that 2.x has the runtime dependency checks enabled by default -- this means to load a plugin all runtime dependencies must be installed so by adding ExFAT support libblockdev will now require exfatutils to be installed to load the FS plugin. Our two biggest users (blivet and udisks) shouldn't be affected (blivet doesn't use the FS plugin and udisks switches the runtime dependency checks off), but it might still be a big change for a stable release. Thoughts @tbzatek @vpodzime ?

vpodzime:
> Thoughts @tbzatek @vpodzime ?

The behavior has been the same in the past releases, *exFAT* is just another FS we do this for. So I don't see a problem here. Any code using libblockdev can do the same thing UDisks2 does.

tbzatek:
> The behavior has been the same in the past releases, _exFAT_ is just another FS we do this for. So I don't see a problem here. Any code using libblockdev can do the same thing UDisks2 does.

Add time factor into the equation and you realize that even previous `libblockdev` releases might suddenly have a problem with newer distros. So I guess adding `exfat` would have only a little impact.

vpodzime (approved):
Looks good to me. (not sure what to review, though)

PR #839, opened 2023-01-18
Remove DMRAID support from the DM plugin

vojtechtrefny:
libdmraid and the dmraid tool are no longer actively maintained and dmraid support will be removed from Blivet (only user of this technology) so we can remove it from libblockdev 3.0 too.

Note that we should merge https://github.com/storaged-project/blivet/pull/1098 before merging this to avoid breaking our CI.


tbzatek (approved):
Nice

PR #844, opened 2023-02-02
Fix BlockDev.DMTech enum after removing DM RAID support

vojtechtrefny:
GI creates wrong member names for enums with just one member, so after removing the RAID technology we get DMTech.DM_TECH_MAP instead of DMTech.MAP.


tbzatek (approved):
Interesting issue, hmm.

poncovka:
Can I ask you for a new release of libblockdev with this fix, please? It is affecting all our CI now.

PR #847, opened 2023-02-10
Run external programs with the C.UTF-8 locale

### LINKED ISSUE #846: Cannot set labels with non-ASCII characters in exFAT
I reproduced this in UDisks but I think that the problem is here. I'm using the master branch in both cases. I confirm that the problem does not happen with the 2.x stable releases.

exFAT filesystems are formatted using `mkfs.exfat` from the `exfatprogs` package. This volume label in this filesystem is a Unicode string ([exFAT spec, section 7.3.3](https://learn.microsoft.com/en-gb/windows/win32/fileio/exfat-specification#733-volumelabel-field)) encoded in UTF-16, but `mkfs.exfat` uses `mbstowcs()` to do the conversion ([here](https://github.com/exfatprogs/exfatprogs/blob/1.2.0/lib/libexfat.c#L234)), which depends on `LC_CTYPE` from the current locale to determine the encoding of the string provided by the user.

However, and even if the OS supports UTF-8 correctly, when libblockdev spawns a process it sets `LC_ALL` to `C` ([here](https://github.com/storaged-project/libblockdev/blob/4a85c9d44f0cd402c1bf5c9ae53d646a3a358fa8/src/utils/exec.c#L215) and [here](https://github.com/storaged-project/libblockdev/blob/4a85c9d44f0cd402c1bf5c9ae53d646a3a358fa8/src/utils/exec.c#L419)), making the conversion fail if the label contains non-ASCII characters ("invalid character sequence in current locale").

bertogg:
Also unset the LANGUAGE variable so it doesn't override the default language (https://sourceware.org/bugzilla/show_bug.cgi?id=16621).

This allows passing a UTF-8 label to mkfs.exfat.

`C.UTF-8` is available since glibc 2.35, and many major distros have supported it downstream for years. However if for some reason `C.UTF-8` is not available then the child's `setlocale()` call will fail and remain with the default (`C`) locale.

Fixes #846

### DIFF
--- a/src/utils/exec.c
+++ b/src/utils/exec.c
@@ -212,7 +212,8 @@ gboolean bd_utils_exec_and_report_status_error (const gchar **argv, const BDExtr
     }
 
     old_env = g_get_environ ();
-    new_env = g_environ_setenv (old_env, "LC_ALL", "C", TRUE);
+    new_env = g_environ_setenv (old_env, "LC_ALL", "C.UTF-8", TRUE);
+    new_env = g_environ_unsetenv (new_env, "LANGUAGE");
 
     task_id = log_running (args ? args : argv);
     success = g_spawn_sync (NULL, args ? (gchar **) args : (gchar **) argv, new_env, G_SPAWN_SEARCH_PATH,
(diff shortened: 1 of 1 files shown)


StorageGhoul:

tbzatek:
Jenkins, ok to test.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny (approved):
Looks good to me, thank you.

Issue #841, opened 2023-01-24
Latest LVM doesn't support shrinking inactive LVs without using --fs ignore


vojtechtrefny:
To protect users from shrinking an LV without shrinking the filesystem on it, latest LVM will refuse to shrink an inactive LV. This can be changed by adding the `--fs ignore` option.

The question is how should `bd_lvm_lvresize` behave. In libblockdev we in general ignore the filesystem and assume that users will use the fs plugin to resize the filesystem first (when shrinking). The existing `-r/--resizefs` option is not part of the API, but users could always use the extra options to add it.

I suggest using the `--fs ignore` option by default and maybe documenting this behaviour.

teigland:
Users incorrectly using lvreduce and corrupting their file systems was the motivation for the change (according to support this happens with surprising frequency.)  I don't know if the users of bd_lvm_lvresize are more careful than users running lvreduce.  Could bd_lvm_lvresize just reflect the same behavior change and return an error if the LV is inactive?

vojtechtrefny:
> I don't know if the users of bd_lvm_lvresize are more careful than users running lvreduce.

It's an API, not a CLI so I would assume people using libblockdev know what they are doing. All the users we have now deal with the resizing manually.

> Could bd_lvm_lvresize just reflect the same behavior change and return an error if the LV is inactive?

What if they want to resize and inactive LV? I might consider changing the `bd_lvm_lvresize` API to include the fs resize options (we are now working on a new major version so we could change the API).

teigland:
> What if they want to resize and inactive LV? I might consider changing the `bd_lvm_lvresize` API to include the fs resize options (we are now working on a new major version so we could change the API).

OK, I think your suggestion of using --fs ignore might be the most reasonable, since it's programmers using this, and the existing pattern is to resize the fs with one API call and the LV with another.  These are clearly different conditions from a user running lvresize.

If you're changing the lvresize API, you might consider adding a flag/parameter to emphasize this issue.   Called something like "reduce_check_fs", when set it would cause "--fs checksize" to be included in the lvresize.  The flag would be documented as requiring the LV to be active to use it.   When set, it would cause lvm to check for a file system using reduced space that would be damaged.  If a fs is found using the space being reduced, then the command would fail without reducing the LV.  Even if the flag is not used by anyone, it may help the API user understand the issue.

tbzatek:
> Could bd_lvm_lvresize just reflect the same behavior change and return an error if the LV is inactive?

[...]

PR #855, opened 2023-02-27
Allow resizing of inactive LVs with latest LVM

vojtechtrefny:
Fixes: #841 

This should be a relatively safe change -- we add the `--fs ignore` option only for inactive LVs basically restoring the old LVM behaviour.


tbzatek (approved):
Sigh, quirks everywhere.

PR #856, opened 2023-03-03
RPM builds using Packit

vojtechtrefny:


vojtechtrefny:
This is all for libblockdev for now -- RPM builds on PRs and daily builds in Copr. I plan to add jobs for downstream builds later, it doesn't really make sense before releasing 3.0.

tbzatek (approved):
Nice!

PR #857, opened 2023-03-08
lvm: Remove multiple public LVM constants from lvm.h

vojtechtrefny:
These are not used by our users and shouldn't be public anyway, we have public "helper" functions for getting thin pool or cache metadata sizes available.
This also removes the libdevmapper dependency from the header file.

Related: https://github.com/storaged-project/udisks/pull/1072


tbzatek (approved):
Interesting.

vojtechtrefny:
I knew the udisks compile job will be useful, I have created https://github.com/storaged-project/udisks/pull/1076 to stop using the `BD_LVM_DEFAULT_CHUNK_SIZE` constant in udisks, `0` is perfectly fine for their use.

PR #858, opened 2023-03-08
spec: Remove dependency on dosfstools and xfsprogs from fs-devel

vojtechtrefny:
This was originally added in d617ba370d2df5dad4a1070aaaf751bba46301c5 but we don't really want to depend on the runtime dependencies and definitely not in the devel package. Libblockdev users needs to deal with the dependencies themselves and use the dependency checks in libblockdev to check for runtime dependencies.


tbzatek (approved):
Yeah, but there's still the ext2 libs dependency and similar libs for other filesystems may appear as hard deps...

vojtechtrefny:
> Yeah, but there's still the ext2 libs dependency and similar libs for other filesystems may appear as hard deps...

Yeah, I am still not sure how I feel about that, but that's a build time dependency so it's different.

PR #861, opened 2023-04-06
tests: Fix test for inactive LV resize

vojtechtrefny:
The issue we were fixing in #855 actually happens only when shrinking the LV, so the test actually didn't test anything.


tbzatek (approved):

vojtechtrefny:
The tests are failing on C9S because the LVM resize feature was backported to 2.03.17 there so the check for the version >= 2.03.19 doesn't work. I don't want to add another check (which might be impossible because 2.03.17 on Fedora doesn't have the feature so using the `--fs` option will also result in an error...) so I'll probably just ignore this until the latest LVM will make it to C9S (which should be soon).

PR #862, opened 2023-04-11
fs: Don't require XFS filesystem to be mounted when getting info

mvollmer:
It's not necessary and potentially interferes with other activities on the system.

(For example, when Cockpit is open while a new Stratis snapshot is created, there will be a well timed request to UDisks2 that ultimately mounts the new snapshot filesystem to get its size, right when Stratis runs xfs_db to change the UUID, which will fail because it is unexpectedly mounted.  Amazingly narrow race, but it happens.)

### DIFF
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -1275,32 +1275,6 @@ gboolean bd_fs_set_uuid (const gchar *device, const gchar *uuid, const gchar *fs
     return device_operation (device, fstype, BD_FS_UUID, 0, NULL, uuid, error);
 }
 
-static BDFSXfsInfo* xfs_get_info (const gchar *device, GError **error) {
-    g_autofree gchar* mountpoint = NULL;
-    gboolean unmount = FALSE;
-    gboolean ret = FALSE;
-    GError *local_error = NULL;
-    BDFSXfsInfo* xfs_info = NULL;
-
-    mountpoint = fs_mount (device, "xfs", &unmount, error);
-    if (!mountpoint)
-        return NULL;
-
-    xfs_info = bd_fs_xfs_get_info (device, error);
-
-    if (unmount) {
-        ret = bd_fs_unmount (mountpoint, FALSE, FALSE, NULL, &local_error);
-        if (!ret) {
-            bd_utils_log_format (BD_UTILS_LOG_INFO, "Failed to unmount %s after getting information about it: %s", device, local_error->message);
-            g_clear_error (&local_error);
-        } else
-            if (g_rmdir (mountpoint) != 0)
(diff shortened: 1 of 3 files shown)


StorageGhoul:

mvollmer:
This is the test failure: https://cockpit-logs.us-east-1.linodeobjects.com/pull-18616-20230411-070053-44f522a8-fedora-testing-daily/log.html#348-2

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
It's interesting that `xfs_info` used to work on unmounted devices, but that changed couple years ago (see 2773109bb4c1f65d7bf8ce21131a3d08a1acd300) and then it changed again a year later https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/commit/?h=6750f2b8aa83091ec9b03a9acdb2fe659cef5fec :)

tbzatek:
> It's interesting that `xfs_info` used to work on unmounted devices, but that changed couple years ago (see [2773109](https://github.com/storaged-project/libblockdev/commit/2773109bb4c1f65d7bf8ce21131a3d08a1acd300)) and then it changed again a year later https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/commit/?h=6750f2b8aa83091ec9b03a9acdb2fe659cef5fec :)

Bah. I would also prefer not to mount any filesystem automatically, apart from creating new race conditions elsewhere there's always possibility to mess things up (think of journal recovery). Also, why wasn't the filesystem mounted readonly in this case?

There are more cases calling `fs_mount()` just to get fs info, e.g. btrfs. Shall we get rid of those for libblockdev-3?

tbzatek (approved):

vojtechtrefny:
> There are more cases calling `fs_mount()` just to get fs info, e.g. btrfs. Shall we get rid of those for libblockdev-3?

What do you suggest to do if it isn't possible to get the information? Return an error? I think the solution is to not use these potentially "dangerous" calls from udisks when reacting to an uevent, which should be possible with latest blkid reading filesystem size and free space directly from the metadata.

tbzatek:
> What do you suggest to do if it isn't possible to get the information? Return an error? 

Yes, I was thinking about moving this responsibility to the caller, to handle mount+unmount on his side and deal with any errors. Just an idea, it would make your life easier!

> I think the solution is to not use these potentially "dangerous" calls from udisks when reacting to an uevent, which should be possible with latest blkid reading filesystem size and free space directly from the metadata.

Absolutely, that's the first thing to do.

mvollmer:
Test is green now, thanks!

https://cockpit-logs.us-east-1.linodeobjects.com/pull-18616-20230417-081125-1978a64f-fedora-testing-dnf-copr/log.html#185

(Well, it's yellow because of a unrelated regression in stratisd, but the snapshot part passes now.)

Issue #423, opened 2019-02-07
Switch to runtime dependency checks by default


vojtechtrefny:
Currenctly libblockdev checks all dependencies during `init`. With more new runtime dependencies, this is no longer the best option, especially for plugins like `fs` with many dependencies that are not always available (like *ntfsprogs*). That's why we have option to run the checks only during runtime if user actually tries to run function that needs some of the dependencies. We should switch to this behaviour by default and keep init checks optional.

tbzatek:
FYI the checks in `master/src/plugins/fs.c:bd_fs_check_deps()` looks weird - the `ret = ret && bd_fs_xfs_is_tech_avail(...)` tests will always fail if `ret` is already `FALSE`.

I'm somewhat missing the point here, failing to understand usefulness of the return value. My suggestion is to do somewhat deeeper refactoring here...

PR #863, opened 2023-04-12
Remove runtime dependency checks when loading plugins

vojtechtrefny:
Currently all runtime dependencies must be available in order to load a plugin. With plugins now covering multiple technologies this doesn't really make sense and all our users currently disable this feature so it makes sense to remove this completely and let users check for runtime dependencies on demand.

Fixes: #423


tbzatek:
Having both `bd_X_check_deps()` and `bd_X_is_tech_avail()` in place is quite confusing so I'm for simplification of the API. The `bd_X_is_tech_avail()` should cover most of the checking cases.

These are mostly runtime dependencies (somewhat optional), the linking deps are perhaps more important.

tbzatek (approved):
Looks good!

tests/btrfs_test.py
```diff
@@ -36,6 +35,9 @@ def setUpClass(cls):
         else:
             BlockDev.reinit(cls.requested_plugins, True, None)
 
+
+class BtrfsTestCase(BtrfsTest):
```
vpodzime, line 39: I think these deserve a better name. If someone goes and adds a new test here, they can easily use `BtrfsTest` as the base class.

docs/3.0-api-changes.xml
```diff
@@ -5,6 +5,16 @@
   This is a list of API changes in libblockdev 3.0 with suggested replacements.
   </para>
 
+  <simplesect><title>Library</title>
+  <para>
+    <literal>bd_switch_init_checks()</literal> has been removed. Checks for runtime dependencies during libblockdev initialization are now disabled and plugins will be loaded even if a runtime dependency is missing.
```
vpodzime, line 10: I'd make this **bold** because this is a quite a big change in the default behavior.

vpodzime (approved):
Looks good to me otherwise.

src/python/gi/overrides/BlockDev.py
```diff
@@ -1123,6 +1123,12 @@ def crypto_luks_uuid(device):
 __all__.append("crypto_luks_uuid")
 
 
+def switch_init_checks(enable):
+    # no-op, init checks are no longer supported
+    return
```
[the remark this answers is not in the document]
vojtechtrefny: This is just a temporary function, I will remove it after releasing 3.0. This is here just to make sure blivet tests in our CI won't start failing after merging this. (There is a comment explaining this above, there are two more functions/overrides like this.)

tests/btrfs_test.py
vojtechtrefny: I plan to do a big test "cleanup" before the 3.0 release, there is a lot of outdated/messy things there. I kept this simple, we have `XXXTestCase` and `XXXTest` for other plugins too.

docs/3.0-api-changes.xml
vojtechtrefny: Ok. Big changes like these will be documented in release notes too.

PR #866, opened 2023-04-14
fs: Don't require XFS filesystem to be mounted when getting info

vojtechtrefny:
Backport of #862 to 2.x


tbzatek (approved):
Fine with me I guess, except of possibly hitting those xfsprogs versions that required the mount.

vojtechtrefny:
> Fine with me I guess, except of possibly hitting those xfsprogs versions that required the mount.

It was fixed in 4.17, even Debian oldstable has 4.20 (and RHEL 8 has 5.0) so we should be fine.

PR #867, opened 2023-04-14
tests: Make sure the "fake" PATH for tests exists

vojtechtrefny:
It's better to fail with an error that run the test case that expects $PATH to be set to a specific (possibly nonexisting) path.


tbzatek (approved):
Hmm, looks weird but I believe you.

PR #869, opened 2023-04-19
part: Fix segfault when adding a partition too big for MSDOS

vojtechtrefny:
Fixes: #564


tbzatek (approved):
Yay, decrypting undocumented return codes for corner cases!

PR #871, opened 2023-04-20
fs/xfs: Open device read-only for xfs_info

mvollmer:
Running xfs_infor for a device that is not mounted ultimately executes "xfs_db -c info". Unfortunately, xfs_db opens the device read-write by default. This results in a udev event.

The udev event in turn might result in another call to xfs_info via UDisks2 and Cockpit, which conspire the requests the size of a filesystem for each uevent on it.

Luckily, xfs_db can be told to open the device read-only, with the "-r" option.


StorageGhoul:

tbzatek:
Jenkins, ok to test.

vojtechtrefny (approved):

mvollmer:
It never ends... :-/

PR #874, opened 2023-04-24
lvm: Add a function to activate LVs in shared mode

vojtechtrefny:
Needed by the new blivet feature to support shared LVM setups.

----

Note: I don't intent to backport it to 2.x, but this will get backported to RHEL so the purpose for this PR is mostly to run the tests in the CI and to get a review.

### DIFF
--- a/src/lib/plugin_apis/lvm.api
+++ b/src/lib/plugin_apis/lvm.api
@@ -1056,6 +1056,22 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
 gboolean bd_lvm_lvactivate (const gchar *vg_name, const gchar *lv_name, gboolean ignore_skip, const BDExtraArg **extra, GError **error);
 
 /**
+ * bd_lvm_lvactivate_shared:
+ * @vg_name: name of the VG containing the to-be-activated LV
+ * @lv_name: name of the to-be-activated LV
+ * @ignore_skip: whether to ignore the skip flag or not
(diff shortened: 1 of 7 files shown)


src/lib/plugin_apis/lvm.api
```diff
@@ -1055,6 +1055,22 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
  */
 gboolean bd_lvm_lvactivate (const gchar *vg_name, const gchar *lv_name, gboolean ignore_skip, const BDExtraArg **extra, GError **error);
 
+/**
+ * bd_lvm_lvactivate_shared:
+ * @vg_name: name of the VG containing the to-be-activated LV
+ * @lv_name: name of the to-be-activated LV
+ * @ignore_skip: whether to ignore the skip flag or not
+ * @shared: whether to activate the LV in shared mode
```
tbzatek, line 1063: Isn't this argument somewhat misleading since you have `_shared` in the function name already?

tbzatek (approved):
Yeah, I think it's good anyway.

src/lib/plugin_apis/lvm.api
vojtechtrefny: I am open to suggestions for a better name. To me this means "new version of the function with `shared` agument".

vojtechtrefny: I understand, but with the new function having the `shared` argument I can just call it and pass a "shared" variable to it instead of using a condition.

Issue #882, opened 2023-05-04
RFE: UUID support for swap


tbzatek:
Now that UDisks is gaining support for setting UUID during `mkfs` and subsequent setting new UUID, and that `swap` is treated by special fstype during `Block.Format()` method, it would be nice to extend the swap plugin with UUID functionality, making it a primary supported attribute just like `label`.

- [ ] extend `bd_swap_mkswap()` with UUID argument
- [ ] add new `bd_swap_set_uuid()`

PR #883, opened 2023-05-05
UUID support for swap

vojtechtrefny:
Fixes: #882

### DIFF
--- a/src/lib/plugin_apis/swap.api
+++ b/src/lib/plugin_apis/swap.api
@@ -25,6 +25,7 @@ typedef enum {
     BD_SWAP_TECH_MODE_ACTIVATE_DEACTIVATE = 1 << 1,
     BD_SWAP_TECH_MODE_QUERY               = 1 << 2,
     BD_SWAP_TECH_MODE_SET_LABEL           = 1 << 3,
+    BD_SWAP_TECH_MODE_SET_UUID            = 1 << 3,
 } BDSwapTechMode;
 
 /**
@@ -99,3 +100,15 @@ gboolean bd_swap_swapstatus (const gchar *device, GError **error);
  * Tech category: %BD_SWAP_TECH_SWAP-%BD_SWAP_TECH_MODE_SET_LABEL
  */
 gboolean bd_swap_set_label (const gchar *device, const gchar *label, GError **error);
+
+/**
+ * bd_swap_set_uuid:
+ * @device: a device to set UUID on
+ * @uuid: UUID that will be set
+ * @error: (out) (optional): place to store error (if any)
+ *
+ * Returns: whether the UUID was successfully set or not
+ *
+ * Tech category: %BD_SWAP_TECH_SWAP-%BD_SWAP_TECH_MODE_SET_UUID
+ */
+gboolean bd_swap_set_uuid (const gchar *device, const gchar *uuid, GError **error);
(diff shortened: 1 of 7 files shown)


src/lib/plugin_apis/swap.api
```diff
@@ -42,6 +42,7 @@ gboolean bd_swap_is_tech_avail (BDSwapTech tech, guint64 mode, GError **error);
  * bd_swap_mkswap:
  * @device: a device to create swap space on
  * @label: (nullable): a label for the swap space device
+ * @uuid: (nullable): UUID for the swap space device
```
tbzatek, line 46: Looking at the `mkswap` manpage, three special values are accepted: `clear`, `random` and `time`. Does it make sense to document them here or just put a reference to the manpage? The UUID string often bears some specific format, but the usual `c1b9d5a2-f162-11cf-9ece-0020afc76f16` is being accepted here.

src/lib/plugin_apis/swap.api
```diff
@@ -99,3 +100,15 @@ gboolean bd_swap_swapstatus (const gchar *device, GError **error);
  * Tech category: %BD_SWAP_TECH_SWAP-%BD_SWAP_TECH_MODE_SET_LABEL
  */
 gboolean bd_swap_set_label (const gchar *device, const gchar *label, GError **error);
+
+/**
+ * bd_swap_set_uuid:
```
tbzatek, line 105: Needs to be added to `docs/libblockdev-sections.txt`?

tbzatek (approved):
Very nice, thank you! :heart_eyes:

src/lib/plugin_apis/swap.api
vojtechtrefny: I swa that too, I don't think we need to list it here, I don't really see the use case for the special options in `mkswap` (`random` is default and `clear` doesn't really make sense when creating a new swap). `swaplabel` doesn't support these.

src/lib/plugin_apis/swap.api
vojtechtrefny: Fixed.

src/lib/plugin_apis/swap.api
tbzatek: Fair enough, little chance of using these anyway I suppose.

src/plugins/swap.h
```diff
@@ -40,10 +41,11 @@ void bd_swap_close (void);
 
 gboolean bd_swap_is_tech_avail (BDSwapTech tech, guint64 mode, GError **error);
 
-gboolean bd_swap_mkswap (const gchar *device, const gchar *label, const BDExtraArg **extra, GError **error);
+gboolean bd_swap_mkswap (const gchar *device, const gchar *label, const gchar *uuid, const BDExtraArg **extra, GError **error);
 gboolean bd_swap_swapon (const gchar *device, gint priority, GError **error);
 gboolean bd_swap_swapoff (const gchar *device, GError **error);
 gboolean bd_swap_swapstatus (const gchar *device, GError **error);
 gboolean bd_swap_set_label (const gchar *device, const gchar *label, GError **error);
+gboolean bd_seap_set_uuid (const gchar *device, const gchar *uuid, GError **error);
```
tbzatek, line 49: Uhm, typo!

tbzatek (changes requested):

tbzatek (approved):
Tested, works fine now, thanks

PR #884, opened 2023-05-05
fs: Add support for setting and checking UUID/volume ID for VFAT

vojtechtrefny:
Support for changing volume ID was added in dosfstools 4.2

### DIFF
--- a/src/plugins/fs/vfat.c
+++ b/src/plugins/fs/vfat.c
@@ -39,25 +39,28 @@ static GMutex deps_check_lock;
 #define DEPS_FSCKVFAT_MASK (1 << DEPS_FSCKVFAT)
 #define DEPS_RESIZEVFAT 3
 #define DEPS_RESIZEVFAT_MASK (1 << DEPS_RESIZEVFAT)
+#define DEPS_FATLABELUUID 4
+#define DEPS_FATLABELUUID_MASK (1 << DEPS_FATLABELUUID)
 
-#define DEPS_LAST 4
+#define DEPS_LAST 5
 
 static const UtilDep deps[DEPS_LAST] = {
     {"mkfs.vfat", NULL, NULL, NULL},
     {"fatlabel", NULL, NULL, NULL},
     {"fsck.vfat", NULL, NULL, NULL},
     {"vfat-resize", NULL, NULL, NULL},
+    {"fatlabel", "4.2", "--version", "fatlabel\\s+([\\d\\.]+).+"},
 };
 
 static guint32 fs_mode_util[BD_FS_MODE_LAST+1] = {
-    DEPS_MKFSVFAT_MASK,   /* mkfs */
-    0,                    /* wipe */
-    DEPS_FSCKVFAT_MASK,   /* check */
-    DEPS_FSCKVFAT_MASK,   /* repair */
-    DEPS_FATLABEL_MASK,   /* set-label */
-    DEPS_FSCKVFAT_MASK,   /* query */
-    DEPS_RESIZEVFAT_MASK, /* resize */
-    0                     /* set-uuid */
+    DEPS_MKFSVFAT_MASK,     /* mkfs */
+    0,                      /* wipe */
+    DEPS_FSCKVFAT_MASK,     /* check */
+    DEPS_FSCKVFAT_MASK,     /* repair */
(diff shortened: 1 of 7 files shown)


tbzatek (commented):
This needs to be changed as well: https://github.com/storaged-project/libblockdev/blob/7e7c224852751ed293cd2cf65886401482f2c771/src/plugins/fs/generic.c#L236

src/plugins/fs/vfat.c
```diff
@@ -328,6 +325,60 @@ gboolean bd_fs_vfat_check_label (const gchar *label, GError **error) {
...
+ *
+ * Returns: whether @uuid is a valid UUID for the vfat file system or not
+ *          (reason is provided in @error)
+ *
+ * Tech category: always available
+ */
+gboolean bd_fs_vfat_check_uuid (const gchar *uuid, GError **error) {
+    guint64 vol_id;
+
+    vol_id = g_ascii_strtoull (uuid, NULL, 16);
```
tbzatek, line 366: I was also wondering whether we should accept VOLID strings as presented by udev - i.e. not just a plain 32-bit hex number, but also the `XXXX-YYYY` form. I would certainly like to be this way for UDisks since we present it that form and thinking about the right place to reformat the ID.

This needs to be consistent across all filesystem plugins. Most of them accept everything `uuid_parse()` accepts such as `%08x-%04x-%04x-%04x-%012x`, however e.g. NTFS also needs hex string without dashes.

tbzatek: Another problem here is that `g_ascii_strtoull()` can stop conversion halfways, returning partially converted value. `endptr` needs to be checked (the second argument) and proper error returned.

For example: input string `28AA-7D6D` gets converted to `28AA` and resulting VOLID becomes `0000-28AA`.

tbzatek:
Also, `exfatlabel` has a switch `--volume-serial`, could you please add similar support please?  :smiling_face_with_three_hearts:

src/plugins/fs/vfat.c
vojtechtrefny: > I was also wondering whether we should accept VOLID strings as presented by udev - i.e. not just a plain 32-bit hex number, but also the `XXXX-YYYY` form. I would certainly like to be this way for UDisks since we present it that form and thinking about the right place to reformat the ID.

Ok, I changed the code to accept vol IDs with the dash in the middle.

> 
> This needs to be consistent across all filesystem plugins. Most of them accept everything `uuid_parse()` accepts such as `%08x-%04x-%04x-%04x-%012x`, however e.g. NTFS also needs hex string without dashes.

Yes, but for NTFS you get the ID from blkid/udev without dashes, e.g. `4A514A2C13B7198B`.

> Another problem here is that `g_ascii_strtoull()` can stop conversion halfways, returning partially converted value. `endptr` needs to be checked (the second argument) and proper error returned.
> 
> For example: input string `28AA-7D6D` gets converted to `28AA` and resulting VOLID becomes `0000-28AA`.

Oh, I didn't realize that, I thought it will fail to parse it. It should be fixed now.

tbzatek: At the end I've made udisks to remove any dashes for certain filesystem types, so we should be covered from both sides.

src/plugins/fs/vfat.c
```diff
@@ -328,6 +325,89 @@ gboolean bd_fs_vfat_check_label (const gchar *label, GError **error) {
...
+    len = strlen (uuid);
+    if (len == 9 && uuid[4] == '-') {
+        new_uuid = g_new0 (gchar, 9);
+        memcpy (new_uuid, uuid, 4);
+        memcpy (new_uuid + 4, uuid + 5, 4);
+    } else
+        new_uuid = g_strdup (uuid);
+
+    vol_id = g_ascii_strtoull (new_uuid, &endptr, 16);
+    if (vol_id == 0 || *endptr) {
```
tbzatek, line 393: Hmm, is `0000-0000` valid UUID? I'd also suggest to check for non-NULL `endptr` before dereferencing it.

vojtechtrefny: > Hmm, is `0000-0000` a valid UUID?

Yes. Or at least `fatlabel` is not complaining about it and will set it.

tbzatek (approved):
No objections for the current code.

PR #889, opened 2023-05-17
mount: Initialize libmount debugging

tbzatek:
This little trick makes libmount to initialize debugging through the LIBMOUNT_DEBUG environment variable.

Tested on UDisks, setting this env. var. in the daemon environment prints lots of information that can be later retrieved from syslog/journal. Daemon restart required.

### DIFF
--- a/src/plugins/fs.c
+++ b/src/plugins/fs.c
@@ -18,6 +18,7 @@
  */
 
 #include <blockdev/utils.h>
+#include <libmount/libmount.h>
 
 #include <check_deps.h>
 #include "fs.h"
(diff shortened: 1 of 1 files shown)


src/plugins/fs/mount.c
```diff
@@ -140,6 +140,7 @@ static gboolean do_unmount (MountArgs *args, GError **error) {
     int ret = 0;
     gboolean success = FALSE;
 
+    mnt_init_debug (0);
```
vojtechtrefny, line 143: Mounting is only a small part of the FS plugin, but I think moving this to `bd_fs_init` won't hurt.

tbzatek: Good idea, this will actually setup debugging early on udisksd startup, spitting debug output even for parsing fstab, mountinfo etc.

tbzatek:
Jenkins, test this please.

vojtechtrefny (approved):

PR #891, opened 2023-05-18
tests: Use short argument switches for nvme connect

tbzatek:
Some weird temporary bug in nvme-cli-2.2.1 caused commandline arguments parsing to fail on long switch names.


tbzatek:
Hmm, looks like a miscompilation of `nvme-cli-2.2.1-1.fc37.x86_64.rpm` in the Fedora buildsystem

PR #892, opened 2023-05-19
tests: Skip NVMe tests on CentOS/RHEL 8

vojtechtrefny:
The required version of libnvme is not available on CentOS/RHEL 8.


tbzatek (approved):

tbzatek (approved):
Nice, looks like the tests have finally passed clean.

PR #906, opened 2023-06-26
fs: Rename BDFs* enums to BDFS*

vojtechtrefny:
Just to be consistent have everything named with the capital FS.


tbzatek (approved):
Yeah, this looks more tidy

PR #911, opened 2023-06-27
Reorder enums

### LINKED ISSUE #351: Make sure last enum values don't change when adding new ones before them
For example in #350, new enum values are added. Which means ``BD_NVDIMM_NAMESPACE_MODE_UNKNOWN``, which should logically be last, now is a different integer value which is an ABI change. We should avoid this in the future by:

a) assign the enum values like the above one some high integer that won't change or
b) assign all enum values explicit integer values using high values for the values that should stay last.

tbzatek:
Codestyle fixes, reordering enum values with the rule: TECH_UNAVAIL first, generic FAILED second.

Fixes #351


vojtechtrefny (approved):
LGTM

vpodzime:
Shouldn't we also do `= 0` with all the first values in those enums?

tbzatek:
> Shouldn't we also do `= 0` with all the first values in those enums?

Hmm, is there any real reason for that? I haven't read the C standard though.

Issue #916, opened 2023-06-28
Build fails when libparted-dev is missing


mbiebl:
Version: 3.0

According to configure.ac, libparted is only required when building with `--with-tools`.
But building with `--without-tools` and libparted missing results in:
```
[...]

PR #919, opened 2023-06-29
vdo_stats: Remove unused libparted include

vojtechtrefny:
I have no idea where this came from, but we don't need (and never did) libparted in vdo_stats.

Fixes: #916


tbzatek (approved):
Awesome!!

PR #922, opened 2023-06-29
dist: Sync spec with downstream

vojtechtrefny:


tbzatek (approved):
Yup, not decided with the Provides/Obsoletes in udisks yet.

tbzatek (approved):

PR #927, opened 2023-07-02
loop: define LOOP_SET_BLOCK_SIZE is not defined

giuliobenetti:
It can happen that linux header linux/loop.h doesn't include LOOP_SET_BLOCK_SIZE definition, so let's define it locally in such case.

### DIFF
--- a/src/plugins/loop.c
+++ b/src/plugins/loop.c
@@ -28,6 +28,10 @@
 #include <blockdev/utils.h>
 #include "loop.h"
 
+#ifndef LOOP_SET_BLOCK_SIZE
+#define LOOP_SET_BLOCK_SIZE	0x4C09
+#endif
+
 /**
  * SECTION: loop
  * @short_description: plugin for operations with loop devices


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

vojtechtrefny:
Out of curiosity, on which systems is this a problem? I didn't expect people to run latest libblockdev on kernel < 4.14. Note that we plan to keep supporting the 2.x release in the future too so it is not necessary to update to 3.0 on older/stable systems.

giuliobenetti:
> Out of curiosity, on which systems is this a problem? I didn't expect people to run latest libblockdev on kernel < 4.14. Note that we plan to keep supporting the 2.x release in the future too so it is not necessary to update to 3.0 on older/stable systems.

This happens under Buildroot while bumping version. Considering the amount of new dependencies for version 3.x and the fact that you will maintain version 2.x I think I will add a second package called udisks3 beside package udisks. Anyway the goal of Buildroot is to support packages in any case if possible, so version 3.x should be compatible with any Linux(or other packages) versions.

tbzatek:
> I think I will add a second package called udisks3 beside package udisks.

I certainly hope you wanted to say `libblockdev3` as `udisks3` might come in the future with completely different API. To recap: `udisks2` needs `libblockdev3` at the moment. It might happen that any of the future `udisks2` releases will require `libblockdev4`, these two packages are not supposed to be aligned by versioning. Better to be clear now than sorry later.

giuliobenetti:
> > I think I will add a second package called udisks3 beside package udisks.
> 
> I certainly hope you wanted to say `libblockdev3` as `udisks3` might come in the future with completely different API. To recap: `udisks2` needs `libblockdev3` at the moment. It might happen that any of the future `udisks2` releases will require `libblockdev4`, these two packages are not supposed to be aligned by versioning. Better to be clear now than sorry later.

Pardon, I've mixed packages up. It is as you've pointed. I need libblockdev3 for udisks2.10.0 at the moment, so what I will do is to simply update libblockdev3 for udisks2.10.0 without adding any other package.

tbzatek:
Good, thanks for confirming :-)

PR #929, opened 2023-07-03
clang fixes

### LINKED ISSUE #925: plugins/fs/generic.c fails to build with clang - invalid C
Initializers like `.mkfs_options = fs_features[BD_FS_TECH_EXT2].mkfs,` are not well defined C, as the designated initializer at file scope is required to be a constant expression. GCC accepts it, clang does not.

tbzatek:
The last commit deserves some testing.

Fixes #925


vojtechtrefny (approved):
Looks good to me.

Issue #918, opened 2023-06-29
Many private symbols exported which are not part of the public API


mbiebl:
This is a follow-up for #366 
Rerunning the check, lists the following exported symbols, which are missing from the header files:
```
bd_btrfs_device_info_get_type
bd_btrfs_filesystem_info_get_type
bd_btrfs_subvolume_info_get_type
bd_crypto_bitlk_info_get_type
bd_crypto_integrity_extra_get_type
bd_crypto_integrity_info_get_type
bd_crypto_keyslot_context_get_type
bd_crypto_luks_extra_get_type
bd_crypto_luks_info_get_type
bd_crypto_luks_pbkdf_get_type
bd_crypto_luks_token_info_get_type
bd_fs_btrfs_info_get_type
bd_fs_exfat_info_get_type
bd_fs_ext2_info_get_type
bd_fs_ext3_info_get_type
bd_fs_ext4_info_get_type
bd_fs_f2fs_info_get_type
bd_fs_features_copy
bd_fs_features_free
bd_fs_features_get_type
bd_fs_mkfs_options_copy
bd_fs_mkfs_options_free
bd_fs_mkfs_options_get_type
bd_fs_nilfs2_info_get_type
bd_fs_ntfs_info_get_type
bd_fs_udf_info_get_type
bd_fs_vfat_info_get_type
bd_fs_xfs_info_get_type
bd_loop_get_autoclear
bd_loop_info_get_type
bd_lvm_cache_stats_get_type
bd_lvm_lvdata_get_type
bd_lvm_pvdata_get_type
bd_lvm_segdata_copy
bd_lvm_segdata_free
bd_lvm_segdata_get_type
bd_lvm_vdolvpoolname
bd_lvm_vdopooldata_get_type
bd_lvm_vdo_stats_get_type
bd_lvm_vgdata_get_type
bd_md_detail_data_get_type
bd_md_examine_data_get_type
bd_nvdimm_namespace_info_get_type
bd_nvme_connect
bd_nvme_controller_info_copy
bd_nvme_controller_info_free
bd_nvme_controller_info_get_type
bd_nvme_device_self_test
bd_nvme_disconnect
bd_nvme_disconnect_by_path
bd_nvme_error_log_entry_copy
bd_nvme_error_log_entry_free
bd_nvme_error_log_entry_get_type
bd_nvme_error_quark
bd_nvme_find_ctrls_for_ns
bd_nvme_format
bd_nvme_generate_host_nqn
[...]

PR #933, opened 2023-07-05
Restrict list of exported symbols via -export-symbols-regex

mbiebl:
While this approach doesn't cover every exported symbol (for that we should probably look into a version script), it at least catches the most common mistakes like in https://github.com/storaged-project/libblockdev/issues/918#issuecomment-1620489927

### DIFF
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -15,7 +15,7 @@ libblockdev_la_CFLAGS += $(GOBJECT_CFLAGS)
 libblockdev_la_LIBADD += $(GOBJECT_LIBS)
 endif
 
-libblockdev_la_LDFLAGS = -L${srcdir}/../utils/ -version-info 3:0:0 -Wl,--no-undefined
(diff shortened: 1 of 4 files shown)


StorageGhoul:

mbiebl:
A more elaborate fix would probably involve symbol versioning and a version script. Not sure how well that would work with stub symbols and all. But if this is something you'd prefer, I could at least try to take a look.

The `-export-symbols-regex` is a rather cheap "hack" to avoid a bit of the whack-a-mole game we've been playing.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
> The `-export-symbols-regex` is a rather cheap "hack" to avoid a bit of the whack-a-mole game we've been playing.

I think this is a good temporary solution, we definitely need to do better in the future. I was thinking about adding the check you are running to our CI. It shouldn't be that hard to run it on the Ubuntu runners in GH actions.

vojtechtrefny (approved):

tbzatek:
Oh nice, good idea.

Issue #934, opened 2023-07-14
slow bd_fs_ntfs_get_info causing timeouts to udisks2 clients and hanging KDE


amurzeau:
Hi,

When starting up KDE with NTFS disks or when connecting a NTFS USB key, udisks2 2.10 calls bd_fs_ntfs_get_info (via bd_fs_get_size) which will run ntfscluster on the NTFS partition to get the volume size and free space.

But ntfscluster can very long time (45s for my NTFS USB key) causing hangs and timeouts, for example:
```
$ time udisksctl status
Error connecting to the udisks daemon: Timeout was reached

real    0m25,033s
user    0m0,000s
sys     0m0,010s
```

It seems that the `ntfsinfo -m /dev/sdb1` tool is sufficient to get the volume total size and free space and is also faster (177ms instead of 45s on the same USB key).
Output sample:
```
Volume Information 
        Name of device: /dev/sdb1
        Device state: 11
        Volume Name: USB5
        Volume State: 91
        Volume Flags: 0x0080
        Volume Version: 3.1
        Sector Size: 512
        Cluster Size: 4096
        Index Block Size: 4096
        Volume Size in Clusters: 15314943
MFT Information 
        MFT Record Size: 1024
        MFT Zone Multiplier: 0
        MFT Data Position: 24
        MFT Zone Start: 786432
        MFT Zone End: 2700799
        MFT Zone Position: 786432
        Current Position in First Data Zone: 2700799
        Current Position in Second Data Zone: 0
        Allocated clusters 246080 (1,6%)
        LCN of Data Attribute for FILE_MFT: 786432
        FILE_MFTMirr Size: 4
        LCN of Data Attribute for File_MFTMirr: 2
        Size of Attribute Definition Table: 2560
        Number of Attached Extent Inodes: 0
FILE_Bitmap Information 
        FILE_Bitmap MFT Record Number: 6
        State of FILE_Bitmap Inode: 80
        Length of Attribute List: 0
        Number of Attached Extent Inodes: 0
FILE_Bitmap Data Attribute Information
        Decompressed Runlist: not done yet
        Base Inode: 6
        Attribute Types: not done yet
        Attribute Name Length: 0
        Attribute State: 3
        Attribute Allocated Size: 1916928
        Attribute Data Size: 1914368
        Attribute Initialized Size: 1914368
        Attribute Compressed Size: 0
        Compression Block Size: 0
        Compression Block Size Bits: 0
        Compression Block Clusters: 0
[...]

PR #935, opened 2023-07-14
Use ntfsinfo instead of ntfscluster for faster bd_fs_ntfs_get_info

vojtechtrefny:
ntfscluster can take as much as 45s to complete on a USB key.

Use ntfsinfo to get size information instead of ntfscluster, ntfsinfo is near instantanous (177ms instead of 45s on the same USB key).

Fixes: #934

### DIFF
--- a/src/plugins/fs/ntfs.c
+++ b/src/plugins/fs/ntfs.c
@@ -20,6 +20,7 @@
 #include <blockdev/utils.h>
 #include <check_deps.h>
 #include <string.h>
+#include <stdio.h>
 
 #include "ntfs.h"
 #include "fs.h"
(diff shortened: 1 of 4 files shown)


src/plugins/fs/ntfs.c
```diff
@@ -400,8 +402,23 @@ BDFSNtfsInfo* bd_fs_ntfs_get_info (const gchar *device, GError **error) {
...
+        g_set_error (error, BD_FS_ERROR, BD_FS_ERROR_PARSE, "Failed to parse NTFS file system information");
+        g_strfreev (lines);
+        bd_fs_ntfs_info_free (ret);
+        return NULL;
+    }
+
+    /* extract data from something like this: "Cluster Size: 4096" */
+    val_start = strchr (*line_p, ':');
+    val_start++;
+    cluster_size = g_ascii_strtoull (val_start, NULL, 0);
```
vpodzime, line 419: This should, ideally, do some error checking (and the same applies to `g_ascii_strtoull()` calls below.

vpodzime (approved):
Looks good to me either way as this doesn't introduce new issues.

PR #939, opened 2023-07-18
fs: Use read-only mount where possible for generic FS functions

vojtechtrefny:
Related: https://github.com/storaged-project/udisks/pull/1146


tbzatek (approved):
This is a step the right direction. However I suppose this will greatly depend on the particular filesystem tool version (or perhaps even kernel driver) but let's hide this detail from the user for now.

PR #941, opened 2023-07-18
loop: Handle empty devices

tbzatek:


tests/loop_test.py
```diff
@@ -57,6 +57,10 @@ def test_loop_setup_teardown_basic(self):
         succ = BlockDev.loop_teardown(self.loop)
         self.assertTrue(succ)
 
+        # test that we get BD_LOOP_ERROR_DEVICE (code 2)
+        msg = r"Failed to get status of the device loop.*: No such device or address \(2\)"
+        with self.assertRaisesRegex(GLib.GError, msg):
+            info = BlockDev.loop_info(self.loop)
```
vojtechtrefny, line 63: Unfortunately this seems to not work reliably.

```
======================================================================
FAIL: test_loop_setup_teardown_basic (loop_test.LoopTestSetupBasic)
Verify that basic loop_setup and loop_teardown work as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jenkins/workspace/libblockdev-pr/arch/x86_64/distro/centos_9s/type/libblockdev/tests/loop_test.py", line 63, in test_loop_setup_teardown_basic
    info = BlockDev.loop_info(self.loop)
AssertionError: Error not raised
```

tbzatek: Removed.

Issue #949, opened 2023-07-25
 libblockdev3 leaks empty directories in bd_fs_get_free_space on btrfs partition


mbiebl:
Version: 3.0.2

Originally filed as downstream bug report https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1042001
What follows is a verbatim copy of  Andre Naujoks <nautsch2@gmail.com> 's bug report.

Dear Maintainer,

libblockdev seems to leak empty temporary directories, when the function bd_fs_get_free_space is called (even unsuccessfully) on a btrfs partition as it calls fs_mount internally, which then either fails and exhibits the behavior OR succeeds and btrfs_get_info leaks the directory.
I am relatively sure that this is not limited to btrfs, but that is what I have here.
I noticed this while steam was running and my /tmp directory was suddenly overflowing with hundreds of blockdev.XXXXXX directories. I don't know what steam does there or if it is in fact steam itself, proton or something else, but libblockdev should not leak the directories.

This little piece of code demonstrates the problem. It does not matter, if the device exists or not. It just cannot be mounted already.
Run it and an empty blockdev.XXXXXX dir remains in /tmp
```
#include <blockdev/blockdev.h>
#include <blockdev/fs/generic.h>
#include <stdio.h>

int main()
{
        GError *err = NULL;
        BDPluginSpec fs_plugin = {BD_PLUGIN_FS, NULL};
        BDPluginSpec *plugins[] = {&fs_plugin, NULL};
        if (!bd_init(plugins, NULL, &err)) {
                fprintf(stdout, "failed to init lib: %s\n", err->message);
        }
        err = NULL;
        guint64 space = bd_fs_get_free_space("/dev/sda1", "btrfs", &err);
        if (space == 0) {
                fprintf(stdout, "failed to get fs size: %s\n", err->message);
        } else {
        	fprintf(stdout, "%d\n", space);
	}

        return 0;
}
```
As far as I can see, the following patch should remedy at least the unsuccessul path, where the device cannot be mounted. However, I have not tested this!!
```
diff --git a/src/plugins/fs/generic.c b/src/plugins/fs/generic.c
index 21032b8a..843de8b1 100644
--- a/src/plugins/fs/generic.c
+++ b/src/plugins/fs/generic.c
@@ -650,6 +650,7 @@ static gchar* fs_mount (const gchar *device, gchar *fstype, gboolean read_only,
             if (!ret) {
[...]

PR #950, opened 2023-07-26
fs: Fix leaking directories with temporary mounts

vojtechtrefny:
Fixes: #949


src/plugins/fs/generic.c
```diff
@@ -715,7 +715,8 @@ static gboolean xfs_resize_device (const gchar *device, guint64 new_size, const
                    from the resize is more important so just ignore the
                    unmount error */
                 g_clear_error (&local_error);
-        }
+        } else
+            g_rmdir (mountpoint);
```
tbzatek, line 719: I think you still need to handle the above cases like ` if (!xfs_info) `  (I'm unable to add comment to this PR since that line wasn't changed)

vojtechtrefny: Thanks, I've just moved the `fs_mount` call after the `bd_fs_xfs_get_info` one -- we can get the information we need without mounting, so it will be easier.

tbzatek (approved):
Right, looks good now

PR #951, opened 2023-08-08
nvme: Use interim buffer for nvme_get_log_sanitize()

tbzatek:
Certain drives tend to return more data than expected resulting in stack corruption. Use an interim buffer large enough to handle the excess bytes.

https://github.com/storaged-project/udisks/issues/1152
https://github.com/linux-nvme/libnvme/issues/684

----

This is a dirty workaround until a proper fix is implemented either in libnvme or in the kernel.

### DIFF
--- a/src/plugins/nvme/nvme-info.c
+++ b/src/plugins/nvme/nvme-info.c
@@ -1026,7 +1026,8 @@ BDNVMESelfTestLog * bd_nvme_get_self_test_log (const gchar *device, GError **err
 BDNVMESanitizeLog * bd_nvme_get_sanitize_log (const gchar *device, GError **error) {
     int ret;
     int fd;
-    struct nvme_sanitize_log_page sanitize_log = ZERO_INIT;
+    char buf[65536] = ZERO_INIT;
+    struct nvme_sanitize_log_page *sanitize_log;
     BDNVMESanitizeLog *log;
     __u16 sstat;
 
@@ -1036,7 +1037,7 @@ BDNVMESanitizeLog * bd_nvme_get_sanitize_log (const gchar *device, GError **erro
         return NULL;
 
     /* send the NVME_LOG_LID_SANITIZE ioctl */
-    ret = nvme_get_log_sanitize (fd, FALSE /* rae */, &sanitize_log);
+    ret = nvme_get_log_sanitize (fd, FALSE /* rae */, (struct nvme_sanitize_log_page *) &buf);
     if (ret != 0) {
         _nvme_status_to_error (ret, FALSE, error);
         g_prefix_error (error, "NVMe Get Log Page - Sanitize Status Log command error: ");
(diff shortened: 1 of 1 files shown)


vojtechtrefny (approved):

igaw:
Makes sense until we figure out how to deal with such devices. I suspect we need to introduce quirks for this. BTW, would it possible to add some debug code which checks if the device writes too much data?

tbzatek:
> BTW, would it possible to add some debug code which checks if the device writes too much data?

We can afford having extra checks here, any suggestion how to achieve that other than analyzing data in the buffer itself?

igaw:
> Both record length 0x940 and 0x5c0 are divisible by 32.
> struct nvme_sanitize_log_page's actual data are 32 bytes

We are requesting 512 bytes and thus the device should only transfer 512 bytes. Older kernels did always use bounce buffers for this operation but with the fix for the dma alignment setting (https://github.com/torvalds/linux/commit/52fde2c07da606f3f120af4f734eadcfb52b04be) newer kernels are using zero copy I/O. 

So we just need one more experiment to have all statements supported by experiements.

Could you build a kernel and hack `blk_rq_map_user_iov` so that always using bounce buffers?

```patch
--- a/block/blk-map.c
+++ b/block/blk-map.c
@@ -627,7 +627,7 @@ int blk_rq_map_user_iov(struct request_queue *q, struct request *rq,
                        struct rq_map_data *map_data,
                        const struct iov_iter *iter, gfp_t gfp_mask)
 {
-       bool copy = false, map_bvec = false;
+       bool copy = true, map_bvec = false;
        unsigned long align = q->dma_pad_mask | queue_dma_alignment(q);
        struct bio *bio = NULL;
        struct iov_iter i;

```

igaw:
Okay, let me play with this. I though I got the right place to force the kernel to use bounce buffers. Obviously, it was the wrong place...

igaw:
@allfoxwy I was stupid, the correct test patch is here: https://github.com/storaged-project/udisks/issues/1152#issuecomment-1685977468

keithbusch:
Maybe nvmecli could just allocate a page aligned and sized buffer instead?

tbzatek:
Let's merge this at its current state as @vojtechtrefny is planning to make a new release tomorrow.

Feel free to continue here even when the ticket is closed. I'm watching the linux-nvme mailing list thread but no clear outcome as of yet.

igaw:
Sorry, got a bit drowned in work. Anyway, Christoph had and idea for another test patch. The idea here is to introduce a 512 byte alignment for the buffer and enable KASAN to see if this device will work then. From our previous tests and bissection, this has good chance to work. Anyway, will look into it... hopefully soon.

keithbusch:
Ok, if the device is overrunning an aligned buffer, I think we can conclude it's not a bad interpretation of a PRP with an offset. The device just always writes a full 4k page into a 512b log. That's not good: the kernel DMA alignment suggestion isn't viable and it's preferred that user space work around this with an over-allocated buffer instead of relying on the kernel to do something about it.

I think the reason you may have seen different overruns in previous experiments may be because the original buffer + extra unwanted data straddles different page offsets, causing the device to transfer the bulk of the corruption to PRP2, which the driver hadn't set and would transfer to nowhere. Different page offsets will cause different observable buffer overruns.

PR #952, opened 2023-08-10
Always use "--fs ignore" with lvresize

mvollmer:
Depending on how lvresize is compiled, it might default to "--fs checksize", which is not what we want for this low level API.

Also, "--fs checksize" does not always succeed even when the filesystem is the right size, such as when libblkid does not work for NTFS.

### DIFF
--- a/src/plugins/lvm-dbus.c
+++ b/src/plugins/lvm-dbus.c
@@ -2653,12 +2653,6 @@ gboolean bd_lvm_lvresize (const gchar *vg_name, const gchar *lv_name, guint64 si
     GVariant *params = NULL;
     GVariant *extra_params = NULL;
     gboolean success = FALSE;
-    BDLVMLVdata *lvinfo = NULL;
-
-    lvinfo = bd_lvm_lvinfo (vg_name, lv_name, error);
-    if (!lvinfo)
-        /* error is already populated */
-        return FALSE;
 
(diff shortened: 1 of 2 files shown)


StorageGhoul:

mvollmer:
The interaction of inactive logical volumes, "--force", and "--fs" is
interesting: With "--fs checksize" and without "--force", a
inactive volume leads to an immediate error.  With "--force", the volume is temporarily activated, the checksize is performed and then the volume is deactivated again.

(When the check fails, or crashes, the volume is left activated. Not sure if that counts as a bug.)

Thus, the bd_lvm_resize would only succeed when the filesystem is already the right size for the new size of the logical volume. Is that would you see?

Anyway, this PR should help Cockpit avoid this situation: https://github.com/lvmteam/lvm2/issues/123

mvollmer:
(I think `--force` should imply `--fs ignore`, but *shrug*.)

tbzatek:
> With "--force", the volume is temporarily activated, the checksize is performed and then the volume is deactivated again.

Does this result in uevents being registered by UDisks, potentially accessing the block devices and blocking deactivation for a brief period of time?

tbzatek (approved):
I'm generally okay with this but I leave it up to @vojtechtrefny for his final decision.

vojtechtrefny (approved):
Thank you. Looks good to me. I hope this is the last fix for this new `lvresize --fs` nonsense...

PR #954, opened 2023-08-29
crypto: Correctly convert passphrases from Python to C

vojtechtrefny:
We can't use ord to convert the Python string to the byte array expected, with ord we'd support only ASCII passphrases and we want to support all unicode characters.

### DIFF
--- a/tests/crypto_test.py
+++ b/tests/crypto_test.py
@@ -448,6 +448,43 @@ class CryptoTestOpenClose(CryptoTestCase):
     def test_luks2_open_close(self):
         self._luks_open_close(self._luks2_format)
 
+    @tag_test(TestTags.SLOW, TestTags.CORE)
+    def test_luks2_open_close_non_ascii_passphrase(self):
+        passphrase = "šššššššš"
+
+        self._luks2_format(self.loop_dev, passphrase)
+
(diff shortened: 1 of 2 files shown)


tbzatek (approved):
Yeah, makes sense to default to UTF-8. Gtk2+ is UTF-8 only and hope that everything modern Web-like too.

I know these are Python bindings but would be perhaps nice to document this, if there's a chance.

tests/crypto_test.py
```diff
@@ -448,6 +448,43 @@ def test_luks_open_close(self):
     def test_luks2_open_close(self):
         self._luks_open_close(self._luks2_format)
 
+    @tag_test(TestTags.SLOW, TestTags.CORE)
+    def test_luks2_open_close_non_ascii_passphrase(self):
+        passphrase = "šššššššš"
```
vpodzime, line 453:
```suggestion
        passphrase = "šššššššš-hů-hů"
```

vpodzime (approved):
ACK (with my critical fix, of course!)

tests/crypto_test.py
tbzatek: Good point! Go for it!

PR #955, opened 2023-09-06
spec: Obsolete vdo plugin packages

vojtechtrefny:
We stopped building the VDO plugin in Fedora 34 but never made it obsolete and it is now breaking upgrades.

Resolves: rhbz#2237477


tbzatek (approved):
yup

PR #956, opened 2023-09-08
spec: Move obsoleted devel subpackages to libblockdev-devel

vojtechtrefny:
Pointed out during testing of fix for rhbz#2237477. Not critical, but it makes more sense for the devel subpackage to obsolete the removed devel subpackages.


tbzatek (approved):
Ah yea, makes sense

PR #958, opened 2023-09-25
Prefer using g_autofree

jelly:
GLib provides nice automatic cleanup macros https://docs.gtk.org/glib/auto-cleanup.html

I'd be happy to convert more of the codebase to it, if this is a desired change.

### DIFF
--- a/src/plugins/check_deps.c
+++ b/src/plugins/check_deps.c
@@ -250,13 +250,13 @@ check_dbus_deps (volatile guint *avail_deps, guint req_deps, const DBusDep *buse
 }
 
 static gboolean _check_util_feature (const gchar *util, const gchar *feature, const gchar *feature_arg, const gchar *feature_regexp, GError **error) {
-    gchar *util_path = NULL;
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
> I'm not a fan of this construct, but it's a purely personal opinion.

You should've said something sooner :-)

```
$ git grep g_autofree | wc -l
58
```

tbzatek (approved):
> You should've said something sooner :-)

I know, I'm being tolerant.

The thing is, while it prevent some leaks, it brings additional confusion when analyzing memory ownership by hand (as we often do) and brings new ownership questions e.g. when the pointer is pulled out of the (current function) frame.

So I'm not completely against, it makes sense when writing a new code where you're fully conscious of what chunks of memory you're sharing further, but I don't see a benefit of rewriting existing code. On the contrary, it may be prone to additional errors. We've been burned by unforeseen consequences in UDisks and still fixing unexpected race conditions, loose or even missing object references and memory ownership issues in general.

vojtechtrefny (approved):

PR #959, opened 2023-09-25
fs: basic bcachefs support

jelly:
I was interested in learning a bit more about libblockdev/udisks so I've looked into adding bcachefs support to libblockdev (although it is not merged yet in the kernel). So far it only supports mkfs/get_info, some notable things which are missing are:

* [ ] tests
* [ ] get_fs_size support and similar functions.
* [ ] set/get label
* [ ] repair
* [ ] resize
* [ ] set uuid (if possible)

Like btrfs this likely needs a specialized plugin for volume's, adding two devices together and other interesting features. I don't for see myself writing support for that. But I would be happy to implement enough to get basic support on the same level as other filesystems.

FYI: I can understand that the lack of bcachefs support in mainline and userspace tooling in CI blocks this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Following storage technologies are supported by libblockdev
  - partitions
    - MSDOS, GPT
  - filesystem operations
-   - ext2, ext3, ext4, xfs, vfat, ntfs, exfat, btrfs, f2fs, nilfs2, udf
+   - ext2, ext3, ext4, xfs, vfat, ntfs, exfat, btrfs, f2fs, nilfs2, udf, bcachefs
    - mounting
  - LVM
    - thin provisioning, LVM RAID, cache, LVM VDO
--- a/misc/install-test-dependencies.yml
+++ b/misc/install-test-dependencies.yml
@@ -69,6 +69,10 @@
       - xfsprogs
     when: ansible_distribution == 'Fedora' and test_dependencies|bool
 
+  - name: Install bcachefs-tools test dependency (Fedora)
+    package: name=bcachefs-tools state=present
+    when: ansible_distribution == 'Fedora' and ansible_distribution_version == '40' and test_dependencies|bool
+
 ####### CentOS 8
   - name: Install basic build tools (CentOS 8)
     package: name={{item}} state=present
--- a/src/lib/plugin_apis/fs.api
+++ b/src/lib/plugin_apis/fs.api
@@ -527,6 +527,7 @@ typedef enum {
     BD_FS_TECH_EXFAT,
     BD_FS_TECH_BTRFS,
     BD_FS_TECH_UDF,
+    BD_FS_TECH_BCACHEFS,
 } BDFSTech;
 
 typedef enum {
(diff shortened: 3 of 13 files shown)


StorageGhoul:

vojtechtrefny:
> Like btrfs this likely needs a specialized plugin for volume's, adding two devices together and other interesting features. I don't for see myself writing support for that. But I would be happy to implement enough to get basic support on the same level as other filesystems.

Important question for this is whether people will want to use bcachefs as "simple" filesystem. We originally didn't have support for btrfs as a simple single device filesystem, just as a multi-device volume manager and we added it as a filesystem because people were asking for it. I am not sure if this will be the same for bcachefs as well. I am not against adding it, but I'd like to avoid supporting something for years that might not be used as much. But it looks simple enough for now.

> I can understand that the lack of bcachefs support in mainline and userspace tooling in CI blocks this PR.

I guess we could realistically expect that bcachefs will be available in next kernel release, right? So it might be available in Fedora 40. My only concern right now is stability of the userspace tools "API", we can expect changes in the output and that's always tricky to support.

jelly:
> > Like btrfs this likely needs a specialized plugin for volume's, adding two devices together and other interesting features. I don't for see myself writing support for that. But I would be happy to implement enough to get basic support on the same level as other filesystems.
> 
> Important question for this is whether people will want to use bcachefs as "simple" filesystem. We originally didn't have support for btrfs as a simple single device filesystem, just as a multi-device volume manager and we added it as a filesystem because people were asking for it. I am not sure if this will be the same for bcachefs as well. I am not against adding it, but I'd like to avoid supporting something for years that might not be used as much. But it looks simple enough for now.

Half supporting something might be worse then lacking support. As I most likely won't be pushing this a lot forward, mostly did it as a learning exercise. So I'll close it for now, anyone who is deeply invested in bcachefs can pick up this PR and push it to the finishing line.

> > I can understand that the lack of bcachefs support in mainline and userspace tooling in CI blocks this PR.
> 
> I guess we could realistically expect that bcachefs will be available in next kernel release, right? So it might be available in Fedora 40. My only concern right now is stability of the userspace tools "API", we can expect changes in the output and that's always tricky to support.

It's now in linux-next, I have no idea when it ends up in mainline. It also depends on Fedora packaging bcachefs-tools which has some Rust so that might not be super trivial.

tbzatek:
Another user request: https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/issues/344

tbzatek:
...and now it's gone from kernel again 😆

jelly:
Let's close it then

hgkamath:
- just a user here
- Excerpts from my comment in https://github.com/storaged-project/blivet/issues/1238 
  - > perhaps going DKMS is the preferred way to get FS-patches to users as quickly as possible, rather than waiting 2 kernel release cycles for the fixes that affect the filesystem that manage a user's precious data. 
  - > File-systems naturally compete with one-another, naturally lend themselves for comparisons. Each one would like to stymie adoption of others. The old timers would like a bueracracy that makes it harder for new-entrants. imho, For its age, bcachefs, in stability, does seem ahead of where other FS were for the same age. **Competition between different FS-types is essential for the data ecosystem**, so that no-FS-type-developer-group gets complacent. The more the supported FS, the more people who try things out, more user experiences and reports on whether a FS-type is reliable or not. Storage utilities play a vital important role in keeping that competition alive and healthy.
- If there has been volunteer contributions to make bcache support happen, partial though they may be, waiting for other fs-tools to become available, whether the fs-code is in-tree or in DKMS does not and should not matter for storage utilities. I would hate to see volunteer code lost for that reason. 
- Even if one is a user of a different FS-type, one needs bcachefs to keep one's fs-type developers on their toes so that quality improving improves. 
- As users who store data, we all need reliable filesystems

PR #965, opened 2023-10-03
fs: hide the progressbar when resizing ntfs

### LINKED ISSUE #963: ntfsresize disable progress bar
While testing filesystem resizing in Cockpit I noticed by default `ntfsresize` enables the progressbar, that leads to some noisy output. Now in Cockpit we should circumvent that people run into this scenario in the first place, however does it not make sense for libblockdev to use `--no-progess-bar` by default? The option has been there [for over 20 years](https://github.com/tuxera/ntfs-3g/commit/c90976e02301612dadd149b390fc2e5ca7461da2#diff-2c62a2bdf78f24c7e2d58d6dc3589852436d9a9124136db9ed9d06e8a23ddd10R249). And I don't see any progress reporting support for resize actions.

image

jelly:
When resizing fails the error output contains unneeded progress information.

Fixes #963


StorageGhoul:

tbzatek:
Now this is byte-to-byte equal with #964 ;-)

vojtechtrefny:
Sorry, I was faster :-)

Issue #944, opened 2023-07-23
Assertion error in tests/library_test.py


mbiebl:
Version: 3.0.2-1

I'm currently trying to run the test suite but I run into the following failures:
```
======================================================================
FAIL: test_force_plugin (library_test.PluginsTestCase.test_force_plugin)
Verify that forcing plugin to be used works as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/autopkgtest.ZLSthX/tree/tests/library_test.py", line 240, in test_force_plugin
    self.assertEqual(BlockDev.lvm_get_max_lv_size(), 1024)
AssertionError: 9223372036854775808 != 1024

======================================================================
FAIL: test_plugin_priority (library_test.PluginsTestCase.test_plugin_priority)
Verify that preferring plugin to be used works as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/autopkgtest.ZLSthX/tree/tests/library_test.py", line 283, in test_plugin_priority
    self.assertEqual(BlockDev.lvm_get_max_lv_size(), 1024)
AssertionError: 9223372036854775808 != 1024

======================================================================
FAIL: test_reload (library_test.PluginsTestCase.test_reload)
Verify that reloading plugins works as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/autopkgtest.ZLSthX/tree/tests/library_test.py", line 198, in test_reload
    self.assertEqual(BlockDev.lvm_get_max_lv_size(), 1024)
AssertionError: 9223372036854775808 != 1024

----------------------------------------------------------------------
Ran 487 tests in 2465.088s

FAILED (failures=3, errors=2, skipped=21)
```

From a cursory glance at tests/library_test.py, the problem might be the following code:
```
        os.system ("rm -f src/plugins/.libs/libbd_lvm2.so")
```

I see several issues:
- The test seems to try to compile the library on the fly and does not use the system provided libblockdev-lvm3
-  The library name appears to be incorrect, shouldn't it use `libbd_lvm.so.3`

mbiebl:
nvm: Just saw
```
        os.system("cp src/plugins/.libs/libbd_lvm.so src/plugins/.libs/libbd_lvm2.so")
```

What remains, is the need to need to compile those plugins on the fly, ie. I will probably need to update my test bed to include all build dependencies.

vojtechtrefny:
[...]

PR #971, opened 2023-10-11
tests: Fail early when recompilation fails in library_test

vojtechtrefny:
The tests in PluginsTestCase recompile the library and if this step fails the test case can fail with an unexpected error that is hard to debug so we should just fail early with some reasonable message instead.

Fixes: #944


tbzatek (approved):
Huh

PR #973, opened 2023-10-11
Add BDPluginSpec constructor and use it in plugin_specs_from_names

vojtechtrefny:
Starting with Python 3.12 we got some weird complains when setting attributes of BlockDev.PluginSpec in plugin_specs_from_names, it's better to add a fully working constructor to create the struct correctly from the Python overrides.

### DIFF
--- a/src/lib/plugins.c
+++ b/src/lib/plugins.c
@@ -40,6 +40,24 @@ void bd_plugin_spec_free (BDPluginSpec *spec) {
     g_free (spec);
 }
 
+/**
+ * bd_plugin_spec_new: (constructor)
+ * @name: %BDPlugin name, e.g. %BD_PLUGIN_LVM
+ * @so_name: (nullable): SO name of the plugin to load or %NULL for default
+ *
+ * Returns: (transfer full): a new plugin spec
+ */
+BDPluginSpec* bd_plugin_spec_new (BDPlugin name, const gchar *so_name) {
+    BDPluginSpec* ret = g_new0 (BDPluginSpec, 1);
+    ret->name = name;
+    /* FIXME: this should be const, but we are already allocating a new string in _copy
+    *         and freeing it in _free
+    */
+    ret->so_name = so_name ? g_strdup (so_name) : NULL;
+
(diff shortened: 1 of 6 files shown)


src/lib/plugins.c
```diff
@@ -40,6 +40,21 @@ void bd_plugin_spec_free (BDPluginSpec *spec) {
...
+ * bd_plugin_spec_new: (constructor)
+ * @name: %BDPlugin name, e.g. %BD_PLUGIN_LVM
+ * @so_name: (nullable): SO name of the plugin to load or %NULL for default
+ *
+ * Returns: (transfer full): a new plugin spec
+ */
+BDPluginSpec* bd_plugin_spec_new (BDPlugin name, const gchar *so_name) {
+    BDPluginSpec* ret = g_new0 (BDPluginSpec, 1);
+    ret->name = name;
+    ret->so_name = so_name ? g_strdup (so_name) : NULL;
```
tbzatek, line 53: allocating a const? typecasting it for `g_free()`?

src/lib/plugins.c
```diff
@@ -40,6 +40,21 @@ void bd_plugin_spec_free (BDPluginSpec *spec) {
     g_free (spec);
 }
 
+/**
+ * bd_plugin_spec_new: (constructor)
+ * @name: %BDPlugin name, e.g. %BD_PLUGIN_LVM
+ * @so_name: (nullable): SO name of the plugin to load or %NULL for default
```
tbzatek, line 46: `BDPlugin` is `enum`, i.e. `int`.

src/lib/plugins.c
vojtechtrefny: I know it's weird, but we already allocate it in `bd_plugin_spec_copy` and free it in `bd_plugin_spec_free` and I didn't want to change it.

tbzatek: Hmm fair enough. Perhaps you could add some `/* FIXME */`s? :-)

tbzatek (approved):

PR #974, opened 2023-10-12
tests: Fix "invalid escape sequence '\#'" warning from Python 3.12

vojtechtrefny:
These escapes are not meant for Python so we just need to mark the string as regular expresion so Python ignores escape characters in it.


tbzatek (approved):
Clever! I think I removed all the bad characters in udisks.

PR #981, opened 2023-10-26
fs: Fix allowed UUID for generic mkfs with VFAT

vojtechtrefny:
We allow using UUID with dash in the middle for VFAT in both check and set_uuid so we should allow it also with the generic mkfs calls.


tbzatek (approved):
Nice and clean.

PR #982, opened 2023-11-02
fs: Add support for getting filesystem min size for NTFS and Ext

vojtechtrefny:
For NTFS and Ext we can get a filesystem minimum size estimate using the resize tools. Blivet is currently using this to get filesystem min size for resize and the resize tools return slightly different number than what we get from the dump/info tools.


vojtechtrefny:
Jenkins, test this please.

tbzatek (approved):
Looks good!

PR #987, opened 2023-11-08
tests: Use BDPluginSpec constructor in LVM DBus plugin tests

vojtechtrefny:
Follow up for #973, I forgot to fix the BDPluginSpec construction for LVM DBus test cases.


tbzatek (approved):
Yup, warnings are gone now.

PR #989, opened 2023-11-13
python: Add a deepcopy function to our structs

vojtechtrefny:
Python's deepcopy by default tries to use pickle which doesn't work on our structs, where possible we should use our C copy constructor.
The deepcopy functionality is used by Anaconda to copy Blivet devicetree which now includes filesystem information provided by libblockdev.

Related: https://github.com/rhinstaller/kickstart-tests/issues/1025


tbzatek (approved):
That's a lot of faith!

PR #990, opened 2023-11-20
lvm-dbus: Minor memory allocation fixes

tbzatek:
https://bugzilla.redhat.com/show_bug.cgi?id=2247319


tbzatek:
Let's see what adding the `G_GNUC_PRINTF (2, 3)` breaks. The argument symbol definition is unchanged and adding the annotation won't be hopefully classified as API break.

vojtechtrefny (approved):

Issue #993, opened 2024-01-09
test suite failure: FAIL: test_create_delete_subvolume (btrfs_test.BtrfsTestCreateDeleteSubvolume.test_create_delete_subvolume)


mbiebl:
Version: 3.0.4, master

running the upstream test `tests/run_tests.py` on Debian sid/trixie results in

```
1778s Test Linux kernel version detection ... ok
1778s» 
1778s ======================================================================
1778s FAIL: test_create_delete_subvolume (btrfs_test.BtrfsTestCreateDeleteSubvolume.test_create_delete_subvolume)
1778s Verify that it is possible to create/delete subvolume
1778s ----------------------------------------------------------------------
1778s Traceback (most recent call last): 
1778s   File "/tmp/autopkgtest.bwbbgi/build.3wN/src/tests/btrfs_test.py", line 212, in test_create_delete_subvolume
1778s     with self.assertRaises(GLib.GError):
1778s AssertionError: Error not raised
1778s» 
1778s ----------------------------------------------------------------------
1778s Ran 486 tests in 1700.229s
1778s»
1778s FAILED (failures=1, skipped=21) 
1778s autopkgtest [20:36:26]: test upstream: -----------------------]
1779s autopkgtest [20:36:27]: test upstream:  - - - - - - - - - - results - - - - - - - - - -
1779s upstream             FAIL non-zero exit status 1
1780s autopkgtest [20:36:28]: @@@@@@@@@@@@@@@@@@@@ summary 
1780s upstream             FAIL non-zero exit status 1 
1780s qemu-system-x86_64: terminating on signal 15 from pid 90336 (/usr/bin/python3)
```

vojtechtrefny:
This is regression in btrfs-progs 6.6.3: https://github.com/kdave/btrfs-progs/issues/730

mbiebl:
Thanks for the reference, closing this issue accordingly

PR #994, opened 2024-01-10
tests: Skip some checks for btrfs errors with btrfs-progs 6.6.3

vojtechtrefny:
Latest btrfs-progs return 0 for some subvolume actions that return an error, we'll skip these checks on 6.6.3 for now.

Fixes: #993


tbzatek (approved):
Nasty.

PR #998, opened 2024-01-20
Move UNUSED definition to common header utils.h

giuliobenetti:
Let's move UNUSED defition to avoid redundancy.

### DIFF
--- a/src/plugins/btrfs.c
+++ b/src/plugins/btrfs.c
@@ -154,7 +154,6 @@ void bd_btrfs_close (void) {
 }
 
 
-#define UNUSED __attribute__((unused))
 
 /**
  * bd_btrfs_is_tech_avail:
@@ -165,7 +164,7 @@ void bd_btrfs_close (void) {
  * Returns: whether the @tech-@mode combination is available -- supported by the
  *          plugin implementation and having all the runtime dependencies available
  */
-gboolean bd_btrfs_is_tech_avail (BDBtrfsTech tech UNUSED, guint64 mode UNUSED, GError **error) {
+gboolean bd_btrfs_is_tech_avail (BDBtrfsTech tech G_GNUC_UNUSED, guint64 mode G_GNUC_UNUSED, GError **error) {
     /* all tech-mode combinations are supported by this implementation of the plugin */
     return check_deps (&avail_deps, DEPS_BTRFS_MASK, deps, DEPS_LAST, &deps_check_lock, error) &&
            check_module_deps (&avail_module_deps, MODULE_DEPS_BTRFS_MASK, module_deps, MODULE_DEPS_LAST, &deps_check_lock, error);
--- a/src/plugins/crypto.c
+++ b/src/plugins/crypto.c
@@ -60,7 +60,6 @@
 #define DEFAULT_LUKS2_SECTOR_SIZE 512
 #endif
 
-#define UNUSED __attribute__((unused))
 
 /**
  * SECTION: crypto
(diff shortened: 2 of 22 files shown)


StorageGhoul:

vojtechtrefny (commented):
I understand the simplification, but I am not a huge fan of adding a unprefixed and internal definition to a public header file. Thoughts @tbzatek?

vojtechtrefny:
Jenkins, ok to test.

giuliobenetti:
> This should all be changed to `G_GNUC_UNUSED` instead.

This is only defined for Glibc AFAIK, so for uclibc and musl we need to define it if not defined.
At this point we still need an internal header where redefine it if not defined.
If you point me the right internal header I can do that and include header where current UNUSED is used.

tbzatek:
> This is only defined for Glibc AFAIK, so for uclibc and musl we need to define it if not defined.

It's a compiler attribute, it should be libc-implementation independent. Besides, if there's a problem with the macro, it should be reported and fixed on glib side. We use glib quite extensively in libblockdev. Another aspect is that this annotation is not used in public headers, only in the actual code.

giuliobenetti:
> > This is only defined for Glibc AFAIK, so for uclibc and musl we need to define it if not defined.
> 
> It's a compiler attribute, it should be libc-implementation independent. Besides, if there's a problem with the macro, it should be reported and fixed on glib side. We use glib quite extensively in libblockdev. Another aspect is that this annotation is not used in public headers, only in the actual code.

AFAICS it is a glibc macro found in "#include <glib.h>"
https://developer-old.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-GNUC-UNUSED:CAPS

So it's not suitable for other libc I deal in Buildroot for libblockdev. I can add a patch local to Buildroot but we prefer to upstream everything possibly. So it would be great if we find a way to have this upstreamed.
If you point me a private header where to place the definition like:
#define LIBBLOCKDEV_UNUSED __attribute__((unused))
I modify the patch.

giuliobenetti:
> > > This is only defined for Glibc AFAIK, so for uclibc and musl we need to define it if not defined.
> > 
> > 
> > It's a compiler attribute, it should be libc-implementation independent. Besides, if there's a problem with the macro, it should be reported and fixed on glib side. We use glib quite extensively in libblockdev. Another aspect is that this annotation is not used in public headers, only in the actual code.
> 
> AFAICS it is a glibc macro found in "#include <glib.h>" https://developer-old.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-GNUC-UNUSED:CAPS
> 
> So it's not suitable for other libc I deal in Buildroot for libblockdev. I can add a patch local to Buildroot but we prefer to upstream everything possibly. So it would be great if we find a way to have this upstreamed. If you point me a private header where to place the definition like:
> 
> ```
> #define LIBBLOCKDEV_UNUSED __attribute__((unused))
> ```
> 
> I modify the patch.

Pardon, I've mixed up glib and glibc. I'm testing right now. @tbzatek you're right, sorry for the noise!
Once checked I come up with a patch using G_GNUC_UNUSED instead of UNUSED.

giuliobenetti:
> G_GNUC_UNUSED

I'm producing the patch right now and then I launch Buildroot autobuilder with all possible combinations, 72 arch/gcc/libc combinations. Once I have a working patch I push force so also your CI can test it :-)

giuliobenetti:
Ok, updated the patch and launched Buildroot autobuilder(45 cases, not 74) and your CI is working, let's see the result.
I've successfully built for Arm+glibc at the moment.

giuliobenetti:
Oops, I've dropped some UNUSED. Going to fix now

vojtechtrefny (approved):
Looks good to me now, thank you.

vojtechtrefny:
You can ignore the Debian testing CI failures, looks like pylint/astroid is broken after latest update, it's not related to this change.

giuliobenetti:
> You can ignore the Debian testing CI failures, looks like pylint/astroid is broken after latest update, it's not related to this change.

Perfect. Indeed I was about to ask you to relaunch it because of cached sources resulting in build failure.
But we're fine now then. Also Buildroot autobuilder is running successfully now.

tbzatek (approved):
Looks good, thanks!

giuliobenetti:
> Looks good, thanks!

Happy to contribute!

PR #999, opened 2024-01-22
RFC: btrfs plugin: use libbtrfsutil

jelly:
This is a quick hack to start off the discussion if libbtrfsutil would be a good fit for extending further features in the btrfs plugin for libblockdev. ( I recall there was a discussion somewhere, but couldn't find the issue)

Some notable things which needs to be verified:
* [ ] License, libbtrfsutil itself is licensed LGPL, is it compatible with libblockdev?
* [ ] Feature completeness, can everything currently be replaced with libbtrfsutil and what is missing, can it be supported in libbtrfsutil and is upstream willing too?
* [ ] Is libbtrfsutil packaged everywhere libblockdev is at the moment used.


StorageGhoul:

vojtechtrefny:
> This is a quick hack to start off the discussion if libbtrfsutil would be a good fit for extending further features in the btrfs plugin for libblockdev.

Definitely something we'd like to do. Every command call we can replace with a library call is a huge improvement.

> ( I recall there was a discussion somewhere, but couldn't find the issue)

That would be https://github.com/storaged-project/libblockdev/issues/552

> 
> Some notable things which needs to be verified:
> 
>     * [ ]  License, libbtrfsutil itself is licensed LGPL, is it compatible with libblockdev?

Yes, libblockdev is also licensed under LGPL 2.1 so there shouldn't be a problem.

> 
>     * [ ]  Feature completeness, can everything currently be replaced with libbtrfsutil and what is missing, can it be supported in libbtrfsutil and is upstream willing too?

It would be great if we could replace everything with libbtrfsutil, but I am ok with replacing only parts of the command calls with the library.

> 
>     * [ ]  Is libbtrfsutil packaged everywhere libblockdev is at the moment used.
> 

Libblockdev is packaged pretty much everywhere these days, but libbtrfsutil seems to be available in both Fedora and Debian/Ubuntu and that's good enough at least for our CI :)

configure.ac
```diff
@@ -229,6 +229,10 @@ AS_IF([test "x$with_fs" != "xno" -o "x$with_crypto" != "xno" -o "x$with_swap" !=
              [Define as neutral value if libblkid doesn't provide the definition])])]
       [])
 
+AS_IF([test "x$with_btrfs" != "xno"],
+      [LIBBLOCKDEV_PKG_CHECK_MODULES([LIBBTRFSUTIL], [libbtrfsutil >= 6.6])],
+      [])
```
vojtechtrefny, line 234: To unblock some of the CI, you'll need to `libbtrfsutil` dependency to few places:
- `dist/libblockdev.spec.in`
- `misc/install-test-dependencies.yml`
- `.packit.yaml`

This will cover Packit and the GH actions checks (we'll also need to add it to our internal CI, but that's something we need to do manually).

jelly:
> > This is a quick hack to start off the discussion if libbtrfsutil would be a good fit for extending further features in the btrfs plugin for libblockdev.
> 
> Definitely something we'd like to do. Every command call we can replace with a library call is a huge improvement.
> 
> > ( I recall there was a discussion somewhere, but couldn't find the issue)
> 
> That would be #552
> 
> > Some notable things which needs to be verified:
> > ```
> > * [ ]  License, libbtrfsutil itself is licensed LGPL, is it compatible with libblockdev?
> > ```
> 
> Yes, libblockdev is also licensed under LGPL 2.1 so there shouldn't be a problem.

Nice!

> > ```
> > * [ ]  Feature completeness, can everything currently be replaced with libbtrfsutil and what is missing, can it be supported in libbtrfsutil and is upstream willing too?
> > ```
> 
> It would be great if we could replace everything with libbtrfsutil, but I am ok with replacing only parts of the command calls with the library.

So from the LGPL/GPL note in the documentation, I can make up that it is basically everything `mkfs`, `repair`  and possibly `btrfs device add`.

> > ```
> > * [ ]  Is libbtrfsutil packaged everywhere libblockdev is at the moment used.
> > ```
> 
> Libblockdev is packaged pretty much everywhere these days, but libbtrfsutil seems to be available in both Fedora and Debian/Ubuntu and that's good enough at least for our CI :)

It's also available in Arch Linux. I should add the required CI changes, for Fedora this would mean a dependency on `libbtrfsutil` which `btrfs-progs` does not depend on (due to the licensing afaik).

src/plugins/btrfs.c
```diff
@@ -454,47 +455,12 @@ gboolean bd_btrfs_delete_subvolume (const gchar *mountpoint, const gchar *name,
...
-    match = g_match_info_fetch (match_info, 1);
-    ret = g_ascii_strtoull (match, NULL, 0);
 
-    g_free (match);
-    g_match_info_free (match_info);
-    g_regex_unref (regex);
-    g_free (output);
+    err = btrfs_util_get_default_subvolume(mountpoint, &ret);
+    if (err)
+        g_set_error (error, BD_BTRFS_ERROR, BD_BTRFS_ERROR_PARSE, btrfs_util_strerror(err));
```
tbzatek, line 463: I think we can afford changing error codes a little bit, if needed. While it would be nice to keep most of them consistent, I imagine the code path would be so different anyway and it would be good to distinguish between minor errors (i.e. "device open error", "parse error", "invalid argument", etc.)

tbzatek (commented):
Very nice!

src/plugins/btrfs.c
jelly: Agreed, currently the error is rather strange when given a path which is not mounted. So this needs more work.
```
AssertionError: ".*(can't|cannot) access.*" does not match "g-bd-btrfs-error-quark: Could not open (2)" 
```

jelly: It's a C style API so the test of the error was in errno.

configure.ac
```diff
@@ -229,6 +229,10 @@ AS_IF([test "x$with_fs" != "xno" -o "x$with_crypto" != "xno" -o "x$with_swap" !=
              [Define as neutral value if libblkid doesn't provide the definition])])]
       [])
 
+AS_IF([test "x$with_btrfs" != "xno"],
+      [LIBBLOCKDEV_PKG_CHECK_MODULES([LIBBTRFSUTIL], [libbtrfsutil >= 5.1])],
```
jelly, line 233: This is a bit arbitrary, needs some investigation which version is really needed.

src/plugins/btrfs.c
```diff
@@ -454,75 +460,37 @@ gboolean bd_btrfs_delete_subvolume (const gchar *mountpoint, const gchar *name,
...
- *                                                 passed to the 'btrfs' utility)
  * @error: (out) (optional): place to store error (if any)
  *
  * Returns: whether the @mountpoint volume's default subvolume was correctly set
  * to @subvol_id or not
  *
  * Tech category: %BD_BTRFS_TECH_SUBVOL-%BD_BTRFS_TECH_MODE_MODIFY
  */
-gboolean bd_btrfs_set_default_subvolume (const gchar *mountpoint, guint64 subvol_id, const BDExtraArg **extra, GError **error) {
-    const gchar *argv[6] = {"btrfs", "subvol", "set-default", NULL, mountpoint, NULL};
```
jelly, line 516: This breaks [udisks](https://github.com/storaged-project/libblockdev/actions/runs/7616878904/job/20744641672?pr=999), which I think we rather not? But -Werror will scream at me if I keep it around. I guess I should mark it with something __unused?

src/plugins/btrfs.c
tbzatek: Different error messages are fine, there's generally no guarantee of stability and clients should never match any strings (it's okay for tests where it is intentional) - that's what the error codes are for.

src/plugins/btrfs.c
tbzatek: Yes, existing public API cannot be changed. Please add `G_GNUC_UNUSED` annotation.

If you want to keep full API compatibility, you'd need to translate any known extra args to library calls/arguments. Porting existing API is tough, creating new API from scratch is much easier.

jelly: I think deprecating the argument would be the best, I am not sure what extra options can be given as `btrfs subvolume set default` only takes:

```
btrfs subvolume set-default <subvolid> <path>
```

So I think for backwards compatibility this would actually work out quite well in this instance. Should I add `G_DEPRECATED` as well to the argument and update the comment?

tbzatek: No need to special annotation for function arguments I guess.

jelly:
I've rebased and ported all functions which can use libbtrfsutil. I haven't managed to land a [new feature yet](https://github.com/kdave/btrfs-progs/pull/736) in libbtrfsutil. But I don't think btrfs-progs wants to block new additions to the API.

This is all still a bit WiP, there are some TODO's regarding error checking and one test which cannot be ported as it overrides `btrfs subvolume list` (The weird docker subvolumes one). But in all it seems to shave off ~ 100 loc.

PR #1001, opened 2024-01-25
Speed up btrfs tests

jelly:
Re-structure the test classes in the btrfs tests to not always create two lio devices in total this saves ~ 30 seconds on my T14s running the tests in a Fedora VM with 4 cores.

Before:

Ran 24 tests in 99.031s

Ater:

Ran 24 tests in 66.436s

Benchmarks aren't the most accurate but should give a rough estimate.

### DIFF
--- a/tests/btrfs_test.py
+++ b/tests/btrfs_test.py
@@ -21,7 +21,6 @@ TEST_MNT = "/tmp/libblockdev_test_mnt"
 def wipefs(device):
     os.system("wipefs -a %s > /dev/null" % device)
 
-
 class BtrfsTest(unittest.TestCase):
     requested_plugins = BlockDev.plugin_specs_from_names(("btrfs",))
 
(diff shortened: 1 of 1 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

tests/btrfs_test.py
```diff
@@ -46,8 +46,8 @@ def test_plugin_version(self):
 
 class BtrfsTestCase(BtrfsTest):
 
-    def setUp(self):
-        self.dev_file = create_sparse_tempfile("btrfs_test", 1024**3)
+    def setUp(self, dev_file_size=1024**3):
```
vojtechtrefny, line 49: Please move this to a new separate function called from `setUp`, `setUp` is inherited from `unittest.TestCase` and I don't like overriding the parameters, we could easily overlook this when creating a new subclass of `BtrfsTestCase` and break things by having a custom `setUp`.

vojtechtrefny (changes requested):
Looks good to me otherwise, thanks, every test speed up is great improvement. Now we need to do this for LVM and other test cases too :-)

vojtechtrefny:
You can ignore the Debian testing failures, we are waiting for new pylint to make it to testing: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1060894

tests/btrfs_test.py
vojtechtrefny: Thanks, this is exactly what I had in mind.

I just don't like adding a new parameters to `setUp`, because it doesn't have any parameters in `unittest.TestCase` and it can cause confusion when adding new test cases in the future.

vojtechtrefny (approved):
Thank you, looks good to me now.

PR #1005, opened 2024-02-24
crypto: Add support for conversion between different LUKS formats

xyakimo1:


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

src/plugins/crypto.c
```diff
@@ -2108,6 +2108,68 @@ gboolean bd_crypto_luks_set_uuid (const gchar *device, const gchar *uuid, GError
...
+                     "Failed to load device: %s", strerror_l (-ret, c_locale));
+        crypt_free (cd);
+        return FALSE;
+    }
+
+    const char *cd_type = crypt_get_type(cd);
+    if (g_strcmp0(cd_type, CRYPT_LUKS1) != 0 && \
+        g_strcmp0 (cd_type, CRYPT_LUKS2) != 0) {
+
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_TECH_UNAVAIL,
```
vojtechtrefny, line 2147: Please use `BD_LVM_ERROR_FAIL` here (and in the next error msg too), `BD_LVM_ERROR_TECH_UNAVAIL` is used only in specific cases where support for something is not available due to missing or old dependencies (e.g. cryptsetup version being too old, in this case the operation is simply not possible, not unsupported).

vojtechtrefny (changes requested):
Looks good to me otherwise, thank you. Also please squash the commits before we'll merge this, thanks.

src/lib/plugin_apis/crypto.api
```diff
@@ -1076,6 +1076,21 @@ gboolean bd_crypto_luks_set_label (const gchar *device, const gchar *label, cons
  */
 gboolean bd_crypto_luks_set_uuid (const gchar *device, const gchar *uuid, GError **error);
 
+/**
+ * bd_crypto_luks_convert:
+ * @device:  a LUKS device to convert to a different version of LUKS
+ * @convert_to_luks_version: the LUKS version to convert to
```
tbzatek, line 1082: I'd suggest to use a shorter naming, i.e. `new_version` or `target_version`.

src/plugins/crypto.c
```diff
@@ -2108,6 +2108,68 @@ gboolean bd_crypto_luks_set_uuid (const gchar *device, const gchar *uuid, GError
...
+
+    ret = crypt_load (cd, CRYPT_LUKS, NULL);
+    if (ret != 0) {
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_DEVICE,
+                     "Failed to load device: %s", strerror_l (-ret, c_locale));
+        crypt_free (cd);
+        return FALSE;
+    }
+
+    const char *cd_type = crypt_get_type(cd);
```
tbzatek, line 2143: Please move all declarations at the beginning of the code block. Also, please add a space between opening parenthesis.

src/plugins/crypto.c
```diff
@@ -2108,6 +2108,68 @@ gboolean bd_crypto_luks_set_uuid (const gchar *device, const gchar *uuid, GError
...
+    if (g_strcmp0(cd_type, CRYPT_LUKS1) != 0 && \
```
tbzatek, line 2144: No need for the backslash.

tbzatek (commented):
Just a couple of codestyle things.

xyakimo1:
@tbzatek, I've fixed codestyle, thank you for your notices.

src/plugins/crypto.c
xyakimo1: I've changed the error to `BD_LVM_ERROR_FAIL` as you asked, but it seems to be a bit off to me -- why LVM if we are in the crypto plugin?

I didn't find explanations to different types of errors; please point me to them if they exist, and I will take them into account next time.

Also, did I specify the tech mode correctly (`BD_CRYPTO_TECH_MODE_MODIFY`)?

vojtechtrefny: > I've changed the error to BD_LVM_ERROR_FAIL as you asked, but it seems to be a bit off to me -- why LVM if we are in the crypto plugin?

Yes, using LVM error code is completely wrong here, I have no idea why I recommended this one :man_facepalming: Looking at the error codes we have in the crypto plugin, it looks like we don't have anything that would fit here so please add a new error code, probably something like `BD_CRYPTO_ERROR_CONVERT_FAILED`.

> Also, did I specify the tech mode correctly (BD_CRYPTO_TECH_MODE_MODIFY)?

Yes, that's correct.

xyakimo1: done, thank you

src/plugins/crypto.c
tbzatek: One more occurrence with `convert_to_type`.

src/plugins/crypto.c
```diff
@@ -2108,6 +2108,69 @@ gboolean bd_crypto_luks_set_uuid (const gchar *device, const gchar *uuid, GError
...
+        g_set_error (error, BD_CRYPTO_ERROR, BD_CRYPTO_ERROR_DEVICE,
+                     "Failed to load device: %s", strerror_l (-ret, c_locale));
+        crypt_free (cd);
+        return FALSE;
+    }
+
+    cd_type = crypt_get_type (cd);
+    if (g_strcmp0 (cd_type, CRYPT_LUKS1) != 0 &&
+        g_strcmp0 (cd_type, CRYPT_LUKS2) != 0) {
+
```
tbzatek, line 2147: Extra empty line.

tbzatek (commented):
Sorry for being such pedantic, it's just that we like to have a clean code.
No functional issues though, this is just about codestyle ;-)

src/plugins/crypto.c
xyakimo1: Thank you for the review, I've made the fixes.

vojtechtrefny (approved):
Thank you

PR #1012, opened 2024-03-26
Fix (potential) crashes in LVM DBus init

vojtechtrefny:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=2247319


tbzatek (approved):
Yup!

PR #1019, opened 2024-04-17
ci: Fix bumping release for Packit builds

vojtechtrefny:
We need to bump the release in the SPEC template so it's not overwritten by the autogen call later.


tbzatek (approved):
hmmmmm :boar:

PR #1024, opened 2024-05-21
utils: Check also for aliases in bd_utils_have_kernel_module

vojtechtrefny:
We want to return true for ext2/3 which are now supported by ext4 module.

### DIFF
--- a/src/utils/module.c
+++ b/src/utils/module.c
@@ -85,6 +85,7 @@ gboolean bd_utils_have_kernel_module (const gchar *module_name, GError **error) 
     gint ret = 0;
     struct kmod_ctx *ctx = NULL;
     struct kmod_module *mod = NULL;
+    struct kmod_list *list = NULL;
     gchar *null_config = NULL;
     const gchar *path = NULL;
     gboolean have_path = FALSE;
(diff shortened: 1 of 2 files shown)


src/utils/module.c
```diff
@@ -100,21 +101,24 @@ gboolean bd_utils_have_kernel_module (const gchar *module_name, GError **error)
...
     if (ret < 0) {
         g_set_error (error, BD_UTILS_MODULE_ERROR, BD_UTILS_MODULE_ERROR_FAIL,
                      "Failed to get the module: %s", strerror_l (-ret, c_locale));
         kmod_unref (ctx);
+        kmod_module_unref_list (list);
         freelocale (c_locale);
         return FALSE;
     }
 
+    mod = kmod_module_get_module (list);
```
tbzatek, line 114: Isn't it better to iterate through the list in case more entries are returned? See https://github.com/lucasdemarchi/kmod/blob/fa67110f896cdef67f42cbc2206ae2a8524acee6/libkmod/README#L53

vojtechtrefny: And do we really need to check all possible matching modules? Isn't it enough for us that to know that there is at least one module matching the name/alias?

tbzatek: Yeah, that crossed my mind as well when having dinner yesterday. I guess any first match might be enough in this case.

PR #1028, opened 2024-06-07
fs: Ignore unused-parameter warning in the FS plugin

vojtechtrefny:
There are some unused parameters in the libext2fs header which together with Wall and Werror means the plugin compilation fails. As a workaround we'll disable the unused-parameter warning for now.

Fixes: #1026

PR #1030, opened 2024-06-12
fs: Fix ignoring errors from libext2fs

vojtechtrefny:
Follow-up for #1028, we still want to show the warning and not completely supress it.


tbzatek (approved):
:yum:

PR #1033, opened 2024-06-13
smart: Add extra_args + related fixes

tbzatek:
This addresses the postponed TODO point from #835.
Related UDisks fixes will come in https://github.com/storaged-project/udisks/pull/1289


src/lib/plugin_apis/smart.api
```diff
@@ -606,15 +606,20 @@ typedef enum {
...
  * bd_smart_ata_get_info:
  * @device: device to check.
+ * @device_type: (nullable): the type of the device or %NULL for autodetection.
  * @error: (out) (optional): place to store error (if any).
  *
  * Retrieve SMART information from the drive.
  *
+ * The optional @device_type argument specifies the type of the device and/or
+ * passthrough protocol.  This is an implementation-specific string, currently
+ * supported only by the smartmontools plugin. See `smartctl -d` for details.
```
vojtechtrefny, line 616: If this is implementation-specific, I think the libatasmart plugin should return an error if you specify a non-NULL device_type. Or at least it should log a warning.

tbzatek: So either I'll rephrase it to something like `currently ignored by the libatasmart plugin` or we document all the applicable values: `ata, scsi[+TYPE], nvme[,NSID], sat[,auto][,N][+TYPE], usbasm1352r,N, usbcypress[,X], usbjmicron[,p][,x][,N], usbprolific, usbsunplus, sntasmedia, sntjmicron[,NSID], sntrealtek, jmb39x[-q],N[,sLBA][,force][+TYPE], jms56x,N[,sLBA][,force][+TYPE], areca,N/E, 3ware,N, hpt,L/M/N, megaraid,N, aacraid,H,L,ID, sssraid,E,S, cciss,N, auto, test`.

The point is that with the current libblockdev plugin abstraction, higher layers have no idea what plugin they're using. There are several other plugin-specific occurrences like these plugins already.

tbzatek: Okay, change to extra args.

vojtechtrefny (approved):

PR #1039, opened 2024-07-02
Upstream kernel VDO support

vojtechtrefny:
Few changes we need in order to support VDO with the latest upstream 6.9 kernel. The biggest change is renaming of the module (`kvdo` → `dm-vdo`) and deprecation of the `/sys/kvdo` API.

### DIFF
--- a/src/plugins/vdo_stats.c
+++ b/src/plugins/vdo_stats.c
@@ -19,10 +19,11 @@
 
 #include <glib.h>
 #include <blockdev/utils.h>
+#include <libdevmapper.h>
+#include <yaml.h>
 
 #include "vdo_stats.h"
-
-#define VDO_SYS_PATH "/sys/kvdo"
+#include "lvm.h"
 
 
 G_GNUC_INTERNAL gboolean
@@ -67,9 +68,9 @@ get_stat_val_double (GHashTable *stats, const gchar *key, gdouble *val) {
 static void add_write_ampl_r_stats (GHashTable *stats) {
     gint64 bios_meta_write, bios_out_write, bios_in_write;
 
-    if (! get_stat_val64 (stats, "bios_meta_write", &bios_meta_write) ||
-        ! get_stat_val64 (stats, "bios_out_write", &bios_out_write) ||
-        ! get_stat_val64 (stats, "bios_in_write", &bios_in_write))
+    if (! get_stat_val64 (stats, "biosMetaWrite", &bios_meta_write) ||
+        ! get_stat_val64 (stats, "biosOutWrite", &bios_out_write) ||
+        ! get_stat_val64 (stats, "biosInWrite", &bios_in_write))
         return;
 
     if (bios_in_write <= 0)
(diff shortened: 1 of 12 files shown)


src/plugins/vdo_stats.c
```diff
@@ -67,9 +68,9 @@ get_stat_val_double (GHashTable *stats, const gchar *key, gdouble *val) {
 static void add_write_ampl_r_stats (GHashTable *stats) {
     gint64 bios_meta_write, bios_out_write, bios_in_write;
 
-    if (! get_stat_val64 (stats, "bios_meta_write", &bios_meta_write) ||
-        ! get_stat_val64 (stats, "bios_out_write", &bios_out_write) ||
-        ! get_stat_val64 (stats, "bios_in_write", &bios_in_write))
+    if (! get_stat_val64 (stats, "biosMetaWrite", &bios_meta_write) ||
```
tbzatek, line 71: Hmm, it's not stated in libblockdev API docs anywhere, however we have an explicit list of key names in UDisks API: https://github.com/storaged-project/udisks/blob/cd8314b638c57b84a3d2855c81fa14c5e4ad2fe2/modules/lvm2/data/org.freedesktop.UDisks2.lvm2.xml#L904-L1011

So either we create just another translation table in UDisks, or we carry it in libblockdev. Kinda reminds me the smart plugin where we had to keep the libatasmart key names.

src/plugins/vdo_stats.c
```diff
@@ -132,76 +133,127 @@ static void add_computed_stats (GHashTable *stats) {
...
-            stats = NULL;
-            break;
-        }
-        g_hash_table_replace (stats, g_strdup (direntry), g_strdup (g_strstrip (val)));
-        g_free (val);
-        g_free (s);
-    }
-    g_dir_close (dir);
-    g_free (stats_dir);
+    yaml_parser_set_input_string (&parser, (guchar *) response, strlen (response));
```
tbzatek, line 198: Who the hell invented putting yaml-formatted string in a dm message??!

src/plugins/vdo_stats.c
```diff
@@ -132,76 +133,127 @@ static void add_computed_stats (GHashTable *stats) {
...
+                /* end of mapping, discard the prefix used */
+                g_free (prefix);
+                prefix = NULL;
+                break;
+            /* actual data */
+            case YAML_SCALAR_TOKEN:
+                if (next_token == PARSE_NEXT_KEY) {
+                    if (prefix) {
+                        key = g_strdup_printf ("%s%s", prefix, (const gchar *) token.data.scalar.value);
+                        len = strlen(prefix);
```
tbzatek, line 238: missing space ;-)

src/plugins/vdo_stats.c
vojtechtrefny: The original output of VDO stats was also in yaml (but formatted differently) so maybe they wanted to keep it? But definitely unexpected.

src/plugins/vdo_stats.c
vojtechtrefny: I saw the udisks documentation and if it was up to me, I'd be ok with changing the "output". VDO was never part of the upstream kernel before so I'd just pretend the past didn't happen and say this is the new "API" and we ignore everything else. I am already ignoring the existence of `kvdo` with this change.

tbzatek: Good point about the first public version. After all, VDO guys need to sign up for some output format stability themselves (unless dm messages are intentionally not part of the guarantees).

tbzatek (approved):
OK, go for it.

src/plugins/vdo_stats.c
vojtechtrefny: The output should be stable, but I wouldn't be surprised if it changes in the future. I hope it will at least always be a valid YAML so the parsing works.

PR #1045, opened 2024-07-16
part: Add human readable partition type to BDPartSpec

vojtechtrefny:
Unfortunately fdisk doesn't provide API to translate between the GUID and the "name" directly, we can only get it for existing partitions.

Related: https://github.com/storaged-project/blivet/issues/1258

### DIFF
--- a/src/plugins/part.h
+++ b/src/plugins/part.h
@@ -51,6 +51,7 @@ typedef struct BDPartSpec {
     guint64 size;
     gboolean bootable;
     guint64 attrs;
+    gchar *type_name;
 } BDPartSpec;
 
 BDPartSpec* bd_part_spec_copy (BDPartSpec *data);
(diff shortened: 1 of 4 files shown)


src/plugins/part.h
```diff
@@ -46,6 +46,7 @@ typedef struct BDPartSpec {
     gchar *uuid;
     gchar *id;
     gchar *type_guid;
+    gchar *type_name;
```
tbzatek, line 49: Uhm, this looks like ABI break. Suggesting to move this member at the end of the struct.

vojtechtrefny: Ok, but it will still be an ABI break even when adding it to the end of the struct, right?

tbzatek: Yes, but it's a lesser evil.

PR #1050, opened 2024-08-15
ci: Remove priority from Testing farm repositories

vojtechtrefny:
Testing farm sets priority of their repositories to 9 which means the repo is prioritized over all other repositories including our Copr repo with daily builds preventing the latest libblockdev from being installed in the reverse dependency test.


tbzatek (approved):
deprioritize!

PR #1063, opened 2024-09-26
smart: Clarify use of ID_ATA_SMART_ACCESS udev property

tbzatek:
Surprising discovery by reading libatasmart sources. Haven't found any other public mention and found this quite useful.

### DIFF
--- a/src/plugins/smart/smart-common.c
+++ b/src/plugins/smart/smart-common.c
@@ -130,16 +130,30 @@
  *
  * ## Device type detection, multipath #
  *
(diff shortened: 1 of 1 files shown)


tbzatek:
I'm still unsure how to take on this in UDisks. Would be nice to respect the udev property as an override for smartmontools as well, though there might be situations where `smartctl`'s autodetection might be smarter.

vojtechtrefny (approved):
Configuring a library with a udev rule. Lovely.

Issue #943, opened 2023-07-23
test suite requires package to be built with tools enabled


mbiebl:
Version: 3.0.2-1

I'm trying to run the test suite (via `python3 tests/run_tests.py` in an autopkgtest environment.
The Debian package for libblockdev is currently built using `--without-tools`.
The test suite does not seem to cope with that properly:

```
======================================================================
ERROR: test_vfat_resize (fs_tests.vfat_test.VfatResize.test_vfat_resize)
Verify that it is possible to resize an vfat file system
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/autopkgtest.ZLSthX/tree/tests/fs_tests/vfat_test.py", line 292, in test_vfat_resize
    succ = BlockDev.fs_vfat_resize(self.loop_dev, 130 * 1024**2)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: g-bd-utils-exec-error-quark: The 'vfat-resize' utility is not available (6)

======================================================================
ERROR: test_vfat_available (fs_tests.vfat_test.VfatTestAvailability.test_vfat_available)
Verify that it is possible to check vfat tech availability
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/autopkgtest.ZLSthX/tree/tests/fs_tests/vfat_test.py", line 47, in test_vfat_available
    available = BlockDev.fs_is_tech_avail(BlockDev.FSTech.VFAT,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: g-bd-utils-exec-error-quark: The 'vfat-resize' utility is not available (6)


...
----------------------------------------------------------------------
Ran 487 tests in 2465.088s

FAILED (failures=3, errors=2, skipped=21
```

vojtechtrefny:
We need something similar to the `config_h.py` we did in UDisks (https://github.com/storaged-project/udisks/pull/821) to be able to tell which compile time options were provided, we generally expect that everything was enabled during the build.

mbiebl:
FYI: I've added https://salsa.debian.org/utopia-team/libblockdev/-/commit/a367942a152bbd1baf156ddf984c146b5f365ed4  for the time being

mbiebl:
Not sure if there is a nicer/better way to skip tests

[...]

PR #1066, opened 2024-10-01
Skip tests for plugins disabled during compile time

vojtechtrefny:
Fixes: #943

The `config_h.py` file functionality is basically copy-pasted from udisks.


tbzatek (approved):
Interesting!

tbzatek (approved):

PR #1071, opened 2024-11-06
fs: Add filesystem size limits to BDFSFeatures

vojtechtrefny:
The size limits are taken from blivet.


vojtechtrefny:
> Interesting, I was expecting at least some _None_ or _INF_ values.

To be fair `16 TiB - 1` is limit of guint64 so it kinda is _INF_ :-)

PR #1072, opened 2024-11-06
nvme: Avoid element-type g-i annotations

tbzatek:
For some reason this is causing more harm and seems to work fine when absent.


vojtechtrefny (approved):
LGTM, fixes `bd_nvme_get_self_test_log` for me.

PR #1076, opened 2024-11-07
Version bump to 3.2.1

vojtechtrefny:
We did a new bugfix release from `3.2.x-branch` so we need to bump the version here as well so we are not behind on `master`


tbzatek (approved):
Ship it!

PR #1077, opened 2024-11-13
tests: Add a simple test case for bd_crypto_device_seems_encrypted

vojtechtrefny:


tbzatek (approved):
The owls are not what they seem!

PR #1078, opened 2024-11-13
packit: Fix replacement of placeholders in post-upstream-clone

nforro:
I'm not exactly sure why, but while the sed expression does work in Copr, it doesn't work in Packit sandbox environment when running propose-downstream, hence the failures: https://github.com/storaged-project/libblockdev/runs/32657399598

This should hopefully fix it.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,7 +1,7 @@
 actions:
   post-upstream-clone:
     - 'cp dist/libblockdev.spec.in dist/libblockdev.spec'
-    - 'sed -i -e "s/@WITH_.+@/1/g" -e "s/@MAJOR_VER@/3/g" dist/libblockdev.spec'
+    - 'sed -i -e "s/@WITH_..*@/1/g" -e "s/@MAJOR_VER@/3/g" dist/libblockdev.spec'
   create-archive:
     - './autogen.sh'
     - './configure'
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
I was wondering why it failed, I suspected it was because I made the release from a non-default branch, but I didn't have time to investigate it further. Interestingly it worked for the previous release two months ago: https://dashboard.packit.dev/jobs/propose-downstream/10806

vojtechtrefny (approved):
Thank you.

nforro:
> Interestingly it worked for the previous release two months ago: https://dashboard.packit.dev/jobs/propose-downstream/10806

Right, but at that time there was still `sed -E`: https://github.com/storaged-project/libblockdev/blob/27881cb6e57eb0146c51d032ea72fbb3cd6bda7e/.packit.yaml#L4

vojtechtrefny:
> > Interestingly it worked for the previous release two months ago: https://dashboard.packit.dev/jobs/propose-downstream/10806
> 
> Right, but at that time there was still `sed -E`:
> 
> https://github.com/storaged-project/libblockdev/blob/27881cb6e57eb0146c51d032ea72fbb3cd6bda7e/.packit.yaml#L4

Oh, I forgot about this change. Thanks again for the fix.

PR #1081, opened 2024-12-01
fixed issue 1080

guazhangRH:


vojtechtrefny (commented):
I am not against merging this, but it feels like something that should be fixed in the code, not in tests. But I don't know enough about NVMe so I'll leave the decision on @tbzatek

tbzatek (changes requested):
I would prefer to avoid using `sleep` for things where we can poll for something. In this case you may want to wait for specific subsystem nqn to appear in the list with certain timeout.

tbzatek:
Superseded by #1124

PR #1084, opened 2024-12-02
tests: Try waiting before removing zram module in KBD tests

vojtechtrefny:
The test started failing on Debian with the zram module being busy lets try to add 1s sleep after the swapoff before running rmmod.


tbzatek (approved):
awful.

PR #1086, opened 2025-01-16
tests: Skip zram tests on Debian because of kernel bug

vojtechtrefny:


tbzatek (approved):
Stop it on Debian first then?

PR #1092, opened 2025-01-31
lvm: Require higher version of LVM for lvm.conf parsing

vojtechtrefny:
The --valuesonly option we use is not available on LVM < 2.03.17


tbzatek (approved):
I guess it's fine.

PR #1094, opened 2025-02-27
tests: Skip LVM VDO tests if 'vdoformat' is not available

vojtechtrefny:
LVM VDO still uses the vdoformat binary which might not be available everywhere.

Related: https://github.com/storaged-project/udisks/issues/1353

### DIFF
--- a/tests/lvm_dbus_tests.py
+++ b/tests/lvm_dbus_tests.py
@@ -2116,6 +2116,9 @@ class LVMVDOTest(LVMTestCase):
         if lvm_version < Version("2.3.07"):
             raise unittest.SkipTest("LVM version 2.3.07 or newer needed for LVM VDO.")
 
+        if not shutil.which("vdoformat"):
+            raise unittest.SkipTest("vdoformat executable not found in $PATH, skipping.")
+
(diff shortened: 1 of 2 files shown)


tbzatek (approved):
I just worry that any missing dependency that could have been installed may pass silently undetected.

vojtechtrefny:
> I just worry that any missing dependency that could have been installed may pass silently undetected.

Yes, this a problem with all these skips, but I am not sure what to do with that. We might try to create some better summary for the tests and actually check the skipped tests, but that would be quite a lot of work.

mbiebl:
I've added "vdo" as test dependency for now. So there is no need to skip the test for Debian's sake:
https://salsa.debian.org/utopia-team/libblockdev/-/commit/695a899274a1b9916185b7c63e174e0b3576e66b

I very much appreciate it though, that you do care for Debian.

vojtechtrefny:
> So there is no need to skip the test for Debian's sake

It still makes sense to skip the test if vdo is not available. We need some way to be sure we are just not skipping all tests because of some CI infrastructure issue, but that's a different problem.

PR #1095, opened 2025-03-06
crypto: Add a function to set persistent flags for LUKS

vojtechtrefny:
This will be used to set the allow-discards flag on LUKS devices during installation by Blivet.


tbzatek (approved):
Looks simple enough!

PR #1097, opened 2025-03-18
smart: Use drive self-assessment as an overall status

tbzatek:
The libatasmart attribute overall status differs slightly from the drive SMART self-assessment and is very sensitive for particular status values. Such status should fit more like a pre-fail warning, no reason to fail hard the global assessment. Even a single reallocated sector would cause a warning, while the drive could be quite healthy otherwise.


vojtechtrefny (approved):
Makes sense, LGTM

PR #1098, opened 2025-03-30
tests: Fix self/sanitize log on loop devices with latest nvmetcli

vojtechtrefny:
With the latest nvmetcli we can now get some logs even from the "fake" nvme devices used in tests. Unfortunately nvmetcli doesn't have an option to get its version so we can't really tell which version we have.

---

We should also check the test results, but right now I am more interested in fixing the CI. Fun fact: Fedora 42 now has newer version of nvmetcli than rawhide so the tests now fail only on F42.

### DIFF
--- a/tests/nvme_test.py
+++ b/tests/nvme_test.py
@@ -203,10 +203,15 @@ class NVMeTestCase(NVMeTest):
         with self.assertRaisesRegex(GLib.GError, r".*Failed to open device .*': No such file or directory"):
             BlockDev.nvme_get_self_test_log("/dev/nonexistent")
 
-        message = r"NVMe Get Log Page - Device Self-test Log command error: Invalid Field in Command: A reserved coded value or an unsupported value in a defined field|NVMe Get Log Page - Device Self-test Log command error: unrecognized"
-        with self.assertRaisesRegex(GLib.GError, message):
-            # Cannot retrieve self-test log on a nvme target loop devices
-            BlockDev.nvme_get_self_test_log(self.nvme_dev)
+        try:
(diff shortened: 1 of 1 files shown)


tbzatek:
I'm able to reproduce this error with `kernel-6.14.0-0.rc6.20250314git695caca9345a.53.fc43.x86_64`, `kernel-6.15.0-0.rc0.20250327git1a9239bb4253.5.fc43.x86_64` and both the `nvmetcli-0.7-15.fc41.noarch` and `nvmetcli-0:0.8-1.fc42.noarch`.

I believe this is kernel related since `nvmetcli` is just a tiny wrapper around debugfs.

tbzatek:
```
[tbzatek@localhost libnvme]$ git bisect good
adee4ed1c8c8de918fa0dd4087d84d28692182e7 is the first bad commit
commit adee4ed1c8c8de918fa0dd4087d84d28692182e7
Author: letli <letian.li@memblaze.com>
Date:   Thu Dec 5 11:30:02 2024 +0800

    ioctl: get_log_page by nvme uring cmd
    
    Use io_uring for fetching log pages.
    
    This showed about a 10% performance improvement for some large log pages.
    
    Signed-off-by: letli <letian.li@memblaze.com>
```

vojtechtrefny:
> I believe this is kernel related since nvmetcli is just a tiny wrapper around debugfs.

I originally thought this to be `nvmetcli` related because it happened only on Fedora 42 in our CI and `nvmetcli` was the only package with newer version in F42, but that was caused by `nvme-cli` not being updated because of some packaging conflicts on rawhide, after updating rawhide manually the tests now fail on both 42 and rawhide.

PR #1101, opened 2025-04-03
tests: Temporarily skip some nvme tests on latest Fedora

vojtechtrefny:
nvme_get_sanitize_log and nvme_get_self_test_log now return some invalid data for our test loop based NVMe devices. We'll skip the tests until this is resolved.

See also #1098 and https://github.com/linux-nvme/libnvme/issues/980

### DIFF
--- a/tests/skip.yml
+++ b/tests/skip.yml
@@ -71,3 +71,9 @@
     - distro: "fedora"
       version: "43"
       reason: "The fake in-memory namespace doesn't support sector mode in kernel 6.15"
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Hmm, just don't forget to revert this once we have proper libnvme fix!

vojtechtrefny:
Closing, the fix will be released soon in Fedora and it doesn't make sense to merge this for just few days and then revert it again.

PR #1104, opened 2025-04-21
[crypto/nvme/smart] Align structs 8 byte for 64-bit platforms

GermanAizek:
@vojtechtrefny, hello.
I'm new contributor at you.
This PR change should be guaranteed to lead to more frequent structs entry into CPU cache, which can greatly affect performance if aligned structures are used frequently. I hope that you know advantage this optimization method at the architectural level codebase and its inconveniences as stylistic ABI breakdown.

More info about technique:
https://stackoverflow.com/a/20882083
https://zijishi.xyz/post/optimization-technique/learning-to-use-data-alignment/ https://en.wikipedia.org/wiki/Data_structure_alignment


tbzatek:
Thank you for your contribution. I'm afraid we can't change public structs on released interfaces due to API stability guarantees.

Issue #1100, opened 2025-03-31
Blivet incorrectly displays BTRFS subvolumes with spaces


Ponywka:
Blivet incorrectly displays BTRFS subvolumes with spaces

I have separated BTRFS subvolumes to exclude some folders from auto-snapshot.

One of them - `@home/ponywka/VirtualBox VMs`:

![Image](https://github.com/user-attachments/assets/e225b04e-0916-4f6a-be51-63c5f357d5e7)

I think it should show full name of BTRFS subvolume - `@home/ponywka/VirtualBox VMs` instead of `@home/ponywka/VirtualBox`

vojtechtrefny:
Moving to libblockdev:

```
BlockDev.btrfs_list_subvolumes("/mnt/test")
Out[5]: 
[BlockDev.BtrfsSubvolumeInfo (BDBtrfsSubvolumeInfo) instance (0x7f1bf866d2b0)
  id: 259
  parent_id: 5
  path: a]

# btrfs subvolume list /mnt/test
ID 259 gen 14 top level 5 path a b
```

But we might need some fixes in blivet as well after this is fixed, device names with spaces will probably cause some issues there too.

PR #1116, opened 2025-07-01
Fix parsing subvolumes with spaces in name

vojtechtrefny:
Fixes: #1100


tbzatek (approved):
I trust you know regexes well enough.

PR #1120, opened 2025-08-14
tests: Fix SCSI debug disks cleanup in SMART tests

vojtechtrefny:
This was originally one of the issues found by pylint and part of #1119 but for some reason fixing this causes an infinite loop in `_setup_scsi_debug` on Debian and CentOS (but not Fedora). We need some more time to investigate this and it's out of scope for the tests speed up so I am keeping this as a draft for now.

### DIFF
--- a/tests/smart_test.py
+++ b/tests/smart_test.py
@@ -1,10 +1,8 @@
 import unittest
 import os
-import glob
-import time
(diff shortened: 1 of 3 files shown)


tbzatek (approved):
Nice cleanup!

PR #1124, opened 2025-09-03
tests: Fix NVMe device timing issues with intelligent polling mechanism

### LINKED ISSUE #1080: can not get nvme ns immediately  after create nvme ctrl
maybe need wait a sec after created nvme ctrl in nvme_test
```
======================================================================
ERROR: test_ns_info (nvme_test.NVMeTestCase.test_ns_info)
Test namespace info
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/rpmbuild/BUILD/libblockdev-3.2.0/tests/nvme_test.py", line 92, in test_ns_info
    info = BlockDev.nvme_get_namespace_info(self.nvme_ns_dev)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: g-bd-nvme-error-quark: Failed to open device '/dev/nvme0n1': No such file or directory (1)
```

```
======================================================================
FAIL: test_connect_multiple_ns (nvme_test.NVMeFabricsTestCase.test_connect_multiple_ns)
Test connect and disconnect multiple namespaces
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/rpmbuild/BUILD/libblockdev-3.2.0/tests/nvme_test.py", line 443, in test_connect_multiple_ns
    self.assertEqual(len(namespaces), NUM_NS)
AssertionError: 1 != 3
----------------------------------------------------------------------
```
nvme-cli-2.10.2-1.el10.x86_64
nvmetcli-0.8-1.el10.noarch

```
$ git diff tests/utils.py 
diff --git a/tests/utils.py b/tests/utils.py
index a9b8e187..9fe93605 100644
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -251,7 +251,7 @@ def find_nvme_ctrl_devs_for_subnqn(subnqn):
                         dev_paths += [path]
                 except:
                     pass
-
+    time.sleep(1)
     ret, out, err = run_command("nvme list --output-format=json --verbose")
     if ret != 0:
         raise RuntimeError("Error getting NVMe list: '%s %s'" % (out, err))

```
```
for i in $(seq 30);do python3 run_tests.py -s nvme_test || break;done
```
can not reproduce the error after apply the patch, maybe someone have a better solution for the error. 
if no, I will create a MR for it.

ermezavx:
Fixes #1080 

This PR addresses the timing issue where NVMe namespace devices may not be immediately available after controller creation, causing intermittent test failures.

## Problem
The current implementation performs immediate device lookup without considering that NVMe devices may need time to appear in the system after target creation and connection.

## Solution
Instead of the simple `sleep(1)` approach proposed in #1081, this implements an intelligent polling mechanism with exponential backoff:

### Key Features
- **Smart Polling**: 100ms initial delay with 1.2x exponential backoff (capped at 500ms)
- **Reasonable Timeout**: 5-second default timeout (configurable)
- **Device Readiness**: Includes `udevadm settle` to ensure proper device initialization
- **Early Return**: Returns immediately when devices are found
- **Backward Compatible**: No breaking changes to existing API

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -274,11 +274,43 @@ def clean_scsi_debug(scsi_debug_dev):
     except:
         pass
 
-def find_nvme_ctrl_devs_for_subnqn(subnqn):
+def _wait_for_nvme_controllers_ready(subnqn, timeout=3):
+    """
+    Wait for NVMe controllers with matching subsystem NQN to be in live state
+    
+    :param str subnqn: subsystem nqn to match controllers against
+    :param int timeout: timeout in seconds (default: 3)
+    """
+    start_time = time.time()
+    
+    while time.time() - start_time < timeout:
+        try:
+            for ctrl_path in glob.glob("/sys/class/nvme/nvme*/"):
+                state_file = os.path.join(ctrl_path, "state")
+                subsysnqn_file = os.path.join(ctrl_path, "subsysnqn")
+                try:
+                    state = read_file(state_file).strip()
+                    controller_subnqn = read_file(subsysnqn_file).strip()
+                    if state == "live" and controller_subnqn == subnqn:
+                        # Found a matching live controller
+                        os.system("udevadm settle")
+                        return
+                except:
+                    continue
+                
+        except:
+            pass
+        
+        time.sleep(1)
+    
+    os.system("udevadm settle")
+
+def find_nvme_ctrl_devs_for_subnqn(subnqn, wait_for_ready=True):
     """
     Find NVMe controller devices for the specified subsystem nqn
 
     :param str subnqn: subsystem nqn
+    :param bool wait_for_ready: whether to wait for controllers to be ready (default: True)
     """
 
     def _check_subsys(subsys, dev_paths):
@@ -295,6 +327,9 @@ def find_nvme_ctrl_devs_for_subnqn(subnqn):
                 except:
                     pass
 
+    # Wait for controllers to be ready if requested
+    if wait_for_ready:
+        _wait_for_nvme_controllers_ready(subnqn)
     ret, out, err = run_command("nvme list --output-format=json --verbose")
     if ret != 0:
         raise RuntimeError("Error getting NVMe list: '%s %s'" % (out, err))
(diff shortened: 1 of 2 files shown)


tbzatek:
The best approach is to wait for a `live` controller status in the respective sysfs path first and then wait for udev to finish probing the device.

ermezavx:
> The best approach is to wait for a `live` controller status in the respective sysfs path first and then wait for udev to finish probing the device.

I have made the changes according to your suggestions. Please help review it. Thank you.

tests/utils.py
```diff
@@ -274,11 +274,52 @@ def clean_scsi_debug(scsi_debug_dev):
...
+    :returns: True if ready, False if timeout
+    """
+    start_time = time.time()
+    wait_time = 0.1  # Start with 100ms
+    max_wait = 0.5   # Cap at 500ms
+    
+    while time.time() - start_time < timeout:
+        try:
+            live_controllers = set()  # Use set for automatic deduplication
+            for ctrl_path in glob.glob("/sys/class/nvme/nvme*/"):
```
tbzatek, line 292: You need to check only those controllers you're waiting for. I.e. the controller's subsystem nqn should match the expected test NQN.

tests/utils.py
```diff
@@ -274,11 +274,48 @@ def clean_scsi_debug(scsi_debug_dev):
...
+                        os.system("udevadm settle")
+                        return True
+                except:
+                    continue
+                
+        except:
+            pass
+        
+        time.sleep(wait_time)
+        wait_time = min(wait_time * 1.2, max_wait)  # Slower exponential backoff
```
vojtechtrefny, line 308: My [original comment](https://github.com/storaged-project/libblockdev/pull/1124#discussion_r2318148827) about this being unnecessarily complicated  is still valid here.

tests/utils.py
```diff
@@ -274,11 +274,48 @@ def clean_scsi_debug(scsi_debug_dev):
...
+    
+    os.system("udevadm settle")
+    return False
```
vojtechtrefny, line 311: Why this function returns boolean? It's not checked anywhere where it is called.

vojtechtrefny (changes requested):

tests/utils.py
ermezavx: You are right, there is no need to return a boolean type.

vojtechtrefny:
Jenkins, ok to test.

PR #1125, opened 2025-09-04
lvm-dbus: Fix calling lvcreate with empty list of PVs

vojtechtrefny:
GLib really dislikes us calling 'g_variant_builder_end' on the empty array in this case.


tbzatek (approved):
Hmm, makes sense, I believe I've seen something similar before.

PR #1136, opened 2025-10-15
ci: Add a GH action to check for exported private symbols

vojtechtrefny:
With this we'll hopefully avoid issues like #1135 in the future.


vojtechtrefny:
This is how a failed job looks like: https://github.com/storaged-project/libblockdev/actions/runs/18528218527/job/52804039215

PR #1165, opened 2026-02-05
Check for non-existing/invalid/empty devices/paths in dm plugin and utils

vojtechtrefny:
## Summary by CodeRabbit

* **Bug Fixes**
  * Empty device-mapper name/node inputs now raise clear validation errors instead of proceeding.
  * Device resolution now verifies resolved paths are block devices and returns explicit errors for invalid targets.

* **Tests**
  * Expanded test coverage for device resolution and symlink handling using storage-backed test setups, including checks for non-block-device failures and symlink resolution behavior.


vojtechtrefny:
> New private symbols found:
_is_block_device

Good to see the test is actually working. (Of course I definitely meant to test this, I wouldn't forget to add the `static` keyword like that.)

tbzatek (approved):

PR #1198, opened 2026-06-02
fs: Fix parsing tune.exfat output with variable whitespace

vojtechtrefny:
The format of tune.exfat output changed from "key : value" to
"key      : value" with variable padding around the colon. Replace
hard-coded prefix matching with key-name matching and a helper that
finds the value after the first colon on the line.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced exFAT filesystem information parsing to reliably extract sector size, sector count, and cluster count.


tbzatek (approved):
The `g_strrstr()` usage looks suspicious but I think that's still okay.

PR #1202, opened 2026-07-31
fs: support non-ASCII VFAT labels via locale-derived codepage

Johnson-zs:
fatlabel defaults to DOS codepage 850, which cannot encode characters outside Western European languages. When a user sets a VFAT label containing CJK, Cyrillic or other non-ASCII characters, fatlabel fails with "Cannot convert input sequence" and the label is not changed.

Pass the OEM codepage matching the system locale to fatlabel via the -c option when the label contains non-ASCII characters. Keep the byte-count check (strlen) in bd_fs_vfat_check_label as a conservative upper bound; fatlabel performs the exact codepage-length check during iconv conversion.

Decode the raw OEM bytes returned by blkid back to UTF-8 in bd_fs_vfat_get_info so labels round-trip correctly. Also add a NULL check to bd_fs_vfat_check_label and tests for non-ASCII label set/get and the byte-limit check.

This mirrors how Windows derives the OEM codepage from the system locale. Note: the codepage is derived from the daemon environment (udisksd), not the calling user's session — this is a known limitation documented in the code.

## Summary by CodeRabbit

* **New Features**
  * VFAT volume labels now support locale-aware conversion for non-ASCII characters.
  * Filesystem labels are converted to UTF-8 when possible.
  * Non-ASCII labels passed to `fatlabel` automatically use the appropriate code page.

* **Bug Fixes**
  * VFAT label validation now enforces the 11-byte encoded limit.
  * Invalid, forbidden, or missing labels are rejected more reliably.
  * Conversion failures preserve valid original label data as a fallback.

### DIFF
--- a/src/plugins/fs/vfat.c
+++ b/src/plugins/fs/vfat.c
@@ -277,25 +277,247 @@ gboolean bd_fs_vfat_repair (const gchar *device, const BDExtraArg **extra, GErro
 }
 
 /**
+ * _vfat_locale_codepage:
+ *
+ * Determine the DOS/OEM codepage that matches the system locale, mirroring
+ * how Windows derives the OEM codepage from the system locale. FAT volume
+ * labels are stored as 8-bit OEM codepage bytes, so the codepage used to
+ * write a label must match the locale of the system that created it.
+ *
+ * The locale is read from the environment (LC_ALL > LC_CTYPE > LANG) rather
+ * than calling setlocale(), to avoid side effects on the host process
+ * (libblockdev runs inside udisksd).
+ *
+ * Returns: the codepage number, or 850 (the dosfstools default) if the locale
+ *          cannot be determined or has no known mapping.
+ */
+static guint
+_vfat_locale_codepage (void) {
+    /* Locale-to-codepage table based on the DOS CODEPAGES table in
+     * fatlabel(8), which mirrors how Windows maps the "Language for
+     * non-Unicode programs" setting to an OEM codepage. Entries are matched
+     * by longest-prefix so region variants (e.g. "zh_TW" over "zh",
+     * "en_GB" over "en") win. When nothing matches, the fatlabel default
+     * CP850 is returned. */
+    typedef struct {
+        const gchar *prefix;
+        guint codepage;
+    } VfatLocaleCodepage;
+    static const VfatLocaleCodepage codepage_table[] = {
+        /* --- CJK (multi-byte) --- */
+        { "zh_CN", 936 }, { "zh_SG", 936 },   /* Chinese (Simplified) / GBK */
+        { "zh_TW", 950 }, { "zh_HK", 950 },   /* Chinese (Traditional) / Big5 */
+        { "zh", 936 },                         /* Chinese default — Simplified */
+        { "ja",    932 },                       /* Japanese / Shift-JIS */
+        { "ko",    949 },                       /* Korean / UHC */
+        /* --- Cyrillic (CP866) --- */
+        { "ba", 866 }, { "be", 866 }, { "bg", 866 },
+        { "ky", 866 }, { "mk", 866 }, { "mn", 866 }, { "ru", 866 },
+        { "tg", 866 }, { "tt", 866 }, { "uk", 866 },
+        { "az@cyrillic", 866 }, { "uz@cyrillic", 866 },
+        /* --- Serbian / Bosnian Cyrillic (CP855) --- */
+        { "sr", 855 }, { "bs@cyrillic", 855 },
+        /* --- Central/Eastern Europe (CP852) --- */
+        { "bs", 852 }, { "cs", 852 }, { "hr", 852 }, { "hu", 852 },
+        { "pl", 852 }, { "ro", 852 }, { "sk", 852 }, { "sl", 852 },
+        { "sq", 852 }, { "sr@latin", 852 }, { "tk", 852 },
+        /* --- Turkish / Azeri / Uzbek (Latin) (CP857) --- */
+        { "tr", 857 }, { "az", 857 }, { "uz", 857 },
+        /* --- Greek (CP737) --- */
+        { "el", 737 },
+        /* --- Baltic (CP775) --- */
+        { "et", 775 }, { "lt", 775 }, { "lv", 775 },
+        /* --- Hebrew (CP862) --- */
+        { "he", 862 }, { "iw", 862 },
+        /* --- Arabic / Persian / Urdu / Uyghur (CP720) --- */
+        { "ar", 720 }, { "fa", 720 }, { "ps", 720 }, { "ur", 720 }, { "ug", 720 },
+        /* --- Thai (CP874) --- */
+        { "th", 874 },
+        /* --- Vietnamese (CP1258) --- */
+        { "vi", 1258 },
+        /* --- English variants (region-specific) --- */
+        { "en_GB", 850 }, { "en_IE", 850 }, { "en_AU", 850 },
+        { "en_CA", 850 }, { "en_NZ", 850 }, { "en_JM", 850 },
+        { "en_BZ", 850 }, { "en_TT", 850 }, { "en_BB", 850 },
+        { "en_AG", 850 }, { "en_BS", 850 }, { "en_BW", 850 }, { "en_GH", 850 },
+        { "en_GM", 850 }, { "en_GY", 850 }, { "en_HK", 850 },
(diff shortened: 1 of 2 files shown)


src/plugins/fs/vfat.c
```diff
@@ -276,43 +276,147 @@ gboolean bd_fs_vfat_repair (const gchar *device, const BDExtraArg **extra, GErro
...
+        locale = g_getenv ("LANG");
+
+    if (locale == NULL || *locale == '\0' ||
+        g_strcmp0 (locale, "C") == 0 || g_strcmp0 (locale, "POSIX") == 0 ||
+        g_str_has_prefix (locale, "C."))
+        return 850;
+
+    /* Extract the language[_territory] part before '.' or '@' */
+    g_autofree gchar *lang = g_strdup (locale);
+    gchar *p;
```
tbzatek: Codestyle: variable definitions should be at the beginning of the code block, not inside. More occurrences in other functions.

src/plugins/fs/vfat.c
```diff
@@ -276,43 +276,147 @@ gboolean bd_fs_vfat_repair (const gchar *device, const BDExtraArg **extra, GErro
...
+        return 936;   /* GBK */
+    if (g_str_has_prefix (lang, "zh_TW") || g_str_has_prefix (lang, "zh_HK"))
+        return 950;   /* Big5 */
+    if (g_str_has_prefix (lang, "ja"))
+        return 932;   /* Shift-JIS */
+    if (g_str_has_prefix (lang, "ko"))
+        return 949;   /* UHC (Korean) */
+    if (g_str_has_prefix (lang, "ru") || g_str_has_prefix (lang, "uk") ||
+        g_str_has_prefix (lang, "be"))
+        return 866;   /* Cyrillic (DOS) */
```
tbzatek: This doesn't look complete. Simple query to AI revealed more codepages in use. That said, there might be complete codepage values somewhere. Anyway, do we want this conversion table here or better to make it more universal? @vojtechtrefny ?

```bash
#!/bin/bash

# 1. Grab the current LANG variable, defaulting to en_US if empty
CURRENT_LANG="${LANG:-en_US.UTF-8}"

# 2. Extract the base language code (e.g., "cs" from "cs_CZ.UTF-8")
LANG_CODE=$(echo "$CURRENT_LANG" | cut -d'_' -f1)

# Extract the region-specific locale (e.g., "en_GB") to handle regional differences
LOCALE_SPECIFIC=$(echo "$CURRENT_LANG" | cut -d'.' -f1)

# 3. Map the language to the appropriate DOS codepage
case "$LANG_CODE" in
    cs|pl|hu|sk|sl|hr|ro|bs|sr)
        CP="852" ;; # Central/Eastern Europe (Slavic/Latin-2)
    ru|uk|be|bg|mk)
        CP="866" ;; # Cyrillic
    de|fr|es|it|nl|pt|sv|fi|da|no|is)
        CP="850" ;; # Western Europe (Multilingual Latin-1)
    en)
        # US English defaults to 437; UK and others often use 850
        case "$LOCALE_SPECIFIC" in
            en_GB|en_IE) CP="850" ;;
            *)           CP="437" ;;
        esac
        ;;
    tr) CP="857" ;; # Turkish
    el) CP="737" ;; # Greek
    et|lv|lt) CP="775" ;; # Baltic
    he) CP="862" ;; # Hebrew
    ar) CP="864" ;; # Arabic
    ja) CP="932" ;; # Japanese (Shift-JIS)
    zh)
        # Differentiate between Traditional and Simplified Chinese
        case "$LOCALE_SPECIFIC" in
            zh_TW|zh_HK) CP="950" ;; # Traditional
            *)           CP="936" ;; # Simplified
        esac
        ;;
    ko) CP="949" ;; # Korean
    th) CP="874" ;; # Thai
    *)  CP="437" ;; # Default DOS fallback (Standard US)
esac

echo "Your locale ($CURRENT_LANG) maps to DOS codepage: $CP"```

Comment (tbzatek) [reply src/plugins/fs/vfat.c]: Some more AI magic, but I see it's still incomplete.

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

// A simple structure to map 2-letter ISO 639-1 language codes to DOS codepages
struct LocaleMap {
    const char *lang_prefix;
    int dos_codepage;
};

static const struct LocaleMap codepage_map[] = {
    {"cs", 852}, {"pl", 852}, {"hu", 852}, {"sk", 852}, {"sl", 852}, 
    {"hr", 852}, {"ro", 852}, {"bs", 852}, {"sr", 852}, // Latin-2 (Central Europe)
    {"ru", 866}, {"uk", 866}, {"be", 866}, {"bg", 866}, // Cyrillic
    {"de", 850}, {"fr", 850}, {"es", 850}, {"it", 850}, 
    {"nl", 850}, {"pt", 850}, {"sv", 850}, {"fi", 850}, // Latin-1 (Western Europe)
    {"tr", 857}, // Turkish
    {"el", 737}, // Greek
    {"et", 775}, {"lv", 775}, {"lt", 775},              // Baltic
    {"he", 862}, // Hebrew
    {"ar", 864}, // Arabic
    {"ja", 932}, // Japanese (Shift-JIS)
    {"ko", 949}, // Korean
    {"th", 874}, // Thai
    {NULL, 437}  // Sentinel / Fallback
};

int get_dos_codepage(const char *locale) {
    if (!locale || strlen(locale) < 2) {
        return 437; // Default US fallback
    }

    // Handle special region-specific cases first
    if (strncmp(locale, "zh_TW", 5) == 0 || strncmp(locale, "zh_HK", 5) == 0) return 950; // Traditional Chinese
    if (strncmp(locale, "zh", 2) == 0) return 936; // Simplified Chinese
    if (strncmp(locale, "en_GB", 5) == 0 || strncmp(locale, "en_IE", 5) == 0) return 850; // UK/Ireland English

    // Extract the primary 2-letter language code
    char lang[3];
    strncpy(lang, locale, 2);
    lang[2] = '\0';

    // Scan the mapping table
    for (int i = 0; codepage_map[i].lang_prefix != NULL; i++) {
        if (strcmp(lang, codepage_map[i].lang_prefix) == 0) {
            return codepage_map[i].dos_codepage;
        }
    }

    return 437; // Standard DOS fallback
}

int main() {
    // Read the current locale from the environment
    const char *env_lang = getenv("LANG");
    
    int cp = get_dos_codepage(env_lang);
    
    printf("Locale '%s' maps to DOS codepage: %d\n", env_lang ? env_lang : "(none)", cp);
    return 0;
}
```

Johnson-zs: Thanks for the review. Rather than keep expanding the locale→codepage table, I'd like to drop it from libblockdev entirely and have a quick sanity check on the direction before reworking the code.

The idea is to separate mechanism from policy: libblockdev passes an explicit codepage to `fatlabel -c` and never maps locales itself. Add three entry points taking a `guint codepage` (0 == omit `-c`, i.e. the current fatlabel default), and keep the existing functions as thin `codepage == 0` wrappers:

- `bd_fs_vfat_set_label_ext(device, label, codepage, error)`
- `bd_fs_vfat_check_label_ext(label, codepage, error)` — enforce the 11-byte limit on the OEM-encoded bytes, not on the UTF-8 `strlen`
- `bd_fs_vfat_label_to_utf8(label, codepage)` — decode blkid's raw OEM bytes back to UTF-8

On the udisks side, `handle_set_label` would read an optional `"codepage"` (uint32) key from the existing `options a{sv}` argument of `org.freedesktop.UDisks2.Filesystem.SetLabel` and pass it into `bd_fs_vfat_set_label_ext`. This is a non-breaking extension: the `SetLabel` DBus signature is unchanged, `"codepage"` is optional (absent → `0` → current behaviour), and the locale→codepage choice stays with the caller that owns the locale context rather than the daemon.

This resolves the two concerns directly. "The table doesn't look complete / codepages grow with the kernel" — there's no table to be complete anymore; any codepage dosfstools and the kernel NLS support can be passed, and libblockdev doesn't go stale as NLS evolves. "Make it more universal / architectural" — the codepage is an explicit parameter threaded up from the caller, with udisks merely relaying it.

No public signatures change — existing `bd_fs_vfat_set_label`/`check_label` keep their signatures and current behaviour, so current callers are unaffected.

vojtechtrefny: I think the current approach (having the (possibly incomplete) table in libblockdev) is better. We should try to make the table as complete as possible (and more codes can be added later). It's not ideal, but moving the responsibility to user of udisks API is also not the best. So I can live with this in libblockdev.

PR #1203, opened 2026-08-10
lvm: Add warning when user specifies invalid VDO index memory size

vojtechtrefny:
The invalid values are silently rounded up by LVM.

This was found by our tests mostly by accident, bigger index now needs more disk space so our tests started failing on systems with the latest LVM.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for VDO index-memory settings during pool creation and conversion.
  - Warns when configured memory is not 256, 512, or 768 MiB, or a multiple of 1024 MiB.
  - Prevents unsupported nonzero index-memory settings from being applied before the operation runs.

- **Tests**
  - Updated VDO coverage for current resource requirements and supported index-memory options.

### DIFF
--- a/src/plugins/lvm/lvm-common.c
+++ b/src/plugins/lvm/lvm-common.c
@@ -19,6 +19,7 @@
 #include <glib.h>
 #include <math.h>
 #include <stdio.h>
+#include <blockdev/utils.h>
 #include <libdevmapper.h>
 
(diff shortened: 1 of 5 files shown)


tbzatek (dismissed):
wooo

tbzatek (approved):

PR #1210, opened 2026-09-03
loop: Add _from_fd variants for the remaining loop functions

### LINKED ISSUE #1208: Extend loop API to consume fd
Extend the loop plugin API to consume open filedescriptors instead of block device paths. This helps consumers to avoid race conditions by opening th device early on their side and just passing to libblockdev further in the process. The fd is being opened by the loop plugin functions anyway.

vojtechtrefny:
Add bd_loop_info_from_fd, bd_loop_teardown_from_fd, bd_loop_set_autoclear_from_fd and bd_loop_set_capacity_from_fd which operate on an already open file descriptor of the loop device instead of opening it by name. This lets consumers open the device early on their side and pass it in, avoiding race conditions.

The existing name-based functions now just open the device and delegate to their _from_fd counterparts (mirroring bd_loop_setup/setup_from_fd) to avoid duplicating the ioctl logic.

Resolves: https://github.com/storaged-project/libblockdev/issues/1208

## Summary by CodeRabbit

* **New Features**
  * Added file-descriptor-based operations for retrieving loop-device information, tearing down devices, managing autoclear settings, and updating capacity.
  * Existing path-based loop-device operations remain available.

* **Documentation**
  * Updated the loop API documentation index with the new operations.

* **Tests**
  * Added coverage for loop-device operations using open file descriptors, including offsets, autoclear settings, teardown, and capacity changes.


src/plugins/loop.c
```diff
@@ -114,11 +116,21 @@ BDLoopInfo* bd_loop_info_copy (BDLoopInfo *info) {
...
+    if (fstat (fd, &statbuf) != 0) {
+        g_set_error (error, BD_LOOP_ERROR, BD_LOOP_ERROR_FAIL,
+                     "Failed to stat the loop device: %m");
+        return NULL;
+    }
+
+    sys_path = g_strdup_printf ("/sys/dev/block/%u:%u/loop/backing_file",
+                                major (statbuf.st_rdev), minor (statbuf.st_rdev));
+
     if (access (sys_path, R_OK) != 0) {
```
tbzatek: Do we need this here? The following call to `g_file_get_contents()` would return an error anyway.

tbzatek (dismissed):
Nice, thanks!

src/plugins/loop.c
vojtechtrefny: This was interesting investigation -- the check was added intentionally in https://github.com/storaged-project/libblockdev/commit/4ff2ce447b28816f2169900b53e192e11f26b12c for the now removed `bd_loop_get_backing_file` function. It doesn't make sense in the `bd_loop_info` function, because for non-existing devices it returns error before trying to get the backing file. I removed the check in a separate commit.

tbzatek (approved):

